PR #61, opened 2019-04-10
Bugfix/panel expansion group styles

J-Kallunki:


src/core/theme/zindexes.ts
```diff
@@ -1,3 +1,4 @@
 export const zindexes = {
```
jporola: might be good idea to document these used z-indexes, if not already

J-Kallunki: Yeah have to think about later, might be refactored also. This is inside theme so at this point only key-names should be something that need to be understand. Else just avoid using z-indexes or use 1, 2, 3 that are local and modular.

PR #70, opened 2019-04-30
Feature/color tokens

J-Kallunki:


src/core/Form/Toggle/Toggle.baseStyles.tsx
```diff
@@ -52,10 +53,10 @@ export const baseStyles = ({ theme = suomifiTheme }: ToggleProps) => css`
         transform: translateX(50%);
       }
       .${svgPrefix}fi-toggle-icon-slide {
-        fill: ${theme.colors.toggleOnSlide};
+        fill: ${alphaHex(0.5)(theme.colors.successBase)};
```
jporola: should we have this color without need for opacity change?

J-Kallunki: That would be PAMU's decision

PR #75, opened 2019-05-02
Feature/update react typings

J-Kallunki:


src/reset/HtmlButton/HtmlButton.tsx
```diff
@@ -27,7 +27,7 @@ const buttonResets = css`
   ${resets.common}
   display: inline-block;
   max-width: 100%;
-  pointer: cursor;
```
jporola: stylelint (or similar) would prevent these kind of mistakes

J-Kallunki: Editor informed the error but need to configure for webpack also

PR #83, opened 2019-05-08
Static icons

jporola:


src/core/Icon/Icon.tsx
```diff
@@ -11,13 +11,18 @@ import {
...
+  IconKeys,
+  StaticIconKeys,
+} from 'suomifi-icons';
+export { IconKeys, StaticIconKeys } from 'suomifi-icons';
 import { logger } from '../../utils/logger';
 
 export interface IconProps extends Omit<CompIconProps, 'src'> {
   /** Icon-name from suomifi-icons */
-  icon?: IconKeys;
+  icon?: IconKeys | StaticIconKeys;
```
J-Kallunki: Ah would have been nice to export only one interface combining these byt yeah lets go with this

PR #84, opened 2019-05-09
Feature/spacing tokens

J-Kallunki:


src/core/theme/utils/spacing.ts
```diff
@@ -0,0 +1,17 @@
+import { spacingTokens } from '../spacing';
+
+type SpaceProp = keyof typeof spacingTokens | '0';
+
+const SpaceVal = (val: SpaceProp) => (val === '0' ? val : spacingTokens[val]);
+
+const space = (type: 'padding' | 'margin') => (
+  t: SpaceProp,
```
jporola: argument naming is little bit misleading because meaning is different when you give, one, two, three or four arguments, but if this is only internal util no big deal

J-Kallunki: Yeah but it uses CSS margin/padding API so does not need abstraction here

PR #138, opened 2019-08-07
Add '<constains semantic elements>'-tags to component styleguide pages

J-Kallunki:
To make more clear when using components when is semantic structure within app to be noted.

### DIFF
--- a/.styleguidist/styleguidist.styles.js
+++ b/.styleguidist/styleguidist.styles.js
@@ -42,7 +42,7 @@ module.exports = {
           position: 'relative',
           top: '-.75em',
           display: 'block',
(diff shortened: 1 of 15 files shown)


.styleguidist/styleguidist.styles.js
```diff
@@ -37,6 +37,16 @@ module.exports = {
     docs: {
       order: '1',
       marginBottom: '24px',
+      '& i.semantics': {
+        '&:before': {
+          position: 'relative',
+          top: '-.75em',
+          display: 'block',
+          content: '"<constains semantic elements>"',
```
aappoalander: Is "constains" a type? Should be "contains"?

PR #144, opened 2019-08-13
[Patch]Order button examples to separate groups

J-Kallunki:
Same format used elsewhere like Block examples


src/core/Button/Button.baseStyles.tsx
```diff
@@ -28,7 +28,7 @@ const negativeStyles = (theme: ThemeProp) => css`
     &[disabled],
     &:disabled {
       opacity: 0.5;
-      background: none;
+      background-color: ${theme.colors.highlightBase};
```
aappoalander: single comment

PR #153, opened 2019-09-04
Bugfix/Simplify icon positioning - closes #148

### LINKED ISSUE #148: [Bugfix] remove translateY from icon default styling
Causes problems with some use cases and the value should probably be adjusted elsewhere instead.

When fixed remove manual adjustment from design-system front page (ContentBoxes).

LJKaski:
Removed translateY from icon basestyle and adjusted affected components accordingly. Updated snapshots and fixed a minor issue which prevented prettier from finding certain files on windows systems.


src/core/Menu/Menu.baseStyles.tsx
```diff
@@ -22,9 +22,10 @@ export const baseStyles = ({
       border-radius: ${theme.radius.basic};
       text-transform: uppercase;
       & > .fi-menu-language_icon {
-        height: 16px;
-        width: 16px;
-        margin-left: ${theme.spacing.xxs};
+        height: 1.2em;
```
LJKaski: This is a minor size adjustment to make the element look closer to the design.

PR #159, opened 2019-09-16
Create asProp for withDefaultTheme-util

### LINKED ISSUE #157: Component 'as' prop not working as expected
'as' prop does not work as expected when used with e.g. Heading component. Component styles are not applied to the rendered component.

Tested with a NoteBox component in suomifi-design-system repo the following way:
<Heading.h3 as='h2'>{title}</Heading.h3>
When built with Gatsby the heading styles are not applied.

blocking #155

J-Kallunki:
Styled-components has `as`-prop that replaces `styled`-template literal tags' element/component. Our components use internal components with `styled` that are not the ones to be replaced. Using withDefaultTheme-util (will be later refactor at another fix) `as`-prop is now internally changed to `asProp` and implementations with that when needed. Link-component supports both public API `as` and internal API `asProp`.

closes #157

### DIFF
--- a/src/components/Heading/Heading.tsx
+++ b/src/components/Heading/Heading.tsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import classnames from 'classnames';
+import { AsProp } from '../../utils/typescript';
 import { HtmlH, HtmlHProps, hLevels } from '../../reset/HtmlH/HtmlH';
 export { hLevels };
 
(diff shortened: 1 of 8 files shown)


src/components/Heading/Heading.tsx
```diff
@@ -1,19 +1,28 @@
...
   render() {
-    const { className, variant = 'h1', as, ...passProps } = this.props;
+    const {
+      className,
+      variant = 'h1',
+      as: asStyled,
+      asProp,
+      ...passProps
+    } = this.props;
+    const as = !!asProp ? asProp : asStyled;
```
aappoalander: Do I understand the API correctly if I assume that this checks that only as or AsProp can be defined? AsProp for internal use and as for the public API? I'd like to see this explained / commented somehow.

aappoalander:
Link and Heading are the only existing components that we allow the use of as? Or are these somehow special cases?

src/components/Heading/Heading.tsx
J-Kallunki: `as` is asProp if defined, if not defined it is AsStyled that is just renamed public API `as`. If both undefined then it is undefined.

J-Kallunki:
> Link and Heading are the only existing components that we allow the use of as? Or are these somehow special cases?

These are the cases that need to be supported. And we cannot automatically support every component so every time support needs to be implemented - is it used in any case? Is it accessible to have wrong element in that component?

J-Kallunki:
Closing as it contains some really mysterious error. Redoing from scratch.

PR #163, opened 2019-09-25
Create internal asProp for defaultTheme-util

### LINKED ISSUE #157: Component 'as' prop not working as expected

J-Kallunki:
Styled-components has as-prop that replaces styled-template literal tags' element/component. Our components use internal components with styled that are not the ones to be replaced. Using defaultTheme-util (will be later refactor at another fix) as-prop is now internally changed to asProp and implementations with that when needed. Link-component supports both public API as and internal API asProp.

closes #157

### DIFF
--- a/src/core/Heading/Heading.tsx
+++ b/src/core/Heading/Heading.tsx
@@ -37,7 +37,7 @@ const StyledHeading = styled(
     smallScreen,
     className,
     variant,
-    asProp,
+    asProp, // as-property is defined internally as asProp and need to be implemented back if used
     ...passProps
   }: HeadingProps) => (
     <CompHeading
--- a/src/core/Heading/Heading.tsx
+++ b/src/core/Heading/Heading.tsx
@@ -9,7 +9,7 @@ import {
 } from '../../components/Heading/Heading';
 import { baseStyles } from './Heading.baseStyles';
 import classnames from 'classnames';
-import { Omit } from '../../utils/typescript';
+import { Omit, asPropType } from '../../utils/typescript';
 import { logger } from '../../utils/logger';
 
 const baseClassName = 'fi-heading';
@@ -27,6 +27,7 @@ export interface HeadingProps
   smallScreen?: boolean;
   /** Change font to smaller screen size and style */
   color?: ColorProp;
+  asProp?: asPropType;
 }
 
 const StyledHeading = styled(
@@ -36,6 +37,7 @@ const StyledHeading = styled(
     smallScreen,
     className,
     variant,
+    asProp,
     ...passProps
   }: HeadingProps) => (
     <CompHeading
@@ -44,6 +46,7 @@ const StyledHeading = styled(
         [smallScreenClassName]: smallScreen,
       })}
       variant={variant === 'h1hero' ? 'h1' : variant}
+      as={asProp}
     />
   ),
 )`
--- a/src/core/Link/Link.tsx
+++ b/src/core/Link/Link.tsx
@@ -6,6 +6,7 @@ import {
   Link as CompLink,
   LinkProps as CompLinkProps,
 } from '../../components/Link/Link';
+import { asPropType } from '../../utils/typescript';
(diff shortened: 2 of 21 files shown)


src/core/theme/utils/defaultTheme.ts
```diff
@@ -1,12 +1,21 @@
...
-  theme = suomifiTheme,
+  theme,
+  as,
   ...props
-}: Partial<T>): T =>
+}: Partial<T> & { as?: asPropType }): T =>
   ({
     ...props,
-    theme,
+    ...(!!as ? { asProp: as } : {}),
```
J-Kallunki: Our HTML-components have `as`-prop defined as they use `style` from styled-components that can handle element/component change with it. The type definition for it is traversed with `extended` to our final UI-components.
https://www.styled-components.com/docs/api#as-polymorphic-prop

Our UI-components use this defaultTheme-function to set defaults for theme. I'm implementing here that at the same time that prop named `as` is taken away and give as `asProp` - so that every time our internal implementation uses `styled` it does not change the internal component it has. Then we need to internally make implementation for `asProp` where we like to use it - otherwise it does nothing even when defined (originally `as`).

src/core/Heading/Heading.tsx
```diff
@@ -44,6 +46,7 @@ const StyledHeading = styled(
         [smallScreenClassName]: smallScreen,
       })}
       variant={variant === 'h1hero' ? 'h1' : variant}
+      as={asProp}
```
J-Kallunki: I decided to make the internal `asProp` -> `as` implementation here at `src/core`-level component as it has nothing to do with `src/components` level implementations. There it will work just fine to use `as` (supported by styled-components).

src/core/theme/utils/defaultTheme.ts
```diff
@@ -1,12 +1,21 @@
 import { suomifiTheme, ThemeProp } from '../';
+import { asPropType } from '../../../utils/typescript';
 
+/**
+ * Do defaults for component props
+ * Add default theme if theme not defined,
+ * Remove as-prop and change it to asProp if defined
+ * @param props Component props
+ */
 export const withDefaultTheme = <
```
J-Kallunki: Typings for the function:
- `T extends { theme: ThemeProp; asProp?: asPropType }` : `T` contains theme and optional asProp
- `Partial<T> & { as?: asPropType }` : given function gets object that has `T` (all optional keys) and optional `as`
- that function returns `T`(containing theme and optional asProp as said earlier)

src/core/Heading/Heading.tsx
```diff
@@ -37,7 +37,7 @@ const StyledHeading = styled(
     smallScreen,
     className,
     variant,
-    asProp,
+    asProp, // as-property is defined internally as asProp and need to be implemented back if used
```
aappoalander: I'd clarify this comment a bit further to explain that asProp needs to be implemented per use case / component if support for that is wanted.

src/core/Link/Link.tsx
```diff
@@ -6,17 +6,20 @@ import {
...
 type LinkVariant = 'default' | 'external';
 export interface LinkProps extends CompLinkProps, ThemeComponent {
   variant?: LinkVariant;
+  asProp?: asPropType;
 }
 
-const StyledLink = styled(({ theme, ...passProps }: LinkProps) => (
-  <CompLink {...passProps} />
+const StyledLink = styled(({ theme, asProp, ...passProps }: LinkProps) => (
+  // as-property is defined internally as asProp and need to be implemented back if used
```
aappoalander: I'd rephrase this "need to be implemented back if used" where ever used to better explain that asProp always needs custom implementation per component if support for it is desired.

PR #195, opened 2019-10-07
Bugfix/do not try to export suomifi theme function

### LINKED ISSUE #194: Exported suomifiTheme is now a function and it was an object

Export object

https://github.com/vrk-kpa/suomifi-ui-components/blob/develop/src/core/theme/index.ts#L49

`import { suomifiTheme } from 'suomifi-ui-components';`

J-Kallunki:
Do not try to export suomifiTheme function outside the library as it uses singleton to get theme from tokens. Exported defaultTokens contains also internal tokens so everything is still available.

Fix so that suomifiTheme function understands also given customized internalTokens.

Closes #194

### DIFF
--- a/src/core/theme/index.ts
+++ b/src/core/theme/index.ts
@@ -22,6 +22,7 @@ export interface SuomifiThemeProp {
   theme: SuomifiTheme;
 }
 
+type DefaultInternalTokens = typeof internalTokens;
 const internalTokens = {
   shadows,
   gradients,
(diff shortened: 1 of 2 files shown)


src/core/theme/index.ts
```diff
@@ -47,11 +48,13 @@ export const defaultTokens = {
...
-  spacing,
-  typography,
-}: SuomifiTokens = defaultTokens) => ({
+  colors = defaultTokens.colors,
+  spacing = defaultTokens.spacing,
+  typography = defaultTokens.typography,
+  ...internalTokensProp
+}: Partial<SuomifiTokens & DefaultInternalTokens> = defaultTokens) => ({
   ...internalTokens,
+  ...internalTokensProp,
```
aappoalander: What is this and where is this used / where does this come from?

J-Kallunki: This is the internal tokens that are customized. You can highlight it with IDE and typescript will tell what it contains. I'll update JSDOC for the function

LJKaski: Great. That's what I figured as well, but updating the JSDOC will definitely make it easier to see.

aappoalander: A bit difficult to follow what instances / types get returned, maybe rename internalTokensProp to be more descriptive when considering the use. In addition, documenting the return values and types might help.

PR #197, opened 2019-10-07
Feature/eslint

aappoalander:
## Description
This PR moves TypeScript linting to ESLINT. Related rules have been configured for ESLint and TSLint has been removed with related dependencies.

## Motivation and Context

Linting rules are now almost aligned with https://github.com/vrk-kpa/suomifi-design-system , some default rules needed to be turned off and some further code style fixes need to be implemented.

## How Has This Been Tested?

There are no changes to component API's or functionality. Only linting and code style is affected.


src/components/Form/Toggle.tsx
```diff
@@ -93,7 +93,6 @@ const componentOrElementWithProps = (
       props,
     ); // component
   }
-  return;
```
LJKaski: Is this and other similar changes here because of the linter change? What changed in the rules?

package.json
```diff
@@ -94,14 +107,7 @@
     "stylelint-custom-processor-loader": "0.6.0",
     "stylelint-processor-styled-components": "1.6.0",
     "ts-jest": "23.10.4",
-    "tslint": "5.11.0",
-    "tslint-config-airbnb": "5.11.1",
```
ketsappi: Are we missing something when dropping the tslint version, `tslint-config-airbnb`? Should we take [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb) to use?

aappoalander: Might make sense to use the eslint version of this. I think the main reason for not using it was to align with designs system repo. Will have a look what it would mean to add this as addition to everything else we use.

src/components/Form/Toggle.tsx
aappoalander: The rules are now aligned with the design system repo and are more strict in some areas. So yes it is directly related to changes in ruleset. There are additional changes required which are now suppressed as warnings in the linter config.

PR #201, opened 2019-10-09
[Feature] index.ts for utils/css

### LINKED ISSUE #152: Create index.ts for utils/css
Create index.ts for utils/css and export all css-utils. Rename all imports to use this 'utils/css'.

ketsappi:
## Description

Issue #152 

## Related Issue

Closes #152 

## Motivation and Context

Being consistent and use the same convention like in other places in the code.

## How Has This Been Tested?

`yarn validate`

### DIFF
--- a/src/components/Panel/PanelExpansion.tsx
+++ b/src/components/Panel/PanelExpansion.tsx
@@ -2,7 +2,7 @@ import React, { Component, ReactNode, Fragment } from 'react';
 import { default as styled } from 'styled-components';
 import { Panel, PanelProps, baseClassName, StyledPanel } from './Panel';
 import { Button, ButtonProps } from '../Button/Button';
-import { allStates } from '../../utils/css/pseudo';
+import { allStates } from '../../utils/css';
 import classnames from 'classnames';
 import {
   PanelExpansionGroupConsumer,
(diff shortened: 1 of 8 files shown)


ketsappi:
_Noted when creating the change and to be documented for later purpose._

---

Is this wanted to be like this; immediately exporting the imported `FlattenSimpleInterpolation`?
https://github.com/vrk-kpa/suomifi-ui-components/blob/39b5d003cf2097f19f3abce8310e51e18cf02de1/src/utils/css/utils.ts#L1-L2

As after the change one place where it is used is now like this:
https://github.com/vrk-kpa/suomifi-ui-components/blob/39b5d003cf2097f19f3abce8310e51e18cf02de1/src/core/theme/utils/typography.ts#L1-L4


---
What is the reason for doing like this by not importing `FlattenSimpleInterpolation` when it is needed straight from the original place?

J-Kallunki:
> _Noted when creating the change and to be documented for later purpose._
> 
> Is this wanted to be like this; immediately exporting the imported `FlattenSimpleInterpolation`?
> https://github.com/vrk-kpa/suomifi-ui-components/blob/39b5d003cf2097f19f3abce8310e51e18cf02de1/src/utils/css/utils.ts#L1-L2
> 
> As after the change one place where it is used is now like this:
> https://github.com/vrk-kpa/suomifi-ui-components/blob/39b5d003cf2097f19f3abce8310e51e18cf02de1/src/core/theme/utils/typography.ts#L1-L4
> 
> What is the reason for doing like this by not importing `FlattenSimpleInterpolation` when it is needed straight from the original place?
> 
> @J-Kallunki

This was intented to have controlled single point export for `FlattenSimpleInterpolation` (if needed t obe modified or anything) but it could be just more understandable to import it always from Styled components.

PR #205, opened 2019-10-10
[Refactor] Set button fullWidth using className

### LINKED ISSUE #176: Refactor Button fullWidth to use className
Button fullWidth is boolean that toggles styles on condition, use conditional className instead and add styles+className at baseStyles without condition.

LJKaski:
## Description

Removed the conditional style application from button baseStyles. Added a  corresponding className to be added conditionally instead.

## Related Issue

Closes #176 

## How Has This Been Tested?

Changes work fine in styleguide locally. Not tested elsewhere yet.

### DIFF
--- a/src/core/Button/Button.baseStyles.tsx
+++ b/src/core/Button/Button.baseStyles.tsx
@@ -113,7 +113,7 @@ export const baseStyles = withSuomifiTheme(
     user-select: none;
   }
 
-  &.fi-button--fullWidth {
+  &.fi-button--fullwidth {
     display: block;
     width: 100%;
   }
--- a/src/core/Button/Button.baseStyles.tsx
+++ b/src/core/Button/Button.baseStyles.tsx
@@ -3,11 +3,6 @@ import { withSuomifiTheme, TokensAndTheme, SuomifiThemeProp } from '../theme';
 import { ButtonProps } from './Button';
 import { element, focus, button } from '../theme/reset';
 
-const fullWidthStyles = css`
-  display: block;
-  width: 100%;
-`;
-
 const negativeStyles = ({ theme }: SuomifiThemeProp) => css`
   &.fi-button--negative {
     background: none;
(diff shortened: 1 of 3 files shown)


src/core/Button/Button.baseStyles.tsx
```diff
@@ -121,7 +113,11 @@ export const baseStyles = withSuomifiTheme(
     user-select: none;
   }
 
-  ${fullWidth && fullWidthStyles}
+  &.fi-button--fullWidth {
```
J-Kallunki: Use lowercase in CSS-classes - with single dash or without

src/core/Button/Button.tsx
```diff
@@ -57,9 +57,13 @@ const StyledButton = styled(
...
       {...passProps}
-      className={classnames(className, {
-        [`${baseClassName}--${variant}`]: variant !== 'default',
-      })}
+      className={classnames(
+        className,
+        {
+          [`${baseClassName}--${variant}`]: variant !== 'default',
+        },
+        fullWidth ? 'fi-button--fullWidth' : '',
```
J-Kallunki: Uses classnames-function here: https://github.com/JedWatson/classnames
It takes format {'class-name': condition} and can be included using new key in the same object as variant is implemented

src/core/Button/Button.baseStyles.tsx
LJKaski: Fixed.

src/core/Button/Button.tsx
LJKaski: Thanks. Fixed this one as well.

src/core/Button/Button.tsx
```diff
@@ -59,6 +60,7 @@ const StyledButton = styled(
       {...passProps}
       className={classnames(className, {
         [`${baseClassName}--${variant}`]: variant !== 'default',
+        [fullWidthClassName]: fullWidth,
```
J-Kallunki: This does not need array syntax as `classnames` can read the key-string. Array syntax is needed when using template-literals. Can be seen from the documentation https://github.com/JedWatson/classnames#usage

J-Kallunki: Actually I was wrong, it needs the Computed property name
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Computed_property_names
as the wanted className is set to the constant not the constant key itself.

PR #207, opened 2019-10-11
Tests for tokens-prop

### LINKED ISSUE #183: Create test for tokens-prop
Customized tokens with any component

J-Kallunki:
## Description

Added tests - snapshot and axe - for tokens-prop usage. Also confirmed that the snapshot contained the correct output of changing only one color token.

## Related Issue

Closes #183

## Motivation and Context

Tokens-prop was not tested and actually creating test confirmed that the implementation needed also to be fixed.

## How Has This Been Tested?

Manually checking the snapshot.
`yarn build`
`yarn start`
`yarn validate`


src/core/theme/index.ts
```diff
@@ -15,13 +15,22 @@ export type SpacingProp = keyof typeof importedTokens.spacing;
 export type SuomifiTheme = ReturnType<typeof suomifiTheme>;
 
 export interface TokensProp {
-  tokens?: SuomifiTokens;
+  /** Custom  Design-tokens or one category of tokens customized, clone defaultTokens for base */
+  tokens?: Partial<SuomifiTokens>;
+}
+
+export interface DefinedTokensProp {
```
aappoalander: This naming feels confusing to me. What is the meaning of "Defined" in this context?

aappoalander: If I understood correctly this typing is used for the resolved tokens that are the result of merge of default and user provided tokens. => "Defined" simply means that all the tokens are present. In addition this is only used internally and not a part of the public API.

Could we use a name that describes all three properties: internal nature, non optional existence and the merge from default and custom tokens? Say ResolvedTokensProp?

aappoalander: Even if ResolvedTokensProp describes the process used to come up with data of this type, it also describes what it is used for and implies that it is defined as internally needed (resolved).

src/core/theme/index.ts
```diff
@@ -74,6 +83,9 @@ export const suomifiTheme = ({
  * @param {function} baseStyles Function that will get components' props including tokens-prop and return CSS-styles
  */
 export const withSuomifiTheme = (
-  baseStyles: <K>(props: K & SuomifiThemeProp) => FlattenSimpleInterpolation,
-) => <T extends SuomifiThemeProp>({ tokens, ...passProps }: TokensProp & T) =>
-  baseStyles({ ...passProps, theme: suomifiTheme(tokens) });
+  baseStyles: <K>(props: K & TokensAndTheme) => FlattenSimpleInterpolation,
```
aappoalander: This typing is always a pleasure to read! I would consider breaking it down for the sake of readability. Not as part of this PR though.

PR #209, opened 2019-10-11
[Bugfix] Added the missing background-color for the Dropdown's baseStyle

### LINKED ISSUE #178: Dropdown has wrong background color
Has been correct (white) before but now (version 0.2.5) color is wrong (gray). See attached screenshot.

image

ketsappi:
## Description

Dropdown component had gray background when it should be white.

### Was able to reproduce the problem 
**Versions:** 0.2.4, 0.2.5. 0.2.6
**Operating system:** macOS Catalina 10.15
**Browsers:** Firefox, Safari

### Was **not** able to reproduce the problem with
**Versions:** 0.2.4, 0.2.5. 0.2.6
**Operating system:** macOS Catalina 10.15
**Browsers:** Chrome, Opera

## Related Issue

Closes #178 

## Motivation and Context

To look consistent with all applicable browsers and with intended colors.

## How Has This Been Tested?

Manually tested after the change with 
**Operating system:** macOS Catalina 10.15
**Browsers:** Chrome, Firefox, Safari, Opera

And seemed to work with all these browsers now.

And ran the validations
`yarn validate`

### DIFF
--- a/src/core/Dropdown/Dropdown.baseStyles.tsx
+++ b/src/core/Dropdown/Dropdown.baseStyles.tsx
@@ -9,6 +9,7 @@ export const baseStyles = withSuomifiTheme(
       position: relative;
       padding-right: 30px;
       text-align: left;
+      background-color: ${theme.colors.whiteBase};
       cursor: pointer;
       &:before {
         content: '';
--- a/src/core/Dropdown/Dropdown.tsx
+++ b/src/core/Dropdown/Dropdown.tsx
@@ -19,7 +19,7 @@ const StyledDropdown = styled(
   ({ tokens, ...passProps }: DropdownProps & InternalTokensProp) => (
     <CompDropdown
       {...passProps}
-      dropdownItemProps={{ className: 'fi-dropdown-item' }}
+      dropdownItemProps={{ className: 'fi-dropdown_item' }}
     />
   ),
 )`


ketsappi:
Noted while doing the fix:

https://github.com/vrk-kpa/suomifi-ui-components/blob/cbc3e490b8b5fbe224d907160789db699483c147/src/core/Dropdown/Dropdown.tsx#L19-L22

Should this be `fi-dropdown_item` instead of the current `fi-dropdown-item`?

src/core/Dropdown/Dropdown.baseStyles.tsx
```diff
@@ -9,6 +9,7 @@ export const baseStyles = withSuomifiTheme(
       position: relative;
       padding-right: 30px;
       text-align: left;
+      background-color: ${theme.colors.whiteBase};
```
aappoalander: Any idea where the grey background is coming from? Are there bleeding styles somewhere that could affect other things as well?

aappoalander:
> Noted while doing the fix:
> 
> https://github.com/vrk-kpa/suomifi-ui-components/blob/cbc3e490b8b5fbe224d907160789db699483c147/src/core/Dropdown/Dropdown.tsx#L19-L22
> 
> Should this be `fi-dropdown_item` instead of the current `fi-dropdown-item`?

As per discussion with @J-Kallunki, this looks like a bug and should be fixed and tested with the other fix in this PR to ensure there are no related underlying issues.

src/core/Dropdown/Dropdown.baseStyles.tsx
J-Kallunki: At this point we dont wanna debug default styles of 3rd party lib

PR #216, opened 2019-10-16
Panel expansion disabled attribute

### LINKED ISSUE #199: Use disabled-attribute for PanelExpansion
Should use this syntax:
```
<button aria-expanded="false">Question goes here</button>
<div hidden="true"></div>
```
Also React renders elements with hidden-attribute after all other elements.

Screenshot 2019-10-09 at 7 31 06

https://vrk-kpa.github.io/suomifi-ui-components/#/Components/Patterns/Panel

## Specifications

- Version: 0.2.6

aappoalander:
## Description
PanelExpansion now uses hidden attribute in addition to aria-hidden. The hidden attribute is conditionally removed when the panel is expanded while aria-hidden is always present and only it's value is derived from expanded state. Hidden content is force rendered by changing the element key when panel is either collapsed or expanded. The hidden prop is handled internally and cannot be provided as prop externally, so there should be no need for any existence checks etc.

## Related Issue

Closes #199 

## Motivation and Context

PanelExpansion was using only aria-hidden for expansion content, but to be accessible and semantically correct, hidden attribute should be used instead. Aria-hidden was left in place as a fallback for browsers that do not implement accessibility features for hidden attribute correctly.

## How Has This Been Tested?

Changes were tested locally with chrome browser and with Edge via BrowserStack using the the build and included styleguide. In addition, both bare accessible component and Suomifi styled component were tested using create react app project. 

The resulting html now also contains the properties provided in #199.

Screenshot 2019-10-23 at 13 56 23
Screenshot 2019-10-23 at 13 56 40

Screenshot 2019-10-23 at 13 55 47
Screenshot 2019-10-23 at 13 56 07

### DIFF
--- a/src/core/Panel/__snapshots__/PanelExpansion.test.tsx.snap
+++ b/src/core/Panel/__snapshots__/PanelExpansion.test.tsx.snap
@@ -274,6 +274,7 @@ exports[`calling render with the same component on the same container does not r
     </svg>
   </button>
   <div
+    aria-hidden="true"
     class="fi-panel-expansion_content c6 fi-panel c1 c2"
     hidden=""
   >
--- a/src/core/Panel/__snapshots__/PanelExpansion.test.tsx.snap
+++ b/src/core/Panel/__snapshots__/PanelExpansion.test.tsx.snap
@@ -274,8 +274,8 @@ exports[`calling render with the same component on the same container does not r
     </svg>
   </button>
   <div
-    aria-hidden="true"
     class="fi-panel-expansion_content c6 fi-panel c1 c2"
+    hidden=""
   >
     Test expansion content
   </div>
--- a/src/components/Panel/PanelExpansion.tsx
+++ b/src/components/Panel/PanelExpansion.tsx
@@ -20,6 +20,7 @@ const contentOpenClassName = `${contentBaseClassName}--open`;
 
 interface StyledPanelExpansionContentProps extends PanelProps {
   openState: boolean;
+  hidden: boolean;
 }
 
 export const StyledPanelExpansionContent = styled(
@@ -169,7 +170,7 @@ export class PanelExpansionItem extends Component<PanelExpansionProps> {
           className={classnames(contentBaseClassName, contentClassName, {
             [contentOpenClassName]: !!openState,
           })}
-          aria-hidden={!openState}
+          hidden={!openState}
         >
           {children}
         </StyledPanelExpansionContent>
--- a/src/components/Panel/PanelExpansion.tsx
+++ b/src/components/Panel/PanelExpansion.tsx
@@ -171,6 +171,8 @@ export class PanelExpansionItem extends Component<PanelExpansionProps> {
             [contentOpenClassName]: !!openState,
           })}
           hidden={!openState}
+          key={String(openState)}
+          aria-hidden={!openState}
         >
           {children}
         </StyledPanelExpansionContent>
--- a/src/core/theme/__snapshots__/tokens.test.tsx.snap
+++ b/src/core/theme/__snapshots__/tokens.test.tsx.snap
@@ -408,8 +408,8 @@ exports[`snapshot testing 1`] = `
         </svg>
       </button>
       <div
-        aria-hidden="true"
         class="fi-panel-expansion_content c9 fi-panel c6 c2"
+        hidden=""
       >
         Test expansion content 1
       </div>
@@ -440,8 +440,8 @@ exports[`snapshot testing 1`] = `
         </svg>
       </button>
       <div
-        aria-hidden="true"
         class="fi-panel-expansion_content c9 fi-panel c6 c2"
+        hidden=""
       >
         Test expansion content 2
       </div>
@@ -472,8 +472,8 @@ exports[`snapshot testing 1`] = `
         </svg>
       </button>
       <div
-        aria-hidden="true"
         class="fi-panel-expansion_content c9 fi-panel c6 c2"
+        hidden=""
       >
         Test expansion content 3
       </div>
--- a/src/core/theme/__snapshots__/tokens.test.tsx.snap
+++ b/src/core/theme/__snapshots__/tokens.test.tsx.snap
@@ -408,6 +408,7 @@ exports[`snapshot testing 1`] = `
         </svg>
       </button>
       <div
+        aria-hidden="true"
         class="fi-panel-expansion_content c9 fi-panel c6 c2"
         hidden=""
       >
@@ -440,6 +441,7 @@ exports[`snapshot testing 1`] = `
         </svg>
       </button>
       <div
+        aria-hidden="true"
         class="fi-panel-expansion_content c9 fi-panel c6 c2"
         hidden=""
       >
@@ -472,6 +474,7 @@ exports[`snapshot testing 1`] = `
         </svg>
       </button>
       <div
+        aria-hidden="true"
         class="fi-panel-expansion_content c9 fi-panel c6 c2"
         hidden=""
       >


src/core/Panel/__snapshots__/PanelExpansion.test.tsx.snap
```diff
@@ -274,8 +274,8 @@ exports[`calling render with the same component on the same container does not r
     </svg>
   </button>
   <div
-    aria-hidden="true"
     class="fi-panel-expansion_content c6 fi-panel c1 c2"
+    hidden=""
```
aappoalander: @J-Kallunki is this ok? Seems to behave differently with test snapshots when compared to an actual browser.

ketsappi:
> ## Motivation and Context
> PanelExpansion was using aria-hidden for expansion content, but to be accessible hidden attribute should be used instead.

Here is an article that demonstrates the differences of `aria-hidden` and `hidden` attributes to get better understanding the reasoning of **why** this change was needed altogether.

https://developer.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/

aappoalander:
> > ## Motivation and Context
> > PanelExpansion was using aria-hidden for expansion content, but to be accessible hidden attribute should be used instead.
> 
> Here is an article that demonstrates the differences of `aria-hidden` and `hidden` attributes to get better understanding the reasoning of **why** this change was needed altogether.
> 
> https://developer.paciellogroup.com/blog/2012/05/html5-accessibility-chops-hidden-and-aria-hidden/

There seems to be an issue with using boolean attributes with IE and Edge, have to look further into it, but some sources seem to indicate using `visibility: hidden` would be sufficient with our existing solution of using `height: 0` when hiding panel content.

src/core/Panel/__snapshots__/PanelExpansion.test.tsx.snap
aappoalander: Also using css and visibility:hidden seems not to be working correctly. At least the hidden node remains in the accessibility tree for edge in with both use cases. When collapsing the group things seem to work better. BrowserStack was used for testing, so that might spice things up to some extent as well.

aappoalander: Conditionally rendering the expansible content (not using hidden at all, but rather not rendering the whole component for the content) seems to work correctly with edge and accessibility tree. This however is not probably the best approach to achieve the functionality.

Currently used aria-hidden attribute is correctly reflected to dom as well, but does not reflect to Edge's accessibility tree.

LJKaski: Tested this in Edge on Windows 10 and it behaved as above. hidden="" prop remained as well as the accessibility tree element, even though the element behaved correctly in the view.

J-Kallunki:
So has it been tested setting `hidden="hidden"` and removing the attribute when not to be set?
https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute

src/core/Panel/__snapshots__/PanelExpansion.test.tsx.snap
aappoalander: To sum things up:
- `hidden` attribute does not seem to behave as expected in Edge/IE
- `visibility: hidden` has also no effect on the accessibility tree in Edge/IE
- conditional rendering of the content seems to work as well as using aria-hidden.

So any combination of these is possible, but Edge and IE require either using aria-hidden in combination with visual hiding technique or conditionally hiding the actual content. If content is conditionally displayed, it might make sense to wrap the content and apply visual hiding there so that animations and other visual representations work as expected.

Waiting for comment from @J-Kallunki before progressing with this.

aappoalander:
> So has it been tested setting `hidden="hidden"` and removing the attribute when not to be set?
> https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute

It has been tested with conditionally setting the hidden prop for content when ever the panel is collapsed / closed. The issue is partially with Reacts way for handling the props => with edge/ie the corresponding attributes remain in dom, so setting `hidden={false}` after it has been set true once does not remove the attribute, nor has any effect on the accessibility tree in Edge/IE. The attribute remains in dom like `hidden=""` and as far as I know, the only way to remove it would be to directly manipulate dom.

J-Kallunki:
> > So has it been tested setting `hidden="hidden"` and removing the attribute when not to be set?
> > https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute
> 
> It has been tested with conditionally setting the hidden prop for content when ever the panel is collapsed / closed. The issue is partially with Reacts way for handling the props => with edge/ie the corresponding attributes remain in dom, so setting `hidden={false}` after it has been set true once does not remove the attribute, nor has any effect on the accessibility tree in Edge/IE. The attribute remains in dom like `hidden=""` and as far as I know, the only way to remove it would be to directly manipulate dom.

I mean setting conditionally like `const hiddenAttribute = hiddenCondition ? {hidden: "hidden"} : {}` and `<Component {...hiddenAttribute} />`

aappoalander:
> > > So has it been tested setting `hidden="hidden"` and removing the attribute when not to be set?
> > > https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attribute
> > 
> > 
> > It has been tested with conditionally setting the hidden prop for content when ever the panel is collapsed / closed. The issue is partially with Reacts way for handling the props => with edge/ie the corresponding attributes remain in dom, so setting `hidden={false}` after it has been set true once does not remove the attribute, nor has any effect on the accessibility tree in Edge/IE. The attribute remains in dom like `hidden=""` and as far as I know, the only way to remove it would be to directly manipulate dom.
> 
> I mean setting conditionally like `const hiddenAttribute = hiddenCondition ? {hidden: "hidden"} : {}` and `<Component {...hiddenAttribute} />`

Yes. That ie exactly what was tested and the end result in Edge/IE is the same, the attribute is not removed from dom.

J-Kallunki:
It might be this https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12087679/

aappoalander:
Could be it. It's difficult to fix without undesired or unknown side-effects since adding/removing the attribute is handled by React. If possible, I'd like to try to find an approach which works for all supported browsers, but does not require accessing dom and directly using e.g. `removeAttribute`.

Could we use hidden attribute for the content container and more traditional method for content it self? So the content container would use hidden attribute, and then additional span or div would use conditional rendering and wrap the actual content?

J-Kallunki:
> Could be it. It's difficult to fix without undesired or unknown side-effects since adding/removing the attribute is handled by React. If possible, I'd like to try to find an approach which works for all supported browsers, but does not require accessing dom and directly using e.g. `removeAttribute`.
> 
> Could we use hidden attribute for the content container and more traditional method for content it self? So the content container would use hidden attribute, and then additional span or div would use conditional rendering and wrap the actual content?

Try this:
```js
const hiddenAttribute = hiddenCondition ? {hidden: "hidden"} : {}
const hiddenKey = hiddenCondition ? 'hidden' : 'unhidden';
<div {...hiddenAttribute}  key={hiddenKey}>
```

src/core/Panel/__snapshots__/PanelExpansion.test.tsx.snap
aappoalander: Using key to force render of expansible content seems to resolve the issue with hidden attribute remaining in dom. So using conditional hidden prop / attribute with aria-hidden as fallback seems to be working with all browsers.

aappoalander:
> > Could be it. It's difficult to fix without undesired or unknown side-effects since adding/removing the attribute is handled by React. If possible, I'd like to try to find an approach which works for all supported browsers, but does not require accessing dom and directly using e.g. `removeAttribute`.
> > Could we use hidden attribute for the content container and more traditional method for content it self? So the content container would use hidden attribute, and then additional span or div would use conditional rendering and wrap the actual content?
> 
> Try this:
> 
> ```js
> const hiddenAttribute = hiddenCondition ? {hidden: "hidden"} : {}
> const hiddenKey = hiddenCondition ? 'hidden' : 'unhidden';
> <div {...hiddenAttribute}  key={hiddenKey}>
> ```

Using the key seems to force re-render for the component / tag that has the hidden attribute. So not updating with edge/ie seems to be solved by using key. IE/Edge do not however respect the hidden attribute alone and expansible content would remain in the accessibility tree with hidden attribute present alone. Keeping aria-hidden as a fallback seems to resolve this issue as well.

PR #218, opened 2019-10-17
[Patch] Button.tertiary colors to match design

ketsappi:
## Description

- New background gradient created for Button.tertiary, previously it used just one color
- Hover background color changed for Button.tertiary
- Updated the snapshot-test accordingly

## Motivation and Context

Button.tertiary colors were not matching the design's colors.

## How Has This Been Tested?

- Manually viewing with the browser
- `yarn validate`

### DIFF
--- a/src/core/Button/Button.baseStyles.tsx
+++ b/src/core/Button/Button.baseStyles.tsx
@@ -71,11 +71,11 @@ const secondaryNoBorderStyles = ({ theme }: SuomifiThemeProp) => css`
 const tertiaryStyles = ({ theme }: SuomifiThemeProp) => css`
   &.fi-button--tertiary {
     ${secondary({ theme })}
-    background: ${theme.colors.highlightLight50};
+    background: ${theme.gradients.highlightLight45ToHighlightLight50};
     border: none;
 
     &:hover {
-      background: ${theme.colors.highlightLight53};
+      background: ${theme.colors.highlightLight50};
     }
 
     &:active {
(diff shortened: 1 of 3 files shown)


src/core/Button/Button.baseStyles.tsx
```diff
@@ -71,11 +71,11 @@ const secondaryNoBorderStyles = ({ theme }: SuomifiThemeProp) => css`
 const tertiaryStyles = ({ theme }: SuomifiThemeProp) => css`
   &.fi-button--tertiary {
     ${secondary({ theme })}
-    background: ${theme.colors.highlightLight50};
+    background: ${theme.gradients.highlightLight45ToHighlightLight50};
```
aappoalander: Wow, that's hard to read, could you omit the the name before 50?  ```highlightLight45To50``` or would it be too vague without it?

ketsappi: True, `highlightLight45To50` sounds better.

Or should we have explicit naming like it is now -> `highlightLight45ToHighlightLight50`?
@J-Kallunki

J-Kallunki: @ketsappi 
It should say how it is formed. If it is using to tokens it should have the tokens names there. If it using values it could say just changing value.

I think here it should include full token names as they are tokens that just happen to have similar base. If we would change value inside token to something not a token we could say what changed.

And if we named internal tokens and just created description comments for them, but I would try to avoid desining internal tokens as they should only say something that it is strongly based on tokens.

Issue #154, opened 2019-09-04
[Bugfix] Fix toggle basic functionality


LJKaski:
Can only be toggled on by dragging, cannot be toggled off. State change triggers twice when clicking the toggle in the styleguide.

J-Kallunki:
Should base on this format:
```html
<label for="suomifi-checkbox-id" class=“suomifi-checkbox">
  <input type="checkbox" id="suomifi-checkbox-id" class=“suomifi-checkbox_input" />
  <svg class=“suomifi-checkbox_icon" />
  <span>Label for the checkbox</span>
</label>
```
```css
.suomifi-checkbox .suomifi-checkbox_input:checked + .suomifi-checkbox_icon {
}
.suomifi-checkbox .suomifi-checkbox_input:focus + .suomifi-checkbox_icon {
}
.suomifi-checkbox .suomifi-checkbox_input {
  position: absolute;
  // SVG-ikonin koko
  width: 1em;
  height: 1em;
  opacity: 0;
}
```
and use idGenerator()

PR #224, opened 2019-10-25
[Feature] Differentiating input and button versions of Toggle

### LINKED ISSUE #136: Make Toggle component use button element not input checkbox

ketsappi:
## Description

Toggle will now have button-component by default and with the `withInput` attribute you can get the input-component version.

`toggleInputProps` and `toggleInputComponent` attributes will only work for the input-version.

Different versions are in separate files: `ToggleInput` and `ToggleButton`

```
suomifi-ui-components/src/components/Form
├── Toggle.tsx
├── ToggleButton.tsx
└── ToggleInput.tsx
```

## Related Issue

This PR closes #136 and is related to the #154 changes (toggle functionality fixes incoming in later PR).

## How Has This Been Tested?
- manually in the browser
- `yarn validate`

### DIFF
--- a/src/components/Form/Toggle.tsx
+++ b/src/components/Form/Toggle.tsx
@@ -5,14 +5,13 @@ import React, {
   ComponentClass,
   FunctionComponent,
 } from 'react';
-import { HtmlLabel, HtmlInput, HtmlSpan } from '../../reset';
 import { idGenerator } from '../../utils/uuid';
-import classnames from 'classnames';
+import { ToggleInput } from './ToggleInput';
+import { ToggleButton } from './ToggleButton';
+export { ToggleButton, ToggleInput };
 
-const baseClassName = 'fi-toggle';
-const toggleDisabledClassName = `${baseClassName}--disabled`;
-const toggleInputClassName = `${baseClassName}_input`;
-const toggleLabelClassName = `${baseClassName}_label`;
+export const baseClassName = 'fi-toggle';
+// const toggleDisabledClassName = `${baseClassName}--disabled`;
 
 export interface ToggleInputProps {
   /** State of input checkbox */
@@ -24,20 +23,6 @@ export interface ToggleInputProps {
   controlled?: boolean;
   'aria-label'?: string;
   'aria-labelledby'?: string;
-}
-
-export class ToggleInput extends Component<ToggleInputProps> {
-  render() {
-    const { disabled = false, controlled, checked, ...passProps } = this.props;
-    return (
-      <HtmlInput
-        disabled={disabled}
-        {...passProps}
-        type="checkbox"
-        checked={!!checked}
-      />
-    );
-  }
 }
 
 export interface ToggleProps {
@@ -57,6 +42,8 @@ export interface ToggleProps {
    * Label element content
    */
   children?: ReactNode;
+  /** Use input instead of button */
+  withInput?: boolean;
   /** Pass custom props to Toggle's input component/element */
   toggleInputProps?: ToggleInputProps;
   /** Customized ToggleInput-component */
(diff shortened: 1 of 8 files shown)


src/components/Form/Toggle.tsx
```diff
@@ -5,14 +5,12 @@ import React, {
...
 import { idGenerator } from '../../utils/uuid';
-import classnames from 'classnames';
+import { ToggleInput } from './ToggleInput';
+import { ToggleButton } from './ToggleButton';
 
-const baseClassName = 'fi-toggle';
-const toggleDisabledClassName = `${baseClassName}--disabled`;
-const toggleInputClassName = `${baseClassName}_input`;
-const toggleLabelClassName = `${baseClassName}_label`;
+export { ToggleButton, ToggleInput };
```
J-Kallunki: Does these need to be exported also separately?

src/components/Form/Toggle.tsx
```diff
@@ -76,97 +61,27 @@ export interface ToggleProps {
...
     return (
-      <HtmlSpan
-        className={classnames(className, baseClassName, {
-          [toggleDisabledClassName]: !!disabled,
-        })}
-      >
-        {!!toggleInputComponent ? (
-          componentOrElementWithProps(toggleInputComponent, newToggleInputProps) // element
+      <React.Fragment>
+        {!!withInput ? (
```
J-Kallunki: You can also just do `return !!withInput ? (...`

src/components/Form/ToggleButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+      defaultChecked: dissMissDefaultChecked,
+      onClick: dissMissOnClick,
+      toggleInputProps,
+      toggleInputComponent,
+      ...passProps
+    } = this.props;
+    const { toggleState } = this.state;
+    const toggleButtonClassName = `${baseClassName}_button`;
+
+    if (toggleInputProps || toggleInputComponent) {
```
J-Kallunki: I would use double negative (`!!toggleInputProps || !!toggleInputComponent`) checking here - also because these props come from outside the library.

src/components/Form/ToggleButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+      logger.error(
+        `ToggleButton does not utilize 'toggleInputProps' and 'toggleInputComponent' props.`,
+      );
+      return false;
```
J-Kallunki: Would it be just ok to give the error and not prevent the normal return?

J-Kallunki:
Example markdown for Toggle could contain something like this:
```
import { Toggle } from 'suomifi-ui-components';

<>
<Toggle onClick={({ toggleState }) => console.log(toggleState)}>
  Toggle button
</Toggle>
<Toggle withInput={true} onClick={({ toggleState }) => console.log(toggleState)}>
  Toggle input
</Toggle>
</>
```

src/components/Form/ToggleButton.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    return (
+      <HtmlButton
+        className={toggleButtonClassName}
```
J-Kallunki: Seem to not be getting this. Should it be that taking `className` from this.props and then using `classnames(toggleButtonClassName, className)`.

I would also like that both toggles get `fi-toggle`, then also `fi-toggle--button` or `fi-toggle--input`.

src/components/Form/ToggleInput.tsx
```diff
@@ -0,0 +1,104 @@
...
+  return;
+};
+
+export class ToggleInput extends Component<ToggleProps> {
+  state: ToggleState = {
+    toggleState: !!this.props.checked || !!this.props.defaultChecked || false,
+  };
+
+  componentWillReceiveProps(nextProps: ToggleProps) {
+    const { checked } = nextProps;
```
J-Kallunki: Is this implementation missing from the button-version?

Also could better rephrase `checked`-prop interface documentation:
`Controlled toggle-state - user actions use onClick to change`

src/core/Form/Toggle/Toggle.tsx
```diff
@@ -54,7 +54,11 @@ export class Toggle extends Component<ToggleProps> {
     const { toggleStatus } = this.state;
 
     return (
```
J-Kallunki: Should this file also contain `Toggle.withInput` and actually implement `variant = 'default' | 'withInput'` to `src/components`? (not boolean)

src/components/Form/Toggle.tsx
ketsappi: I made it simpler and now returning with the unneeded React.Fragment. Committed.

src/components/Form/ToggleButton.tsx
ketsappi: Changed to use double negation.

src/components/Form/ToggleButton.tsx
ketsappi: No more preventing normal return; just giving the error now.

src/components/Form/ToggleInput.tsx
ketsappi: I added the missing implementation for button-version. And also rephrased the interface documentation.

ketsappi:
> Example markdown for Toggle could contain something like this:
> 
> ```
> import { Toggle } from 'suomifi-ui-components';
> 
> <>
> <Toggle onClick={({ toggleState }) => console.log(toggleState)}>
>   Toggle button
> </Toggle>
> <Toggle withInput={true} onClick={({ toggleState }) => console.log(toggleState)}>
>   Toggle input
> </Toggle>
> </>
> ```

I added an example use case also for the `withInput`version.

src/components/Form/ToggleButton.tsx
ketsappi: Not sure I understood the latter part correctly, because if I change the ToggleInput to use `fi-toggle--input` instead of previous `fi-toggle_input` it will then show checkbox before the component like in the picture.

Screenshot 2019-10-30 at 14 35 03

J-Kallunki: `fi-toggle_input` should not be changed. Just that the `span` and `button` would both contain `fi-toggle` but also `fi-toggle--button` for the button and `fi-toggle--input` for the input.

src/components/Form/Toggle.tsx
ketsappi: No longer exporting the ToggleButton.
This should now be solved with recent commit.

src/components/Form/ToggleButton.tsx
ketsappi: This should be now as you described in the recent commit.

src/components/Form/ToggleButton.tsx
```diff
@@ -0,0 +1,63 @@
...
+      logger.error(
+        `ToggleButton does not utilize 'toggleInputProps' and 'toggleInputComponent' props.`,
+      );
+    }
+
+    return (
+      <HtmlButton
+        className={toggleButtonClassName}
+        aria-disabled={disabled}
+        disabled={disabled}
```
aappoalander: These might not work with IE as expected unless something is forcing a complete re-render of the component. IE/Edge do not handle boolean attributes always properly as was discovered in non-related PR earlier. If an attribute remains in DOM when set to false (e.g. <div disabled="" >...</div>) an easy fix is to add a key and bind the value to what ever needs to trigger the "re-render".

src/core/Form/Toggle/Toggle.tsx
ketsappi: Now in the recent commit, using the variant instead of the 'withInput' boolean.

src/components/Form/ToggleButton.tsx
ketsappi: I tried with Win 10 and Edge and when I added `disabled={true}` it added `disabled=""` to the DOM.
But when I changed the value to `disabled={false}` the `disabled=""` was gone from the DOM.

So I did not make change regarding this for now as it seems to work.

PR #227, opened 2019-10-28
[Bugfix] Only allow placeholder for search input

### LINKED ISSUE #185: Only allow SearchInput to use placeholder
1. Make sure that `SearchInput `-component `placeholder` is implemented the same way as `placeholder` in `TextInput`.
2. Remove the variant of `TextInput` component that uses `placeholder`.
3. Add `placeholder` attribute to `TextInput` -interface without description. Remove it from html output. Add `logger.error` that it is not accessible when using `placeholder`-prop/attribute and it is ignored.

https://www.smashingmagazine.com/2018/06/placeholder-attribute/

LJKaski:
## Description

Adjusted TextInput and SearchInput components to the following standard:

1. Placeholder is not allowed in TextInput, and the user gets an error in console if one is used.
2. Placeholder is allowed in SearchInput, but only it is forced to match labelText. User gets warning in console if different placeholder is used.
3. Label can be made hidden or visible in both TextInput and SearchInput

Had to make some bigger adjustments since SearchInput uses TextInput. Need to make some changes to the way these work. Did it as follows:

1. Made a new class of BaseTextInput which essentially is the old TextInput without the required error when using placeholder.
2. Exported another class by the name of TextInput, which extends BaseTextInput and adds the error message and a slightly different logic regarding the label and placeholder.
3. Set the SearchInput to extend the BaseTextInput class to allow the use of placeholder

The BaseTextInput is not listed in the Index.ts.

## Related Issue

https://github.com/vrk-kpa/suomifi-ui-components/issues/185

Closes #185 

Placeholder is not accessible in many cases and it shouldn't be used instead of label. Now it can only be used in SearchInput and even there only when the placeholder matches the label.

## How Has This Been Tested?

yarn validate and manual testing.

### DIFF
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -10,8 +10,8 @@ import {
 } from '../../reset';
 import { VisuallyHidden } from '../Visually-hidden/Visually-hidden';
 import { Paragraph, ParagraphProps } from '../Paragraph/Paragraph';
-import { logger } from '../../utils/logger';
 import classnames from 'classnames';
+import { logger } from '../../utils/logger';
 
 const baseClassName = 'fi-text-input';
 const disabledClassName = `${baseClassName}--disabled`;
@@ -20,7 +20,7 @@ const inputBaseClassName = `${baseClassName}_input`;
 
 export interface TextInputLabelProps extends HtmlLabelProps {}
 
-type Label = 'default' | 'screenreader';
+type Label = 'hidden' | 'visible';
 
 export interface TextInputProps extends Omit<HtmlInputProps, 'type'> {
   /** Custom classname for the input to extend or customize */
@@ -42,17 +42,17 @@ export interface TextInputProps extends Omit<HtmlInputProps, 'type'> {
   /** Label */
   labelText: string;
   /** Label displaymode
-   * default: show above, screenreader: show as placeholder
+   * visible: show above, hidden: show as placeholder
    */
   labelMode?: Label;
-  /** Input placeholder-text if not using screenreader-labelmode */
+
   placeholder?: string;
   /** Input container div to define custom styling */
   inputContainerProps?: HtmlDivProps;
   children?: ReactNode;
 }
 
-export class TextInput extends Component<TextInputProps> {
+export class BaseTextInput extends Component<TextInputProps> {
   render() {
     const {
       className,
@@ -67,12 +67,7 @@ export class TextInput extends Component<TextInputProps> {
       ...passProps
     } = this.props;
 
-    const ifScreenreader = labelMode === 'screenreader';
-    if (ifScreenreader && !!placeholder) {
-      logger.warn(
-        'Placeholder not used as label is hidden and using label instead',
-      );
-    }
+    const hideLabel = labelMode === 'hidden';
 
     return (
       <HtmlLabel
@@ -81,7 +76,7 @@ export class TextInput extends Component<TextInputProps> {
           [disabledClassName]: !!passProps.disabled,
         })}
       >
-        {ifScreenreader ? (
+        {hideLabel ? (
           <VisuallyHidden>{labelText}</VisuallyHidden>
         ) : (
           <Paragraph {...labelTextProps}>{labelText}</Paragraph>
@@ -91,11 +86,24 @@ export class TextInput extends Component<TextInputProps> {
             {...passProps}
             className={classnames(inputBaseClassName, inputClassName)}
             type="text"
-            placeholder={ifScreenreader ? labelText : placeholder}
+            placeholder={hideLabel ? labelText : ''} // halutaanko että on sauma muutenkin käyttää labeltextiä tässä
           />
           {children}
         </HtmlDiv>
       </HtmlLabel>
     );
+  }
+}
+
+export class TextInput extends BaseTextInput {
+  render() {
+    const { placeholder, ...passProps } = this.props;
+
+    if (!!placeholder) {
+      logger.error(
+        'Using placeholder in text inputs is not accessible and it is ignored. Use labelText instead',
+      );
+    }
+    return <BaseTextInput {...passProps} />;
   }
 }
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -86,7 +86,7 @@ export class BaseTextInput extends Component<TextInputProps> {
             {...passProps}
             className={classnames(inputBaseClassName, inputClassName)}
             type="text"
-            placeholder={hideLabel ? labelText : ''} // halutaanko että on sauma muutenkin käyttää labeltextiä tässä
+            placeholder={hideLabel ? labelText : ''}
           />
           {children}
         </HtmlDiv>
@@ -97,13 +97,48 @@ export class BaseTextInput extends Component<TextInputProps> {
 
 export class TextInput extends BaseTextInput {
   render() {
-    const { placeholder, ...passProps } = this.props;
+    const {
+      className,
+      inputClassName,
+      labelText,
+      labelMode,
+      placeholder,
+      labelProps,
+      labelTextProps,
+      inputContainerProps,
+      children,
+      ...passProps
+    } = this.props;
 
     if (!!placeholder) {
       logger.error(
         'Using placeholder in text inputs is not accessible and it is ignored. Use labelText instead',
       );
     }
-    return <BaseTextInput {...passProps} />;
+
+    const hideLabel = labelMode === 'hidden';
+
+    return (
+      <HtmlLabel
+        {...labelProps}
+        className={classnames(labelBaseClassName, className, {
+          [disabledClassName]: !!passProps.disabled,
+        })}
+      >
+        {hideLabel ? (
+          <VisuallyHidden>{labelText}</VisuallyHidden>
+        ) : (
+          <Paragraph {...labelTextProps}>{labelText}</Paragraph>
+        )}
+        <HtmlDiv {...inputContainerProps}>
+          <HtmlInput
+            {...passProps}
+            className={classnames(inputBaseClassName, inputClassName)}
+            type="text"
+          />
+          {children}
+        </HtmlDiv>
+      </HtmlLabel>
+    );
   }
 }
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -60,7 +60,6 @@ export class BaseTextInput extends Component<TextInputProps> {
       inputClassName,
       labelText,
       labelMode,
-      placeholder,
       labelProps,
       labelTextProps,
       inputContainerProps,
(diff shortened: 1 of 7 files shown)


src/components/Form/TextInput.tsx
```diff
@@ -67,21 +67,65 @@ export class TextInput extends Component<TextInputProps> {
...
+            placeholder={hideLabel ? labelText : !!placeholder ? labelText : ''}
+          />
+          {children}
+        </HtmlDiv>
+      </HtmlLabel>
+    );
+  }
+}
+
+export class TextInput extends BaseTextInput {
```
LJKaski: This is written in this manner to allow the different logic for showing/disabling placeholder. If there's a cleaner way to utilize the BaseTextInput and overwrite the placeholder logic, let me know.

src/core/Form/SearchInput/__snapshots__/SearchInput.test.tsx.snap
```diff
@@ -90,41 +90,49 @@ exports[`calling render with the same component on the same container does not r
...
-  height: 1px;
-  width: 1px;
-  margin: -1px;
-  padding: 0;
-  border: 0;
-  overflow: hidden;
+  top: 50%;
+  right: 16px;
+  margin-top: -0.5em;
 }
```
LJKaski: I actually don't really get where some of these changes came from. Enlighten me if you can. Things look normal.

src/core/Form/TextInput/TextInput.md
```diff
@@ -13,18 +13,19 @@ import { TextInput } from 'suomifi-ui-components';
...
+    labelMode="hidden"
     labelText="Test with hidden label"
+    defaultValue="Test with hidden label"
   />
   <TextInput.error
-    labelMode="screenreader"
+    labelMode="hidden"
     labelText="Error with hidden label"
-    value="Error with hidden label"
+    defaultValue="Error with hidden label"
```
LJKaski: Replaced value with defaultValue to make the example inputs interactive in the styleguide. Text couldn't be edited when label was set the way it was.

src/components/Form/TextInput.tsx
```diff
@@ -20,7 +20,7 @@ const inputBaseClassName = `${baseClassName}_input`;
 
 export interface TextInputLabelProps extends HtmlLabelProps {}
 
-type Label = 'default' | 'screenreader';
+type Label = 'hidden' | 'visible';
```
ketsappi: Hmm.. would it possible to use enum instead? 

E.g:
`
enum Label {
  Hidden = "hidden",
  Visible = 'visible'
};
`

Then we might avoid possible typo when later using enum values instead of strings.

`const hideLabel = labelMode === Label.Hidden;`

ketsappi:
If I currently try to use different placeholder than the `labelText` I get this warning:

> Placeholder different than label is not supported, using labelText instead

If the user is not able to use any other value than same as in `labelText`, is there really need to have the `placeholder` attribute for user to enter something own? The `placeholder` can just take the value straight from the `labelText` under the hood.

src/components/Form/TextInput.tsx
```diff
@@ -42,17 +42,17 @@ export interface TextInputProps extends Omit<HtmlInputProps, 'type'> {
   /** Label */
   labelText: string;
   /** Label displaymode
-   * default: show above, screenreader: show as placeholder
+   * visible: show above, hidden: show as placeholder
    */
```
ketsappi: Maybe add `@default 'visible'` to JSDoc as it seems to be default when not telling the `labelMode`

LJKaski:
> If the user is not able to use any other value than same as in `labelText`, is there really need to have the `placeholder` attribute for user to enter something own? The `placeholder` can just take the value straight from the `labelText` under the hood.

Good point, and that's something I though as well. Could be a better way to do it. It is like it is because it allows for visible label without a placeholder (not a very common use case though) and it let's the developer know what is being done under the hood. Any value given for placeholder results in it getting the value of the labelText, and the warn is to let the developer know. If we decide that the placeholder should always be there, then your solution sounds like the way to go.

src/components/Form/TextInput.tsx
```diff
@@ -67,21 +67,65 @@ export class TextInput extends Component<TextInputProps> {
...
+          <VisuallyHidden>{labelText}</VisuallyHidden>
+        ) : (
+          <Paragraph {...labelTextProps}>{labelText}</Paragraph>
+        )}
+        <HtmlDiv {...inputContainerProps}>
+          <HtmlInput
+            {...passProps}
+            className={classnames(inputBaseClassName, inputClassName)}
+            type="text"
+            placeholder={hideLabel ? labelText : !!placeholder ? labelText : ''}
```
ketsappi: Maybe instead of empty string, '', with the last option use `undefined`.

Because when it is now empty string, React seems to add the `placeholder` attribute to input anyway. And reading the code it seems that it is not wanted.

E.g what happened when the last option was happening:
`<input class="fi-text-input_input sc-dnqmqq yydqa" placeholder="" type="text">`

ketsappi:
> > If the user is not able to use any other value than same as in `labelText`, is there really need to have the `placeholder` attribute for user to enter something own? The `placeholder` can just take the value straight from the `labelText` under the hood.
> 
> Good point, and that's something I though as well. Could be a better way to do it. It is like it is because it allows for visible label without a placeholder (not a very common use case though) and it let's the developer know what is being done under the hood. Any value given for placeholder results in it getting the value of the labelText, and the warn is to let the developer know. If we decide that the placeholder should always be there, then your solution sounds like the way to go.

If it is not possible to change the actual text shown with the `placeholder` attribute, should it be like `showPlaceholder` boolean-attribute instead? Then you can still have the logic of visible label without a placeholder.

src/components/Form/TextInput.tsx
J-Kallunki: So should this be using the `BaseTextInput` component passing props and only changing those props that are needed to be different?

src/core/Form/SearchInput/SearchInput.md
```diff
@@ -4,5 +4,6 @@ import { SearchInput } from 'suomifi-ui-components';
 <SearchInput
   onBlur={event => console.log(event.target.value)}
   labelText="Search..."
+  labelMode="hidden"
```
J-Kallunki: Example with and without?
Can wrap multiple components here with `<></>`

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -3,47 +3,75 @@ import { default as styled } from 'styled-components';
...
+      }}
       {...passProps}
       className={classnames(className, baseClassName)}
       inputClassName={classnames(inputClassName, inputBaseClassName)}
-      inputContainerProps={{ className: inputContainerBaseClassName }}
     />
   ),
 )`
   ${props => baseStyles(props)}
+  ${props => inputBaseStyles(props)}
```
J-Kallunki: I think this should go first above the baseStyles given here (CSS definitions order overriding).

src/components/Form/TextInput.tsx
LJKaski: It was implemented like that at one point, but that would require figuring out a way to get the different placeholder visibility logic to the actual html component. Didn't get there yet. I'm open for suggestions.

src/core/Form/SearchInput/SearchInput.md
LJKaski: Had that when testing myself. Probably a good idea to have them both.

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -3,47 +3,75 @@ import { default as styled } from 'styled-components';
...
  * Use for user inputting search text
  */
-export class SearchInput extends Component<TextInputProps> {
+export class SearchInput extends Component<SearchInputProps> {
   render() {
+    const { labelText, placeholder } = this.props;
+
+    if (!!placeholder && labelText !== placeholder) {
+      logger.warn(
+        'Placeholder different than label is not supported, using labelText instead',
```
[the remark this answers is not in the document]
LJKaski: From what I understood the labelText is required, and in this implementation it gets used as placeholder if the placeholder has a value. Might be clearer to replace the placeholder prop with showPlaceholder boolean prop, so the logic would be easier to grasp.

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -3,47 +3,75 @@ import { default as styled } from 'styled-components';
...
 
 const baseClassName = 'fi-search-input';
 const inputContainerBaseClassName = `${baseClassName}_input-container`;
 const inputBaseClassName = `${baseClassName}_input`;
 const iconBaseClassName = `${baseClassName}_icon`;
 
-export interface SearchInputProps extends Omit<TextInputProps, 'variant'> {}
+export interface SearchInputProps extends Omit<TextInputProps, 'variant'> {
+  /** Placeholder text for search field. Must match labelText */
+  placeholder?: string;
```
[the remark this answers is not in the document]
LJKaski: I think previously it was so that if label was made visible the placeholder disappears. At the moment both can be visible at the same time if desired, but the placeholder gets its value from labelText in any case. As mentioned in another comment, a boolean might be a clearer way to implement this. Or we can have it as it was, with just the options for a hidden label with labelText as placeholder or a visible label without a placeholder.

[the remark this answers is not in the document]
J-Kallunki: Yeah I would suggest not allowing to set `placeholder`-prop and HTML placeholder-attribute is always from labelText if conditions met.

src/components/Form/TextInput.tsx
aappoalander: Might be worth checking how the placeholder behaves with IE since e.g. hidden boolean attribute was not properly added / removed when the value was changed. This might not be the case / might not matter with this component, but maybe worth checking still?

src/core/Form/SearchInput/__snapshots__/SearchInput.test.tsx.snap
aappoalander: Changing style import / adding order might affect things, but might as well have nothing to do with these...

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -19,31 +21,45 @@ const StyledTextInput = styled(
...
     inputClassName,
     ...passProps
   }: TextInputProps & InternalTokensProp) => (
-    <TextInput
+    <CompSearchInput
+      {...passProps}
+      labelTextProps={{
+        ...labelTextProps,
+        className: classnames(
+          labelTextProps.className,
```
aappoalander: Am I reading this wrong or is this set as undefined above? If so, could this line be removed?

J-Kallunki: Above uses spread defaults. It checks that if `labelTextProps` not defined - className is undefined.
It could also use just empty object as default.

PR #230, opened 2019-10-30
Install suomifi-design-tokens 0.2.1

J-Kallunki:
## Description

- Install `suomifi-design-tokens`
- Fix CSS-values to be `string | number | ValueUnit`
- Create utils for tokens using `suomifi-design-tokens` format
- Implement BodyTextBold-token internally
- Fix markdowns of presenting tokens at styleguidist

## Motivation and Context

Design tokens where separated to `suomifi-design-tokens`-package

## How Has This Been Tested?

`yarn validate`
Styleguidist selected pages


.styleguidist/typography.md
```diff
@@ -2,14 +2,14 @@ Suomifi-styleguide typography
...
-import {
-  typographyTokens,
-  fontFamily
-} from '../src/core/theme/typography';
+import { Text, Heading } from '../src';
+import { suomifiTheme } from '../src/core/theme';
+
 import clipboardCopy from 'clipboard-copy';
 
+const fontFamily = 'Source Sans Pro';
```
aappoalander: Is there a reason why this is used differently from what is defined in tokens? ```"'Source Sans Pro', 'Helvetica Neue', 'Arial', sans-serif"?```

src/core/Colors/Colors.baseStyles.tsx
```diff
@@ -41,7 +45,7 @@ export const baseStyles = withSuomifiTheme(
         }
 
         &--key {
-          ${theme.typography.bodySemiBold}
+          ${internalTypographyTokens.bodySemiBold}
```
aappoalander: I assume this is not yet implemented / added to theme, so we use an internal set instead?

src/core/Colors/Colors.tsx
```diff
@@ -32,33 +32,36 @@ export class Colors extends Component<ColorsProps> {
...
+            const hslaAsHex = hslaToHex(color.toString());
+            const item = (
+              <Color
+                keyName={key.toString()}
+                color={color.toString()}
+                key={key.toString()}
+                onClick={copyKey(key.toString())}
+                {...props}
+              >
+                <div className="fi-color__name">{color.toString()}</div>
```
aappoalander: What is the value type of color if not string? ```value.hsl``` from tokens should already be of type string?

src/core/Breadcrumb/BreadcrumbLink.tsx
```diff
@@ -43,7 +46,8 @@ export const BreadcrumbLink = ({
       <Icon
         icon="linkBreadcrumb"
         className={iconClassName}
-        color={withSuomifiDefaultProps(passProps).tokens.colors.blackBase}
+        color={colorValue(passProps)('blackBase')}
```
aappoalander: Would it be possible to use e.g. ```passprops.tokens.colors.blackBase.hsl``` instead of the colorValue utility function or would that break modularity somehow? What are the benefits of using a utility instead of directly accessing the values in components?

src/utils/css/utils.ts
```diff
@@ -9,21 +9,54 @@ export const clearfix = css`
   }
 `;
 
+/**
+ * Return CSS compatible string
+ * @param cssValue number, string or {value: number, unit: string | null}
+ */
+export const cssValueToString = (
```
aappoalander: Another option could be to just export all tokens as css ready-to-use from tokens repo. This way we could always directly use the tokens without having an extra layer of complexity. This should be handled separately though, in case feasible.

aappoalander:
Looks good overall even though the scope of the PR is quite large. Some things might benefit from refactoring the tokens repository exports, but that should be captured in additional issues if seen feasible.

src/core/theme/utils/typography.ts
```diff
@@ -1,20 +1,19 @@
...
 
 type TypographyTokensAsCssProp = {
-  [key in keyof TypographyTokens]: FlattenSimpleInterpolation
+  [key in keyof TypograhpyDesingTokens]: FlattenSimpleInterpolation
 };
 export interface TypographyUtil extends TypographyTokensAsCssProp {}
 export class TypographyUtil {
   static instance: TypographyUtil;
-  constructor(tokens: TypographyTokens) {
+  constructor(tokens: TypograhpyDesingTokens) {
```
ketsappi: Just a typo, but 'Typogra**hp**yDesi**ng**Tokens' should be 'Typogra**ph**yDesi**gn**Tokens'  instead?

Might be worth checking the other places also for 'Desi**ng**' vs 'Desi**gn**'

src/core/theme/utils/spacing.ts
```diff
@@ -1,19 +1,29 @@
...
+export const spacingUtils = (spacing: SpacingDesingTokens) =>
+  Object.entries(spacing).reduce(
+    (retObj, [key, value]) => ({
+      ...retObj,
+      [key]: cssValueToString(value),
+    }),
+    {} as { [key in keyof SpacingDesingTokens]: FlattenSimpleInterpolation },
+  );
+
+// TODO make this whole util again to work with suomifiTheme-util
```
ketsappi: Is this **TODO** comment still valid or is it already done and could be removed?

src/core/theme/utils/typography.ts
aappoalander: This issue comes from the tokens repo. Already fixed, but has to be taken into use in ui-components.

src/utils/css/utils.ts
aappoalander: As discussed, this is a topic for another issue once we have figured out what is actually needed.

src/core/Breadcrumb/BreadcrumbLink.tsx
aappoalander: It is also ok to keep this abstraction to create looser coupling with design-system-tokens repo.

Issue #228, opened 2019-10-29
Mouse cursor won't change when moving over a component that is disabled


ketsappi:
## The Problem

Mouse cursor won't change when moving over a component that is disabled.

## Proposed Solution

The components that have `disabled`-attribute should have the `cursor: not-allowed;` in the disabled-style.

### Of the existing components, the following should have this
- Button
- SearchInput
- TextInput
- Toggle

### These components currently haven't the `disabled`-attribute, but if it is added; then these should also have the change
- Dropdown
- Menu

## Additional Info

[cursor - CSS: Cascading Style Sheets | MDN - Mozilla](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor)

J-Kallunki:
Create `src/components/utils/css.ts`-file
and `disabledCursor` -util/mixin there with `css` (from styled components).

Use that mixin to `src/components` where disabled is used with:
```
styled((props: ...) => (
  ...
))`
  &.componentfoobar--disabled {
    ${disabledCursor}
  }
  `;
```

PR #249, opened 2019-11-13
[Feature] Mouse cursor change for disabled SearchInput & TextInput

ketsappi:
## Description

When hovering mouse over disabled SearchInput & TextInput, mouse cursor will now change to give better feedback to the user.

## Related Issue

Related #228 

## Motivation and Context

To give better feedback for the user when the component is in disabled state.

## How Has This Been Tested?
- `yarn test`
- `yarn validate`
- manually testing with browser

### DIFF
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -12,6 +12,8 @@ import { VisuallyHidden } from '../Visually-hidden/Visually-hidden';
 import { Paragraph, ParagraphProps } from '../Paragraph/Paragraph';
 import { logger } from '../../utils/logger';
 import classnames from 'classnames';
+import styled from 'styled-components';
+import { disabledCursor } from '../utils/css';
 
 const baseClassName = 'fi-text-input';
 const disabledClassName = `${baseClassName}--disabled`;
@@ -53,7 +55,7 @@ export interface TextInputProps extends Omit<HtmlInputProps, 'type'> {
   children?: ReactNode;
 }
 
-export class BaseTextInput extends Component<TextInputProps> {
+class BaseTextInput extends Component<TextInputProps> {
   render() {
     const {
       className,
@@ -94,6 +96,14 @@ export class BaseTextInput extends Component<TextInputProps> {
   }
 }
 
+export const StyledBaseTextInput = styled((props: TextInputProps) => (
+  <BaseTextInput {...props} />
+))`
+  &.${disabledClassName} {
+    ${disabledCursor}
+  }
+`;
+
 export class TextInput extends Component<TextInputProps> {
   render() {
     const { placeholder, ...passProps } = this.props;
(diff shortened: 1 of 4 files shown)


ketsappi:
> The solution looks and seems to work fine. The red cursor somehow felt quite offensive though. Would a grey one suffice or is the color a necessary part of the action and information conveyed? That way the color wouldn't jump out so much from the rest of the color scheme.

There shouldn't be any color adjustments for the cursor by this change. It is gray when tested with macOS with Chrome, Firefox, Safari and Opera.
Maybe it is because of Windows 10 and the default for that cursor?

src/components/Form/TextInput.tsx
```diff
@@ -94,6 +96,14 @@ export class BaseTextInput extends Component<TextInputProps> {
   }
 }
 
+export const StyledBaseTextInput = styled((props: TextInputProps) => (
```
LJKaski: Should these styles be here or simply in the TextInput/SearchInput basestyles? Are there even cases where one would want a disabled search input on the screen?

aappoalander: I was wondering the same. In addition, I'd avoid exporting something named with arbitrary Styled-prefix, because the styled does not have any particular meaning outside this context (why styled?).

ketsappi: > Should these styles be here or simply in the TextInput/SearchInput basestyles? 

Maybe they can be moved there if it possible; might need to change then also how it is done with the Button's one to keeping it constant. Currently by this way, one advantage is that there is no need to tell 'disabledClassName' again in the baseStyles (DRY).

---

> Are there even cases where one would want a disabled search input on the screen?

Maybe there is no use for disabled search, that is true. Maybe we should then remove this from the SearchInput.

---

>.. I'd avoid exporting something named with arbitrary Styled-prefix, because the styled does not have any particular meaning outside this context (why styled?).

**Currently** exporting it out with styled-prefix to keep consistent with other places when using styled-components and because it is used in the `SearchInput.tsx` also.

aappoalander: > Maybe there is no use for disabled search, that is true. Maybe we should then remove this from the SearchInput.
We cannot know whether there is usecase for this or not, so I'd keep the functionality.

> Currently exporting it out with styled-prefix to keep consistent with other places when using styled-components and because it is used in the SearchInput.tsx also.
Let's keep it consistent then. Are these also exported externally meaning not only within the library? Because like mentioned, word ```Styled``` does not make any sense in the context of library API... TwoHardThings?

J-Kallunki: > Are these also exported externally meaning not only within the library? Because like mentioned, word Styled does not make any sense in the context of library API... TwoHardThings?

Seems that it is not exported in any of the indexs'

Issue #250, opened 2019-11-13
Remove support for custom svg for Icon component


aappoalander:
## The Problem

The Icon component will support passing custom props to the underlying svg-element. Currently react-svg is used to support custom icons when an svg file source is provided (src -prop), but this library does not allow passing custom props to svg.

In addition supporting custom svgs with the Icon component could be considered out of scope for the ui-components library as all supported icons should be available using the name enumerable prop.

To keep things uniform support for custom svg's should be dropped.

## Proposed Solution

Remove support for src-prop for Icon component. Remove unnecessary components and resources related to this custom support.

## Alternative Solutions

An alternative solution would be to restrict custom props to what is supported by react-svg.

PR #252, opened 2019-11-13
[Feature] Icon component custom props

### LINKED ISSUE #156: Icons have issues in strict Content-Security-Policy environment
In order to work-around the style attribute issue explained in https://github.com/vrk-kpa/suomifi-icons/issues/15 I need to provide `className` for customization and set `color` attribute to either `null` or `{}` otherwise leaving it `undefined` it ends up with the default color set in https://github.com/vrk-kpa/suomifi-ui-components/blob/develop/src/core/Icon/Icon.tsx#L85

I think having to do:
```
<Icon color={{}} className="info" icon="infoFilled" mousePointer />
```
or
```
<Icon color={null} className="info" icon="infoFilled" mousePointer />
```

is not intended and should probably be addressed in https://github.com/vrk-kpa/suomifi-icons/issues/15

aappoalander:
## Description

This PR update suomifi-icons to version 0.0.10 and introduces support for Icon component custom props. The custom props are now passed forward to underlying svg.

In addition, includes a fix to restore support for custom icon colors.

## Related Issue

Closes #156 
Relates to #250 

## Motivation and Context

This update adds support for strict CSP when using Icons and setting the colors.

## How Has This Been Tested?

Tested by creating a new local package using Verdaccio and with create react app typescript application with strict CSP.

### DIFF
--- a/src/core/Icon/Icon.tsx
+++ b/src/core/Icon/Icon.tsx
@@ -84,24 +84,15 @@ export class Icon extends Component<IconProps> {
     } = withSuomifiDefaultProps(this.props);
     const { className, ariaLabel } = this.props;
 
+    const iconColor =
+      color !== undefined ? color : colorValue({ tokens })('depthDark27');
+
     if (!!src) {
-      return (
-        <StyledIcon
-          src={src}
-          {...passProps}
-          color={colorValue({ tokens })('depthDark27')}
-        />
-      );
+      return <StyledIcon src={src} {...passProps} color={iconColor} />;
     }
 
     if (icon !== undefined) {
-      return (
-        <StyledSuomifiIcon
-          {...passProps}
-          icon={icon}
-          color={colorValue({ tokens })('depthDark27')}
-        />
-      );
+      return <StyledSuomifiIcon {...passProps} icon={icon} color={iconColor} />;
     }
 
     logger.warn(
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "react-svg": "7.2.2",
     "styled-components": "4.3.2",
     "suomifi-design-tokens": "0.3.0",
-    "suomifi-icons": "0.0.8",
+    "suomifi-icons": "0.0.10",
     "uuid": "3.3.2"
   },
   "peerDependencies": {
--- a/src/components/Svg/Svg.tsx
+++ b/src/components/Svg/Svg.tsx
@@ -19,6 +19,8 @@ export interface SvgProps {
   /** Show mouse cursor as hand-pointer */
   mousePointer?: boolean;
   testId?: string;
+  /** Allow passing custom attributes */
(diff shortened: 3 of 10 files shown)


src/core/Icon/Icon.tsx
```diff
@@ -84,24 +84,15 @@ export class Icon extends Component<IconProps> {
     } = withSuomifiDefaultProps(this.props);
     const { className, ariaLabel } = this.props;
 
+    const iconColor =
```
aappoalander: This does not take style props into account, color might be defined there as well.

ketsappi: It seems that user is not able to give custom color for icons that starts with: `staticIllustration`.

Should the illustration versions support of changing the "main color"? Like currently they have that orange color.

aappoalander: I think the colors for static icons are fixed inside the SVG's and are not supposed to be changed. The only thing worth supporting for those that comes to mind is "dark mode", but this is not supported by anything we currently have.

aappoalander: I do agree, that the API is kind of confusing since it's possible to define a color that does not have any effect on the result.

src/core/Icon/Icon.tsx
```diff
@@ -84,24 +84,15 @@ export class Icon extends Component<IconProps> {
...
-        <StyledSuomifiIcon
-          {...passProps}
-          icon={icon}
-          color={colorValue({ tokens })('depthDark27')}
-        />
-      );
+      return <StyledSuomifiIcon {...passProps} icon={icon} color={iconColor} />;
     }
 
     logger.warn(
```
LJKaski: Not related to these changes, but I remember the policy being that we don't want nested ternary operators.

aappoalander: I guess so, If you want to, you could try to write this in a better way?

LJKaski: The easiest way would probably be to just pre-check a part of it and use that in the operator.
Something along the lines of:

const classNameErrorText = !!className ?  " with className: {className}" : '' ;

      `Icon ERROR${
        !!ariaLabel
          ? ` with aria-label: ${ariaLabel}`
          : classNameErrorText
      }`

aappoalander: To write this in a different way would require for example a mutable data structure, a separate function with returns, or an if-elseif-else structure. => this format is shorter cleaner and easier to read. Won't change it unless explicit linter rules are applied to force this.

src/core/Icon/Icon.tsx
aappoalander: @J-Kallunki any thoughts on this?

aappoalander: As agreed while discussing the issue, the API will remain as is for now.

PR #256, opened 2019-11-15
[Feature] Mouse cursor change for disabled Toggle

ketsappi:
## Description

When hovering mouse over disabled **Toggle**, mouse cursor will now change to give better feedback to the user.

## Related Issue

Related #228

## Motivation and Context

To give better feedback for the user when the component is in disabled state.

## How Has This Been Tested?
- `yarn test`
- `yarn validate`
- manually testing with browser


ketsappi:
> Just that someone test that the cursor works also for the icon with Firefox/Edge/IE

Tested now with Win10 Firefox and Edge and seems to work fine.
_(Not able to test with IE11 as it has problems to show the whole styleguidist)_

PR #266, opened 2019-11-26
[Fix] Color token presentation accessibility

### LINKED ISSUE #173: Fix Color-token presentation accessibility

Use `figure` and code-generated `svg`.
https://codepen.io/tigt/post/semantic-html-for-colors#are-you-kidding-me-safari-0

https://github.com/vrk-kpa/suomifi-ui-components/blob/develop/src/core/Colors/Colors.tsx#L49

https://vrk-kpa.github.io/suomifi-ui-components/#/Introduction/Colors

## Specifications

- Version: 0.2.6

ketsappi:
## Description

- Using `figure` and `svg` to show the color token
- Able to use keyboard to switch between color tokens with `Tab`-key
- Able to use keyboard to copy the key of currently selected color token with `Enter`-key
- Added snapshot test to see possible structure changes in the future

## Related Issue

Closes #173 

## Motivation and Context

Change increases accessibility of the colors by changing the semantics and adding support for keyboard actions.

## How Has This Been Tested?
- manually in the browser (macOS 10.15.1:  Firefox, Chrome, Safari, Opera)
- `yarn validate`

### DIFF
--- a/src/core/Colors/Colors.baseStyles.tsx
+++ b/src/core/Colors/Colors.baseStyles.tsx
@@ -17,18 +17,37 @@ export const baseStyles = withSuomifiTheme(
       flex-direction: column;
       justify-content: flex-end;
       float: left;
-      width: 192px;
-      height: 180px;
+      width: 208px;
+      height: 196px;
       max-width: 100%;
-      padding: ${theme.spacing.s};
       color: ${readableColor(color)};
-      background-color: ${color};
       border-bottom: 4px solid ${color};
       cursor: pointer;
       z-index: 2;
+      position: relative;
+      margin: 0;
 
       &:hover {
         border-bottom-color: ${readableColor(color)};
+      }
+
+      figcaption {
+        display: flex;
+        flex-direction: column;
+        justify-content: flex-end;
+        position: absolute;
+        left: 8px;
+        bottom: 8px;
+      }
+
+      svg {
+        width: 100%;
+        height: 100%;
+
+        rect {
+          width: 100%;
+          height: 100%;
+        }
       }
 
       .fi-color_name {
--- a/src/core/Colors/Colors.baseStyles.tsx
+++ b/src/core/Colors/Colors.baseStyles.tsx
@@ -36,8 +36,8 @@ export const baseStyles = withSuomifiTheme(
         flex-direction: column;
         justify-content: flex-end;
         position: absolute;
-        left: 8px;
-        bottom: 8px;
+        left: ${theme.spacing.s};
+        bottom: ${theme.spacing.s};
       }
 
       svg {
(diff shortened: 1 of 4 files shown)


src/core/Colors/Colors.baseStyles.tsx
```diff
@@ -17,20 +17,39 @@ export const baseStyles = withSuomifiTheme(
       flex-direction: column;
       justify-content: flex-end;
       float: left;
-      width: 192px;
-      height: 180px;
+      width: 208px;
```
aappoalander: Where do these magic numbers come from? Looks like these were just updated, but this kind of arbitrary looking styles based on absolute values would benefit from comments in my opinion. These are used with styleguidist?

src/core/Colors/Colors.baseStyles.tsx
```diff
@@ -17,20 +17,39 @@ export const baseStyles = withSuomifiTheme(
...
         border-bottom-color: ${readableColor(color)};
       }
 
+      figcaption {
+        display: flex;
+        flex-direction: column;
+        justify-content: flex-end;
+        position: absolute;
+        left: 8px;
+        bottom: 8px;
```
aappoalander: Do we have design tokens that would suit this usecase?

ketsappi: You mean the `left` and `bottom` values?

Like this?
```scss
left: ${theme.spacing.s};
bottom: ${theme.spacing.s};
```

aappoalander: Yes, not sure if it's relevant here, but I like this better since those are defined.

ketsappi: This change was made and committed.

src/core/Colors/Colors.baseStyles.tsx
ketsappi: These values for `width` and `height` was updated to keep the visual presentation, in the styleguidist, like it was previously; looking square.

Possibly later this could be done dynamically based on available size, but at this point this it not worth the effort.

LJKaski: If something was done regarding the visual presentation, I think a screenshot or two would make the review a lot easier, even if they are just to show that things stayed the same, like in this case.

ketsappi: I didn't add any screenshots for the PR as there were no visual changes needed to be done with this fix. And change to these values here were done only to keep the same appearance as before.

But if there will be later changes that will change the color tokens presentation to be more dynamic, screenshots would be very much needed to show the difference eg. in some example cases (_different browser width_).

PR #273, opened 2019-12-27
[Enhancement] Added Text Variant type for text component

### LINKED ISSUE #270: TextVariant for Text-component
## The Problem

Not being consistent in codebase.

https://github.com/vrk-kpa/suomifi-ui-components/blob/ed1e6a9863443dff08e3815d687d613523380916/src/core/Text/Text.tsx#L24

Following components have their variants in similar case:

- Breadcrumb
- Button
- Link
- Panel
- TextInput

## Proposed Solution

`type TextVariant = 'body' | 'lead' | 'bold';`

`variant?: TextVariant;`

saksham93:
Signed-off-by: sakshamgupta93 <saksham.gupta@nagarro.com>

Added Text Variant type for text component

Related to #270 

Closes #270 

Making the codebase consistent for the components

### DIFF
--- a/src/core/Text/Text.tsx
+++ b/src/core/Text/Text.tsx
@@ -12,6 +12,8 @@ import classnames from 'classnames';
 const baseClassName = 'fi-text';
 const smallScreenClassName = `${baseClassName}--small-screen`;
 
+type TextVariant = 'body' | 'lead' | 'bold';
+
 export interface TextProps extends CompTextProps, TokensProp {
   /** Change font to smaller screen size and style */
   smallScreen?: boolean;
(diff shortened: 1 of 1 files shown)


saksham93:
@J-Kallunki @riittap @RosaHilden @SeppoRanninranta - Could you please look into this PR if this is ok? Thank you.

aappoalander:
@saksham93 Thank you for your contribution and sorry for the late reply. Our code base seems to be a bit inconsistent with handling types for variants. In most of the cases these types are not used anywhere outside the components, so we chose to take the opposite direction and not explicitly declare types for variants for these interfaces. Some that already have these types unnecessarily declared should be fixed later on. Because of this we will be closing this PR.

PR #276, opened 2020-01-10
[BreakingChange] Button variant 'negative' changed to 'inverted'

ketsappi:
## Description

Button variant 'negative' changed to 'inverted' to match design. 

This is a breaking change, so users should use 'inverted' instead of previous 'negative' after this change goes live.

**Before**: `<Button.negative>Button.negative</Button.negative>`
**Now**: ` <Button.inverted>Button.inverted</Button.inverted>`

## Motivation and Context

Naming of the variant changed to reflect the design to prevent any possibility of miscommunication or confusion.

## How Has This Been Tested?

- Run locally in the browser
- `yarn validate`

### DIFF
--- a/src/core/Button/Button.md
+++ b/src/core/Button/Button.md
@@ -36,11 +36,11 @@ import { Button } from 'suomifi-ui-components';
 
 <>
   <div example="negative">
-    <Button.negative>Button.negative</Button.negative>
+    <Button.inverted>Button.inverted</Button.inverted>
 
-    <Button.negative disabled fullWidth icon="login">
-      Button.negative disabled fullWidth icon="login"
-    </Button.negative>
+    <Button.inverted disabled fullWidth icon="login">
+      Button.inverted disabled fullWidth icon="login"
+    </Button.inverted>
   </div>
 </>;
 ```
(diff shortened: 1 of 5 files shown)


src/core/Button/Button.md
```diff
@@ -36,11 +36,11 @@ import { Button } from 'suomifi-ui-components';
 
 <>
   <div example="negative">
```
ketsappi: This 'negative' was left as it was because in general it indicates just negative version for background colouring for example shown in styleguidist.

It is used in the styleguidist styles:
https://github.com/vrk-kpa/suomifi-ui-components/blob/0876abc22ee2c2b73cb0ca16c2414e4ac0777793/.styleguidist/styleguidist.styles.js#L98

--- 

@LJKaski What are your thoughts about these?
Should we also rename these to `inverted`, even though they are not really the Button-variant? 

Maybe it would make sense, as the `negative` in this case is like the opposite to white background on styleguidist and maybe the `inverted` is more describing than the `negative`?

LJKaski: I'd keep the naming consistent and name these inverted as well.

ketsappi: This change is now committed.

PR #277, opened 2020-01-15
[BreakingChange] Rename Menu to LanguageMenu

ketsappi:
## Description

- Renamed the Menu component to LanguageMenu
- Removed the 'language' variant from the LanguageMenu

## Motivation and Context

Renamed to match component name with the design.

## How Has This Been Tested?

- manually in the browser


src/core/LanguageMenu/LanguageMenu.md
```diff
@@ -0,0 +1,28 @@
...
+  LanguageMenu,
+  LanguageMenuItem,
+  LanguageMenuLink
+} from 'suomifi-ui-components';
+
+<LanguageMenu className="language-menu-language-test" name="FI">
+  <LanguageMenuItem onSelect={() => console.log('FI')} selected>
+    Suomeksi (FI)
+  </LanguageMenuItem>
+  <LanguageMenuLink href="/sv">På svenska (SV)</LanguageMenuLink>
```
LJKaski: Would it be clear enough without the href, since the link actually works but doesn't lead anywhere relevant?

LJKaski: let's leave it for now. It demonstrates the use well and it should be quite clear to our demographic that links in the demo are not meant to work.

PR #278, opened 2020-01-16
[BreakingChange]Rename Panel to Expander

ketsappi:
## Description

- PanelExpansion renamed to Expander
- PanelExpansionGroup renamed to ExpanderGroup
- Existing Panel features moved to Expander
- Panel removed

## Motivation and Context

Change was made to match naming and other parts with the design system.

## How Has This Been Tested?
- manually in the browser
- `yarn test`
- `yarn run test:lint`
- `yarn run tsc --noEmit`


src/components/Expander/Expander.tsx
```diff
@@ -47,11 +62,11 @@ const StyledPanelExpansionTitle = styled(
   ${allStates('cursor: pointer;')}
 `;
 
-interface PanelExpansionState {
+interface ExpanderState {
   openState: boolean;
 }
 
-export interface PanelExpansionProps extends PanelProps {
+export interface ExpanderProps extends CommonExpanderProps {
```
aappoalander: ExpanderBaseProps? BaseExpanderProps? SharedExpanderProps? Common is not bad either, just make the decision count since it will be difficult to change it afterwards.

src/components/Expander/ExpanderGroup.tsx
```diff
@@ -18,7 +18,7 @@ export const StyledPanel = styled((props: HtmlDivProps) => (
   max-width: 100%;
 `;
 
-interface PanelExpansionGroupState {
+interface ExpanderGroupState {
   /** Panels that are open */
   openPanels: number[];
```
LJKaski: Some of these still speak of panels instead of expanders. The most obvious ones can be spotted from the styleguidist prop descriptions.

src/components/Expander/Expander.tsx
ketsappi: I made the change and it is now `SharedExpanderProps` that sounds better as `CommonExpanderProps` sounded... too common.

Issue #226, opened 2019-10-28
SSR-support


J-Kallunki:
## Expected Behavior or Possible Solution

Test and fix SSR-support based on Styled Components 
https://www.styled-components.com/docs/advanced#server-side-rendering

## Actual Behavior and Screenshots

Related Issues:
#202 

## Steps to Reproduce the Problem

create minimal SSR setup

## Specifications

- Version: 0.2.6

### Environment: (if related to the issue)

Issue #202, opened 2019-10-09
installation fails


halonki-cinia:
## Steps to Reproduce the Problem

- install library with `yarn add suomifi-ui-components`
- you get following error messages
```
[11:50:33 serveFrontend] TS7016: Could not find a declaration file for module 'styled-components'. '/home/halonki-cinia/vrk/sevi-vespa/sevi-ui/node_modules/styled-components/dist/styled-components.cjs.js' implicitly has an 'any' type.
[11:50:33 serveFrontend]   Try `npm install @types/styled-components` if it exists or add a new declaration (.d.ts) file containing `declare module 'styled-components';`
```
- install additional types
`npm install @types/styled-components @types/reach__menu-button --save-dev`
- you get an error `Cannot find name 'SvgrComponent'`
and bunch of these kind of errors:
```
[13:06:29 serveFrontend] ERROR in /home/halonki-cinia/vrk/sevi-vespa/sevi-ui/node_modules/suomifi-ui-components/dist/es2015/core/Heading/Heading.d.ts(17,43):
[13:06:29 serveFrontend] TS2344: Type '"content" | "results" | "form" | "lang" | "onError" | "key" | "children" | "cite" | "data" | "label" | "span" | "style" | "summary" | "title" | "pattern" | "defaultChecked" | "defaultValue" | ... 344 more ... | "smallScreen"' does not satisfy the constraint '"content" | "results" | "form" | "lang" | "onError" | "key" | "children" | "cite" | "data" | "label" | "span" | "style" | "summary" | "title" | "pattern" | "defaultChecked" | "defaultValue" | ... 343 more ... | "smallScreen"'.
[13:06:29 serveFrontend]   Type '"onAuxClick"' is not assignable to type '"content" | "results" | "form" | "lang" | "onError" | "key" | "children" | "cite" | "data" | "label" | "span" | "style" | "summary" | "title" | "pattern" | "defaultChecked" | "defaultValue" | ... 343 more ... | "smallScreen"'.
```
```
[13:06:29 serveFrontend] ERROR in /home/halonki-cinia/vrk/sevi-vespa/sevi-ui/node_modules/typescript/lib/lib.dom.d.ts(5292,11):
[13:06:29 serveFrontend] TS2300: Duplicate identifier 'FormData'.
```

## Specifications

- Version: 0.2.6

### Environment: (if related to the issue)

- Library not working with? 
   Suomi.fi service
- CRA/Gatsby/Next.js version?
[...]

Issue #150, opened 2019-09-01
Use Prettier-Eslint?


J-Kallunki:
https://github.com/prettier/prettier-eslint

PR #280, opened 2020-02-11
[Feature] Changed to use tsdx, tslint changed to eslint and others

ketsappi:
## Description

- Changed the build system to use [tsdx](https://github.com/jaredpalmer/tsdx)
  - tsdx is built on top of rollup
- `styled-components` is now a peer dependency and needs to be installed by the user of this library
  - Also telling about this and the requirement of the `react` in the README's **Install**-section
- tslint changed to eslint
  - some changes done based on new rules / adjustments
- Dropped the usage instructions of bare accessible components mentioning from the README as it is not correctly working as intended
  - this will be added back later on, when it is working again 
- updated some dependencies with newer version
- changed some dependencies to new version as old one was deprecated

## Related Issue

Relates to following issues: #226, #202 & #150 

## Motivation and Context

- Making it simpler to build the library
  - tsdx will create all the needed builds
- Getting rid of old tslint
- Steps taken to support SSR; for what we are able to do on the library's side

## How Has This Been Tested?

- Locally in various SSR-template projects
- `yarn validate`


.eslintrc
```diff
@@ -1,9 +1,76 @@
...
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-empty-interface": [
+      "warn",
+      {
+        "allowSingleExtends": true
+      }
+    ],
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-inferrable-types": [
+      "warn",
```
aappoalander: Not sure if it's a good idea to use warnings since those might not work in Circle CI. At least worth checking before merging this setup.

package.json
```diff
@@ -1,11 +1,10 @@
 {
   "name": "suomifi-ui-components",
-  "version": "0.4.1",
+  "version": "0.5.0-alpha.1",
```
aappoalander: Is this a good idea to change the version with this PR? I'd propose to keep it as is and make a separate PR and a tag for 0.5.0.

src/components/Form/ToggleButton.tsx
```diff
@@ -55,7 +55,7 @@ export class ToggleButton extends Component<ToggleProps> {
 
     if (!!toggleInputProps || !!toggleInputComponent) {
       logger.error(
-        `ToggleButton does not utilize 'toggleInputProps' and 'toggleInputComponent' props.`,
+        'ToggleButton does not utilize toggleInputProps and toggleInputComponent props.',
```
aappoalander: Escaping characters should be possible here? \'

src/core/theme/index.ts
```diff
@@ -58,13 +62,13 @@ const internalTokens = {
...
-  colors,
-  spacing,
+  colors: colorTokens,
+  spacing: spacingTokens,
   typography: suomifiDesignTokens.typography,
   values: suomifiDesignTokens.values,
 };
 
-export const defaultTokens = {
+export const defaultThemeTokens = {
```
aappoalander: Is this exported outside the library / part of API? If it is, then I'd try to keep it as it was.

package.json
ketsappi: Good point. Didn't notice the version number increase that we made for the alpha release when making PR.

Let's do like you suggested; keeping it at the previous `0.4.1` in this PR and increase it in the PR of `release/v0.5.0` little later.

src/components/Form/ToggleButton.tsx
ketsappi: I changed this when looking through the warning given by eslint and thought that there might be no reason to use single quotes around the props to give anything more to the reader of warning message.

This current eslint rule allows also the previous version. 
https://github.com/vrk-kpa/suomifi-ui-components/blob/dca31e6cc60c34523004d1c599ee5102cfba084d/.eslintrc#L100

Maybe we can return this highlighted code to its previous value if it is giving more to the reader?

src/core/theme/index.ts
ketsappi: I don't see that it is used outside the library. Here is the two places where it is used:

---

1.) 

https://github.com/vrk-kpa/suomifi-ui-components/blob/dca31e6cc60c34523004d1c599ee5102cfba084d/src/core/theme/utils/defaults.ts#L11-L12

---

2.) 

https://github.com/vrk-kpa/suomifi-ui-components/blob/dca31e6cc60c34523004d1c599ee5102cfba084d/src/core/theme/tokens.test.tsx#L8

.eslintrc
ketsappi: So, should we use errors instead of warnings?
In this particular rule or in all rules? 

Ideally there is no warnings at all, so basically we could use just errors 🤔

aappoalander: Does the CI pipeline pass with these settings?

aappoalander: Seems to work so this is ok for now in my opinion!

src/components/Form/ToggleButton.tsx
ketsappi: Reverted to the previous value in latest commit.

PR #284, opened 2020-02-20
[Bugfix]Toggle focus

ketsappi:
## Description

- Fixes the focus outline of Toggle for both, the input- and button-version
- Fixes the disappearing focus issue when toggling the state of input-version
- Updated snapshot test accordingly

## Background info
Here are some additional info to support the changes made.

### checked attribute

> If present on a checkbox type, it indicates that the checkbox is checked by default (when the page loads). It does not indicate whether this checkbox is currently checked: if the checkbox’s state is changed, this content attribute does not reflect the change.

Source: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefchecked

Therefore the `key={String(toggleState)}` was removed from the ToggleInput.

### Space/Enter will not read changed state

Using `Space` or `Enter` will not read the state on **Safari** when toggling the state of the ToggleButton. _(On Chrome, it will read the state also with Space and Enter)_
But as the VoiceOver guides, using `Ctrl+Opt+Space`, the state is read nicely when toggling. 

Here are two examples which also works like this with VoiceOver.
- [Toggle Button example (Mute Audio)](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role)
- [Toggle Button (switch)](https://scottaohara.github.io/a11y_styled_form_controls/src/toggle-button-switch/)

So this should be okay as it currently is after this PR. Will be go through accessibility clinic, but might be not possible to change/force the VoiceOver action after all.

## How Has This Been Tested?
- `yarn validate`
- Tested with VoiceOver on macOS Safari
- Tested with VoiceOver on iOS Safari

**EDIT**: _Added information how this works different on Safari vs Chrome_

### DIFF
--- a/src/core/Form/Toggle/Toggle.baseStyles.tsx
+++ b/src/core/Form/Toggle/Toggle.baseStyles.tsx
@@ -25,9 +25,18 @@ export const baseStyles = withSuomifiTheme(
         }
       }
     }
-  & > .fi-toggle_label {
-    cursor: pointer;
-  }
+    & + .fi-toggle--with-input {
+      &:focus-within {
+        outline: 0;
+        ${focus({ theme, noPseudo: true })}
+      }
+      &:focus-within:not(:focus-visible) {
+        outline: 0;
+        &:after {
+          content: none;
+        }
+      }
+    }
   & > .fi-toggle_input {
     ${element({ theme })}
     ${font({ theme })('bodyText')}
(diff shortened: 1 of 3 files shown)


src/core/Form/Toggle/Toggle.baseStyles.tsx
```diff
@@ -13,9 +13,30 @@ export const baseStyles = withSuomifiTheme(
   ${element({ theme })}
   ${font({ theme })('bodyText')}
   background-color: ${theme.colors.whiteBase};
-  & > .fi-toggle_label {
-    cursor: pointer;
-  }
+    & + .fi-toggle--with-button {
```
aappoalander: Are these changes a bug fix to something that was broken earlier?

ketsappi: If you mean the `& > .fi-toggle_label {` it was removed as it has no effect after the structure change of input-version. 
The cursor is still correct, `pointer` so didn't see need to add separate rule to set it to `pointer` again.

aappoalander: And this was tested using different browsers?

ketsappi: On desktop tested with Safari and Chrome. And Safari on mobile.

_Additional note, VoiceOver will read the state change with Space and Enter on desktop Chrome but not with Safari._

PR #286, opened 2020-02-26
Feature/update reach

aappoalander:
## Update reach menu-button version and remove component-component

This PR updates reach menu-button version to latest available and adjusts implementation to match the new API. In addition, this PR removes Reach component-component. Also adds Reach popover as dependency to support Dropdown width alignment with menu-button.

## Motivation and Context

Reach menu-button did not work properly after upgrading other dependencies and had to be updated. Component-component was no longer used and had to be removed.

## How Has This Been Tested?

These changes have been tested using styleguidist, Create React App with JS and TS and two separate React SSR templates.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -38,6 +38,18 @@ declare module '@reach/menu-button' {
 
   export const MenuList: React.SFC<MenuListProps>;
 
+  export type Position = (
+    targetRect?: PRect | null,
+    popoverRect?: PRect | null,
+  ) => React.CSSProperties;
+
+  export type MenuPopoverProps = React.HTMLAttributes<HTMLDivElement> & {
+    children: React.ReactNode;
+    position?: Position;
+  };
+
+  export const MenuPopover: React.SFC<MenuPopoverProps>;
+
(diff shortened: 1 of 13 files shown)


types/index.d.ts
```diff
@@ -38,6 +38,18 @@ declare module '@reach/menu-button' {
...
+    popoverRect?: PRect | null,
+  ) => React.CSSProperties;
+
+  export type MenuPopoverProps = React.HTMLAttributes<HTMLDivElement> & {
+    children: React.ReactNode;
+    position?: Position;
+  };
+
+  export const MenuPopover: React.SFC<MenuPopoverProps>;
+
```
ketsappi: Was these some custom types or are those just missing from the the existing [@types/reach__menu-button](https://www.npmjs.com/package/@types/reach__menu-button) _(seems to be 8 months old...)_ ?

aappoalander: These are copied from the library. We do have some custom stuff here in the reach module definition and I'm going to create a following ticket to ditch this completely if possible. Just did not want to mix imported reach types with our custom definition due to conflict risk.

PR #294, opened 2020-03-09
[Breakinghange] Update library to use design tokens version 0.7.1

LJKaski:
## Description

Updated the library to use the latest, 0.7.1 release of the suomifi-design-tokens. Update came with some name changes to color tokens, which is a breaking change. Implemented the new token names to the component library.

## Motivation and Context

Designs changed and we want to make all the breaking changes we can before the 1.0 release.

## How Has This Been Tested?

Running locally and yarn test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,7 +111,7 @@
     "normalize.cssinjs": "1.1.0",
     "polished": "3.4.0",
     "react-svg": "7.2.2",
-    "suomifi-design-tokens": "0.6.0",
+    "suomifi-design-tokens": "0.7.1",
     "suomifi-icons": "0.1.0",
     "uuid": "3.3.2"
   },
--- a/src/core/theme/colors.ts
+++ b/src/core/theme/colors.ts
@@ -113,7 +113,7 @@ export const gradients = {
     colors.whiteBase,
   )} 0%, ${alphaHex(0)(colors.whiteBase)} 100%)`,
   depthLight2: `linear-gradient(0deg, ${colors.depthLight2} 0%, ${colors.whiteBase} 100%)`,
-  highlightLight2TohighlightLight3: `linear-gradient(0deg, ${colors.highlightLight2}, ${colors.highlightLight3})`,
+  highlightLight2ToHighlightLight3: `linear-gradient(0deg, ${colors.highlightLight2}, ${colors.highlightLight3})`,
 };
 
 export const outlines = {
--- a/src/core/theme/colors.ts
+++ b/src/core/theme/colors.ts
@@ -11,65 +11,65 @@ export const { colors } = suomifiDesignTokens;
 const {
   whiteBase,
   blackBase,
-  blackLighten42,
+  blackLight1,
   brandBase,
-  depthDark27,
+  depthDark1,
   depthBase,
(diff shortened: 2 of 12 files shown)


src/core/theme/colors.ts
```diff
@@ -112,8 +112,8 @@ export const gradients = {
   whiteBaseNegative: `linear-gradient(-180deg, ${alphaHex(0.1)(
     colors.whiteBase,
   )} 0%, ${alphaHex(0)(colors.whiteBase)} 100%)`,
-  depthLight26: `linear-gradient(0deg, ${colors.depthLight26} 0%, ${colors.whiteBase} 100%)`,
-  highlightLight45ToHighlightLight50: `linear-gradient(0deg, ${colors.highlightLight45}, ${colors.highlightLight50})`,
+  depthLight2: `linear-gradient(0deg, ${colors.depthLight2} 0%, ${colors.whiteBase} 100%)`,
+  highlightLight2TohighlightLight3: `linear-gradient(0deg, ${colors.highlightLight2}, ${colors.highlightLight3})`,
```
ketsappi: If following the previous naming highlightLight2To**h**ighlightLight3 should be highlightLight2To**H**ighlightLight3.

package.json
```diff
@@ -111,7 +111,7 @@
     "normalize.cssinjs": "1.1.0",
     "polished": "3.4.0",
     "react-svg": "7.2.2",
-    "suomifi-design-tokens": "0.6.0",
+    "suomifi-design-tokens": "0.7.1",
```
aappoalander: Why go from 0.6.0 to 0.7.1? 0.7.0 should be sufficient?

LJKaski: 0.7.1 had one more name change that was due to erroneous documentation for 0.7.0.

src/core/theme/colors.ts
aappoalander: One additional letter should be capital.

package.json
aappoalander: Ah, this branch is not rebased with the current develop, so it shows the diff wrong (current develop is 0.7.0, not 0.6.1). Might be worth rebasing even though there are no conflicts.

This should then be 0.8.0 because of breaking changes.

aappoalander: And the version raise should actually be with a separate PR...

ketsappi: Might be problematic to update version with the changes PR as then it depends in which order they are then able to merged / or other PR's version need to be adjusted if something is merged before it etc.

I would suggest to continue to increase version only on separate PR.

LJKaski: Are you looking at the right version number? That 0.7.1 is the version of the tokens. I did not touch the ui-components version number which is still 0.7.0.

And yes, it is rebased :)

LJKaski: The next ui-components release should be 0.8.0.

aappoalander: 😂 My bad! Was indeed looking at wrong version number. So no issues with dependencies, a separate PR is required anyways to increase the minor version for ui-comps.

src/core/theme/colors.ts
LJKaski: Fixed.

PR #295, opened 2020-03-09
[Feature] TextInput state text & colors

ketsappi:
## Description

- Adds new prop `statusText` for the TextInput
  - it is shown under the text field
  - will not reserve space when the prop is given; space is taken only when the prop actually has some value ( !== undefined)
- TextInput text color won't anymore change based on variant as per design
- TextInput will now have the `status`-prop exposed publicly (previously `variant`)
  - status = 'default' | 'error' | 'success'
- Added example to styleguidist how to use new `statusText` in conjunction with `status` prop
- The possibility to use the static ones (TextInput.success / TextInput.error) were removed

## Motivation and Context

Changes made so that it will match the design and able to show the statusText for giving more context if needed.

## How Has This Been Tested?
- Locally on the browser
- `yarn validate`

### DIFF
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -20,8 +20,8 @@ const baseClassName = 'fi-text-input';
 const disabledClassName = `${baseClassName}--disabled`;
 const labelBaseClassName = `${baseClassName}_label`;
 const inputBaseClassName = `${baseClassName}_input`;
-const stateTextClassName = `${baseClassName}_stateText`;
-const stateTextLabelClassName = `${stateTextClassName}_label`;
+const statusTextClassName = `${baseClassName}_statusText`;
+const statusTextLabelClassName = `${statusTextClassName}_label`;
 
 export interface TextInputLabelProps extends HtmlLabelProps {}
 
@@ -56,8 +56,8 @@ export interface TextInputProps extends Omit<HtmlInputProps, 'type'> {
   /** Input container div to define custom styling */
   inputContainerProps?: HtmlDivProps;
   children?: ReactNode;
-  /** Showing the state text; like validation error beneath the component */
-  stateText?: string;
+  /** Showing the status text; like validation error beneath the component */
+  statusText?: string;
 }
 
 class BaseTextInput extends Component<TextInputProps> {
@@ -71,13 +71,13 @@ class BaseTextInput extends Component<TextInputProps> {
       labelTextProps,
       inputContainerProps,
       children,
-      stateText,
+      statusText,
       id: propId,
       ...passProps
     } = this.props;
 
     const hideLabel = labelMode === 'hidden';
-    const generatedId = idGenerator(`${propId}-stateText`);
+    const generatedId = idGenerator(`${propId}-statusText`);
 
     return (
       <HtmlLabel
@@ -91,7 +91,7 @@ class BaseTextInput extends Component<TextInputProps> {
         ) : (
           <Paragraph {...labelTextProps}>{labelText}</Paragraph>
         )}
-        <HtmlDiv className={stateTextClassName}>
+        <HtmlDiv className={statusTextClassName}>
           <HtmlDiv {...inputContainerProps}>
             <HtmlInput
               id={propId}
@@ -102,8 +102,8 @@ class BaseTextInput extends Component<TextInputProps> {
             />
             {children}
           </HtmlDiv>
-          <HtmlLabel className={stateTextLabelClassName} id={generatedId}>
-            {stateText}
+          <HtmlLabel className={statusTextLabelClassName} id={generatedId}>
+            {statusText}
           </HtmlLabel>
         </HtmlDiv>
       </HtmlLabel>
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -7,6 +7,7 @@ import {
   HtmlInputProps,
   HtmlDiv,
   HtmlDivProps,
+  HtmlSpan,
 } from '../../reset';
 import { VisuallyHidden } from '../Visually-hidden/Visually-hidden';
 import { Paragraph, ParagraphProps } from '../Paragraph/Paragraph';
@@ -22,7 +23,6 @@ const labelBaseClassName = `${baseClassName}_label`;
 const inputBaseClassName = `${baseClassName}_input`;
 const statusTextClassName = `${baseClassName}_statusText`;
 const statusTextContainerClassName = `${statusTextClassName}_container`;
-const statusTextLabelClassName = `${statusTextClassName}_label`;
 
 export interface TextInputLabelProps extends HtmlLabelProps {}
 
@@ -103,9 +103,9 @@ class BaseTextInput extends Component<TextInputProps> {
             />
             {children}
           </HtmlDiv>
-          <HtmlLabel className={statusTextLabelClassName} id={generatedId}>
+          <HtmlSpan className={statusTextClassName} id={generatedId}>
             {statusText}
-          </HtmlLabel>
+          </HtmlSpan>
         </HtmlDiv>
       </HtmlLabel>
     );
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -78,7 +78,7 @@ class BaseTextInput extends Component<TextInputProps> {
     } = this.props;
 
     const hideLabel = labelMode === 'hidden';
-    const generatedId = idGenerator(`${propId}-statusText`);
+    const generatedId = `${idGenerator(propId)}-statusText`;
 
     return (
       <HtmlLabel
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -14,11 +14,14 @@ import { logger } from '../../utils/logger';
 import classnames from 'classnames';
 import styled from 'styled-components';
 import { disabledCursor } from '../utils/css';
+import { idGenerator } from '../../utils/uuid';
 
 const baseClassName = 'fi-text-input';
 const disabledClassName = `${baseClassName}--disabled`;
 const labelBaseClassName = `${baseClassName}_label`;
 const inputBaseClassName = `${baseClassName}_input`;
+const stateTextClassName = `${baseClassName}_stateText`;
+const stateTextLabelClassName = `${stateTextClassName}_label`;
 
 export interface TextInputLabelProps extends HtmlLabelProps {}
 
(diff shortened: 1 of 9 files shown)


src/components/Form/TextInput.tsx
```diff
@@ -83,13 +91,19 @@ class BaseTextInput extends Component<TextInputProps> {
...
+          <HtmlDiv {...inputContainerProps}>
+            <HtmlInput
+              {...passProps}
+              className={classnames(inputBaseClassName, inputClassName)}
+              type="text"
+              aria-describedby={generatedId}
+            />
+            {children}
+          </HtmlDiv>
+          <HtmlLabel className={stateTextLabelClassName} id={generatedId}>
```
aappoalander: Is it intended to apply the id prop to label? Shouldn't the id be added to the actual input element or does aria change the way these are handled? How does this work overall if I give the TextInput component an id? Can I expect that id to be present in the underlying input and if not, could it have undesired side effects?

src/core/Form/SearchInput/SearchInput.test.tsx
```diff
@@ -5,7 +5,11 @@ import { axeTest } from '../../../utils/test/axe';
 import { SearchInput } from './SearchInput';
 
 const TestSearchInput = (
-  <SearchInput labelText="Test search input" data-testid="textinput" />
+  <SearchInput
+    labelText="Test search input"
+    data-testid="textinput"
+    id="test-id"
```
aappoalander: As mentioned above, if the id is available like this and is used to access the component in DOM, what should the enduser expect to get, a label or an input? Or does it make any difference?

src/components/Form/TextInput.tsx
ketsappi: It is by mistake, label should have it's own id and not the input's id-prop. Will be fixed.

aappoalander:
> TextInput will now have the `variant`-prop exposed publicly
> * variant = 'default' | 'error' | 'success'
For example Material UI defines the "state" API like this: https://material-ui.com/api/input/
The variants are actually boolean props => <input error id...>

Using "state"-term might be misleading as it could also refer to the "state" of the component rather than status/style like error/success/default.

We could also imitate the FormControl structure from MUI and use a statusText/Label component to provide that functionality. https://material-ui.com/components/text-fields/#api

Anyways, the internal implementation is of less importance in my opinion. The public API is the thing that matters and with that I'd use terms that are as general/widely used as possible.

ketsappi:
> > TextInput will now have the `variant`-prop exposed publicly
> > 
> > * variant = 'default' | 'error' | 'success'
> >   For example Material UI defines the "state" API like this: https://material-ui.com/api/input/
> >   The variants are actually boolean props =>
> 
> Using "state"-term might be misleading as it could also refer to the "state" of the component rather than status/style like error/success/default.
> 
> We could also imitate the FormControl structure from MUI and use a statusText/Label component to provide that functionality. https://material-ui.com/components/text-fields/#api
> 
> Anyways, the internal implementation is of less importance in my opinion. The public API is the thing that matters and with that I'd use terms that are as general/widely used as possible.

Currently design talk about "help text" that is between the input's label and the actual input component. And there doesn't seem to be good name for this stateText.
Otherwise e.g `InputHelperText` would be good, as it intended to use with TextInput or SearchInput.

**But** of course we could make it even more generic that it could be used with other components too, e.g Toggle. Just need some more thinking of how it should be implemented and such. Then name could be like `HelperText`. Then user should be instructed how to use it correctly (should add `aria-describedby` prop for component and the id of the HelperText) etc.

LJKaski:
Is the font size specified for this? Seems quite large/bulky, but of course it has to be of decent size for accessibility and usability.

aappoalander:
> **But** of course we could make it even more generic that it could be used with other components too, e.g Toggle. Just need some more thinking of how it should be implemented and such. Then name could be like `HelperText`. Then user should be instructed how to use it correctly (should add `aria-describedby` prop for component and the id of the HelperText) etc.

I propose we create a following ticket for this and implement it internally for now. Maybe we could do a component implementation internally and not export the component just yet? Would of course require a prop to set it for the inputs before the component is publicly available. Maybe name it something like inputStateText which makes a clear distinction between state and this text?

ketsappi:
> Is the font size specified for this? Seems quite large/bulky, but of course it has to be of decent size for accessibility and usability.

It uses the `bodySemiBoldSmall` from the theme as used in the design. But indeed, it is showing quite large.

https://github.com/vrk-kpa/suomifi-ui-components/blob/faf4133fc424dbbbeaf4df3593a3f5874a27d862/src/core/Form/TextInput/TextInput.baseStyles.tsx#L32

aappoalander:
> Maybe name it something like inputStateText which makes a clear distinction between state and this text?
Could be InputHelperText component and inputHelperText prop as well if implemented according to my proposal.

ketsappi:
> I propose we create a following ticket for this and implement it internally for now. Maybe we could do a component implementation internally and not export the component just yet? Would of course require a prop to set it for the inputs before the component is publicly available. Maybe name it something like inputStateText which makes a clear distinction between state and this text?
Could be InputHelperText component and inputHelperText prop as well if implemented according to my proposal.

Okay, that sounds reasonable for now. I will create a following ticket for this and implement as you suggested.

src/components/Form/TextInput.tsx
aappoalander: This should work too, unless one gives the same id to multiple inputs?

ketsappi: Yes, that is true.

But the `id` [must be unique in the whole document](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id) by the definition.
It was possible to give the `id` attribute for the TextInput even before this change, so user could use it wrong already.

If not supporting like this, I couldn't find a way to have working snapshots as `idGenerator` generates new one for every test run. But I'm open for suggestion for alternative working solution.

aappoalander:
Based on the discussions yesterday, let's wait for the decision about reserving space for additional texts in advance vs. growing the component dynamically.

src/components/Form/TextInput.tsx
aappoalander: It's fine, my latest comment should be ignored.

src/core/Form/TextInput/TextInput.md
```diff
@@ -29,3 +29,25 @@ import { TextInput } from 'suomifi-ui-components';
...
+    labelText="Test TextInput"
+    statusText={statusText}
+    variant={variant}
+  />
+
+  <button onClick={() => setErrorState(!errorState)}>
+    Toggle error state
+  </button>
+</>;
+```
```
ketsappi: I added an example to styleguidist how to use the `statusText` and `variant`.  Do others see this as useful or should it be removed?

I also used the basic button for not to have this example dependent of Button component. Is this okay, for others too?

aappoalander:
> TextInput will now have the `variant`-prop exposed publicly
> * variant = 'default' | 'error' | 'success'

I'd propose we change this something that is more in line with other API's in our library. Most common way in widely used libs seems to be a boolean prop for all of these. If we wan't to keep it explicit so that only one can be chosen, then maybe something like _status_ could be more descriptive. Variant in my opinion refers more to something that does not change on runtime like Text.lead which equals to Text variant="lead".

ketsappi:
> I'd propose we change this something that is more in line with other API's in our library. Most common way in widely used libs seems to be a boolean prop for all of these. If we wan't to keep it explicit so that only one can be chosen, then maybe something like _status_ could be more descriptive. Variant in my opinion refers more to something that does not change on runtime like Text.lead which equals to Text variant="lead".

- The possibility to use the static ones (TextInput.success / TextInput.error) were now removed.
- `variant` prop is renamed to `status` to better describe it's use and to prevent confusion

src/components/Form/TextInput.tsx
```diff
@@ -14,11 +14,14 @@ import { logger } from '../../utils/logger';
 import classnames from 'classnames';
 import styled from 'styled-components';
 import { disabledCursor } from '../utils/css';
+import { idGenerator } from '../../utils/uuid';
 
 const baseClassName = 'fi-text-input';
 const disabledClassName = `${baseClassName}--disabled`;
 const labelBaseClassName = `${baseClassName}_label`;
 const inputBaseClassName = `${baseClassName}_input`;
+const statusTextClassName = `${baseClassName}_statusText`;
```
aappoalander: Should this be statusTextContainerClassName and statusTextClassName (since status text is not a label)?

src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -12,22 +12,38 @@ export const baseStyles = withSuomifiTheme(
...
   &.fi-text-input--success {
     & .fi-text-input_input {
-      color: ${theme.colors.successBase};
       border-color: ${theme.colors.successBase};
     }
   }
+  &.fi-text-input--disabled {
+    & .fi-text-input_input {
+      color: ${theme.colors.depthBase};
+      background-color: ${theme.colors.depthLight30};
```
aappoalander: is this up to date? Was the tokens update already merged?

src/components/Form/TextInput.tsx
ketsappi: After the recent commit, these are now like in following:

https://github.com/vrk-kpa/suomifi-ui-components/blob/4240cc93b646b319c771e41b9c4887a0635d6c8a/src/components/Form/TextInput.tsx#L23-L25

src/core/Form/TextInput/TextInput.baseStyles.tsx
ketsappi: No, not yet rebased to recent changes on develop. Will have to do it next.

LJKaski:
Screen reader seems to read the error message twice for some reason upon entering the component with error status on. It reads the error message, then the NVDA "invalid entry" that comes from the aria-invalid tag, then the error message again. Default and success variations skip the aria-invalid message as they should but still read the error message twice.

ketsappi:
> Screen reader seems to read the error message twice for some reason upon entering the component with error status on. It reads the error message, then the NVDA "invalid entry" that comes from the aria-invalid tag, then the error message again. Default and success variations skip the aria-invalid message as they should but still read the error message twice.

Interesting. 
When does this occur? 
 - When you do page refresh? 
- When you focus the input-field?

LJKaski:
> > Screen reader seems to read the error message twice for some reason upon entering the component with error status on. It reads the error message, then the NVDA "invalid entry" that comes from the aria-invalid tag, then the error message again. Default and success variations skip the aria-invalid message as they should but still read the error message twice.
> 
> Interesting.
> When does this occur?
> 
> * When you do page refresh?
> * When you focus the input-field?

When the input-field is focused.

ketsappi:
> > > Screen reader seems to read the error message twice for some reason upon entering the component with error status on. It reads the error message, then the NVDA "invalid entry" that comes from the aria-invalid tag, then the error message again. Default and success variations skip the aria-invalid message as they should but still read the error message twice.
> > 
> > 
> > Interesting.
> > When does this occur?
> > 
> > * When you do page refresh?
> > * When you focus the input-field?
> 
> When the input-field is focused.

This seems to be issue with Chrome on macOS too. 
On Safari it tells it only once. 
On the other hand Firefox doesn't say anything (or I just can not find a way to use VoiceOver correctly with Firefox)

PR #296, opened 2020-03-11
Release 0.8.0

LJKaski:
# Description

Raise version number to 0.8.0 in preparation of publishing the recent changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "suomifi-ui-components",
-  "version": "0.7.0",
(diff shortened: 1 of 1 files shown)


ketsappi:
Just a small note, but would be nice if releases follow some naming pattern.
Could the PR name be like **Release/0.80** or **Release 0.8.0** or similar.

Now the name is little vague and some might read that some dependency version was just updated.

Of course we have that **release** label also, but still.

PR #298, opened 2020-03-13
[Feature] Language menu improvements

aappoalander:
## Description
Adjust language menu-button size to match designs (height 40px) and change menu icon rotation when menu is open.

## Motivation and Context

To allow menu to be used with buttons in designs where height is based on button defaults, the language menu height needed to be changed. Also, in order to better communicate the functionality of the menu button when the menu list is visible, the icon should be rotated 180 degrees from the original.

## How Has This Been Tested?
Tested using styleguidist, CRA typescript project and design system site using verdaccio.

### DIFF
--- a/src/components/LanguageMenu/LanguageMenu.tsx
+++ b/src/components/LanguageMenu/LanguageMenu.tsx
@@ -56,12 +56,12 @@ type OptionalLanguageMenuListProps = {
 export interface LanguageMenuProps {
   /** Name or content of menubutton */
   name: React.ReactNode | ((props: { isOpen: boolean }) => React.ReactNode);
-  /** onMenuOpen function callback for open / close event */
-  onMenuOpen?: (isOpen: boolean) => void;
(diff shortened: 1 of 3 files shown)


src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -85,20 +88,31 @@ export class LanguageMenu extends Component<LanguageMenuProps> {
...
-            <MenuList {...menuListProps}>{children}</MenuList>
-          )}
+          {({ isOpen }: { isOpen: boolean }) => {
+            return (
+              <Fragment>
+                <MenuButton
+                  {...passProps}
+                  className={classnames(
+                    menuButtonClassName,
+                    isOpen ? menuButtonOpenClassName : '',
```
ketsappi: Not a real problem but possible alternative solution to prevent "empty" ternary operation choice is by changing to:
`isOpen && menuButtonOpenClassName,`

PR #302, opened 2020-03-25
[BreakingChange] Update to 5px based spacing and inset spacing.

aappoalander:
## Description

This PR introduces new 5px based spacing tokens and inset spacing tokens and updates the components to use those internally.

The components remain mostly as-is. The exported theme will change and some existing exported spacing values will change.

## Motivation and Context

To match current designs, spacing system needed to be changed to be 5px based, but also support for smaller 2px based system for component internal spacing was needed.

## How Has This Been Tested?
Tested with suomifi-design-system and create-react-app using verdaccio.

### DIFF
--- a/.styleguidist/spacing.md
+++ b/.styleguidist/spacing.md
@@ -4,7 +4,6 @@ import { colors } from '../src/core/theme/colors';
 import { cssValueToString } from '../src/utils/css';
 import { element, fonts } from '../src/core/theme/reset';
 import { spacing } from '../src/core/theme/spacing';
-import { suomifiTheme } from '../src/core/theme';
 import { Text } from '../src/core/Text/Text';
 import clipboardCopy from 'clipboard-copy';
 
--- a/.styleguidist/spacing.md
+++ b/.styleguidist/spacing.md
@@ -4,6 +4,7 @@ import { colors } from '../src/core/theme/colors';
 import { cssValueToString } from '../src/utils/css';
 import { element, fonts } from '../src/core/theme/reset';
 import { spacing } from '../src/core/theme/spacing';
+import { suomifiTheme } from '../src/core/theme';
 import { Text } from '../src/core/Text/Text';
 import clipboardCopy from 'clipboard-copy';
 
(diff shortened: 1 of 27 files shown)


.styleguidist/styleguidist.sections.js
```diff
@@ -47,7 +47,7 @@ module.exports = {
         },
         {
           name: 'Spacing',
-          content: './.styleguidist/spacing.md',
+          content: './.styleguidist/spacing.insetXld',
```
ketsappi: This should be `spacing.md`.

aappoalander: Fixed. Was an error with search replace. Good catch!

src/core/Block/__snapshots__/Block.test.tsx.snap
```diff
@@ -43,51 +43,131 @@ exports[`calling render with the same component on the same container does not r
 }
 
 .c0.fi-block--margin-xxs {
-  margin: 2px;
+  margin: 5px;
```
aappoalander: These still require investigation. Not sure if this test has any real implications, or is it just used to test all combinations for spacing.

aappoalander: So it looks like some props are generated out of tokens and now include the updated values and new inset-tokens as well. We should probably consider what to do with these.

aappoalander: It was decided to exclude the inset tokens from the components public API's. Inset spacing is available in design tokens and can be used with traditional generic API's like style that is mostly supported by the components as well.

.styleguidist/spacing.md
```diff
@@ -4,6 +4,7 @@ import { colors } from '../src/core/theme/colors';
 import { cssValueToString } from '../src/utils/css';
 import { element, fonts } from '../src/core/theme/reset';
 import { spacing } from '../src/core/theme/spacing';
+import { suomifiTheme } from '../src/core/theme';
```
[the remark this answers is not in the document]
aappoalander: Nope, forgot it after removing the added preview container size calculation. Removed now.

PR #308, opened 2020-04-03
[Feature] Tsdx css configuration

aappoalander:
## Description

This PR add support for css and sass handling for the project. TSDX is no configured to handle, extract and bundle styles imported directly to ts and tsx files. Typescript loader for TSDX was changed to enable proper handling of css imports.

## Motivation and Context

In preparation for migrating away from styled-components to css-modules, the build needs support for css and sass. To support css modules, some additional configuration is still needed, but basic cases should be covered now.

In addition the build process is now simpler since no separate handling of css is required.

Also jest tests now support css imports and testing with classnames through identity object proxy.

## How Has This Been Tested?

Tested by running snapshot tests which remain unchanged and building the Styleguidist for visual inspection. Also the exported bundled main.css remains unchanged. Also tested with create react app TS project as well as with simple SSR setup using verdaccio locally.


tlehtimaki:
Just out of interest I would have liked to ask, that what made you move away from style-components? What was the key reason? Performance perhaps? (no obligation to answer obviously, but it would be nice to know :) )

aappoalander:
> Just out of interest I would have liked to ask, that what made you move away from style-components? What was the key reason? Performance perhaps? (no obligation to answer obviously, but it would be nice to know 😃 )

@tlehtimaki We actually didn't. At least not yet. This feature allows us to move to that direction later on if styled-components proves to be too cumbersome to configure for projects using this library. Especially configuring for SSR and very strict CSP is not straightforward by any means. For now, we just mainly want to bundle some global css from 3rd party dependencies (Reach).

tlehtimaki:
So it was just my misunderstanding. Thanks for the answer.

PR #309, opened 2020-04-08
[Feature] Update all dependencies to latest

aappoalander:
## Description

This PR upgrades all dependencies to the latest available stable version and resets the lock-file to resolve sub-dependencies again as well. Includes adjustments and fixes related to breaking changes in dependencies.

## Motivation and Context

To better allow upgrading vulnerable dependencies, it's better to have all dependencies up-to-date. Updating React and Styled-components did not introduce breaking changes and the suomifi-ui-components  backwards compatibility should remain as previously.

## How Has This Been Tested?

WIP: Not yet tested with the exception of included Styleguidist. Needs to be thoroughly tested with existing test projects and especially components with more than just cosmetic style changes need further testing before merging these changes.


src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -62,7 +62,10 @@ export interface LanguageMenuProps {
     OptionalLanguageMenuPopoverProps
   >;
   /** Menu items: MenuItem or MenuLink */
-  children?: Array<React.ReactElement<LanguageMenuPopoverItemsProps>>;
+  children?:
```
aappoalander, line 65: Could this cause unwanted side-effects? This was fixed due to type error after updating TS and linters.

src/core/Dropdown/Dropdown.test.tsx
```diff
@@ -28,6 +28,9 @@ test(
       'aria-valid-attr-value': {
         enabled: false,
       },
+      region: {
+        enabled: false,
+      },
```
aappoalander, line 33: These need further investigation, an additional ticket needs to be created and libs axe utility needs to be updated to do a deep merge for options provided as parameter and util defaults.

src/core/Expander/ExpanderGroup.tsx
```diff
@@ -33,22 +33,21 @@ const StyledExpanderGroup = styled(
...
 )`
-  ${props => baseStyles(props)};
+  ${(props) => baseStyles(props)};
 `;
 
 const OpenAllButton = ({
   children,
   ...passProps
 }: ExpanderOpenAllButtonProps) => {
-  const Component = React.Children.toArray(children)[0];
```
aappoalander, line 43: This is another thing to test thoroughly. React.Children.toArray relies on JSX elements and the runtime type information of these is very limited. This broke after updating TypeScript and linters.

src/core/Form/TextInput/TextInput.md
```diff
@@ -12,7 +12,7 @@ import { TextInput } from 'suomifi-ui-components';
 
 <>
   <TextInput
-    onBlur={event => console.log(event.target.value)}
+    onBlur={(event) => console.log(event.target.value)}
```
aappoalander, line 15: Some console logs seem to be empty in styleguidist. Needs to be fixed?

src/core/LanguageMenu/LanguageMenu.test.tsx
```diff
@@ -31,6 +31,9 @@ test(
       'aria-valid-attr-value': {
         enabled: false,
       },
+      region: {
```
aappoalander, line 34: Same axe util things apply to this as above.

src/utils/css/pseudo.ts
```diff
@@ -13,7 +13,7 @@ type absolutePseudos = 'before' | 'after';
 
 export const absolute = (pseudo: absolutePseudos) => css`
   position: relative;
-  &:${/* sc-prop */ pseudo} {
+  &: ${/* sc-prop */ pseudo} {
```
aappoalander, line 16: Does this change affect behaviour? It was automatically applied by prettier after update.

src/utils/test/axe.ts
```diff
@@ -4,7 +4,7 @@ import { axe } from 'jest-axe';
 
 export const axeTest = (
   element: ReactElement<any>,
-  options: any = {},
+  options: any = { rules: { region: { enabled: false } } },
```
aappoalander, line 7: This needs an update to merge defaults with provided props as default in conflict case.

src/core/Dropdown/Dropdown.test.tsx
aappoalander: There is now a shallow merge of options.rules with default options. In case same key is provided with options.rules, it overrides corresponding default option rule. Deep merge is not desirable for rules object since it could result into unwanted rule-sets that are difficult to understand.

aappoalander: Created an issue to our backlog to investigate if we should enable region rule.

src/core/LanguageMenu/LanguageMenu.test.tsx
aappoalander: This was fixed and region is now disabled by default and does not need to be set here as well.

src/utils/test/axe.ts
aappoalander: Rules are now shallow merged using provided values as primary and defaults as secondary.

src/utils/css/pseudo.ts
aappoalander: The behaviour does indeed change. This is probably a prettier bug (https://github.com/prettier/prettier/pull/6794). So added a prettier-ignore rule for this block. Can be removed once bug is fixed and functionality tested again.

src/components/LanguageMenu/LanguageMenu.tsx
aappoalander: As the prop is optional, this change does not affect API behavior. However, leaving out children completely has a side-effect of not rendering the menu at all. I'm wondering if we should make the prop mandatory and remove the replace-component wrapper completely.

src/core/Expander/ExpanderGroup.tsx
aappoalander: This seems to work as intended. The typeof children is either object, string or null etc. So in case we provide a string or something not Suomifi Button, it is rendered inside unstyled button. Button is rendered as is.

src/core/Form/TextInput/TextInput.md
aappoalander: Nevermind, as it is onBlur, the event fires even if the value of the input has not changed when we navigate away from the input.

aappoalander:
> Styleguidist and typescript test app seem to work ok with the changes.
> 
> Got some console warnings and errors about deep requiring uuid (_Deep requiring like `const uuidv4 = require('uuid/v4');` is deprecated as of [uuid@7.x](mailto:uuid@7.x)._) as well as dynamic styles being used on some of the styleguidist pages (Typography, Spacing and Icons)

Uuidv4 is now fixed. It was caused by updating deps to latest.

What comes to dynamic creation of styled components, this issue concerns only dev and is not related to updating deps. The console warning for using styled inside render was added to styled-components v5, the related functionality has not changed. => I propose not to do anything about it.

PR #315, opened 2020-04-15
[Feature] Add checkbox component

LJKaski:
## Description
Adds first version of a Checkbox component to the library. Supports suomi.fi styles and has basic functionalities and variants in place.

Still has some placeholder features and styles such as the focus style. Styles also need to be rearranged or -written to pass stylelint tests if deemed necessary. Currently violates the no-descending-specificity rule due to the way css-in-js styles are written in general, but has the rule disabled for the file.

## Motivation and Context
Checkbox is a basic form component that needs to be found in the component library.

## How Has This Been Tested?
Some automated jest tests including component-level Axe tests as well as manual testing. Also tested with NVDA screen reader.

## Notes
The possible use of aria-live for the status message should be discussed


src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+export class Checkbox extends Component<CheckboxProps> {
+  state = {
+    checkboxState: !!this.props.checked || !!this.props.defaultChecked || false,
+  };
+
+  /**
+   * This is deprecated and should be replaced
+   */
+  componentWillReceiveProps(nextProps: CheckboxProps) {
```
ketsappi, line 92: As you have commented already about the deprecation, but pasting the React's documentation about this for explanation:
https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,175 @@
...
+          className={classnames(className, checkboxBaseClassNames.label)}
+          {...passProps}
+        >
+          {children}
+        </HtmlLabel>
+        {hintText ? (
+          <HtmlSpan className={checkboxBaseClassNames.hintText} id={hintTextId}>
+            {hintText}
+          </HtmlSpan>
+        ) : null}
```
ketsappi, line 162: This might be more readable without the need of ternary operator and returning `null` value.

After change it would be like:

```tsx
{!!hintText && (
	<HtmlSpan className={checkboxBaseClassNames.hintText} id={hintTextId}>
		{hintText}
	</HtmlSpan>
)}
```

src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+        {statusText ? (
+          <HtmlSpan
+            className={checkboxBaseClassNames.statusText}
+            id={statusTextId}
+          >
+            {statusText}
+          </HtmlSpan>
+        ) : null}
```
ketsappi, line 171: Same possibility as with the `hintText` above ^^

src/components/index.ts
```diff
@@ -8,6 +8,7 @@ export {
   DropdownItemProps,
 } from './Dropdown/Dropdown';
 export { TextInput, TextInputProps } from './Form/TextInput';
+export { Checkbox, CheckboxProps } from './Form/Checkbox';
```
ketsappi, line 11: Should this also export the `CheckboxInputProps`?

src/core/Form/Form.tsx
```diff
@@ -6,3 +6,4 @@ export {
 } from './Toggle/Toggle';
 export { TextInput, TextInputProps } from './TextInput/TextInput';
 export { SearchInput, SearchInputProps } from './SearchInput/SearchInput';
+export { Checkbox, CheckboxProps } from './Checkbox/Checkbox';
```
ketsappi, line 9: Should this also export `CheckboxInputProps`?

src/index.tsx
```diff
@@ -10,6 +10,8 @@ export { UnstyledButton } from './core/Button/UnstyledButton';
 export { Dropdown, DropdownProps } from './core/Dropdown/Dropdown';
 export { DropdownItem, DropdownItemProps } from './core/Dropdown/DropdownItem';
 export {
+  Checkbox,
+  CheckboxProps,
```
ketsappi, line 14: Should this also export `CheckboxInputProps`?

src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+type CheckboxVariant = 'small' | 'large';
+type CheckboxStatus = 'default' | 'error' | 'disabled';
+
+export interface CheckboxInputProps {
+  /** State of input checkbox */
+  checked?: boolean;
+  /** Custom classname for the container */
+  className?: string;
+  disabled?: boolean;
```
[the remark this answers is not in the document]
LJKaski: Good point. Started thinking if this is even needed. Are there cases where we want to pass the disabled attribute only to the input and not the entire component?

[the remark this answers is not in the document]
ketsappi: True. I don't think there are cases where only the input is disabled.

src/components/Form/Checkbox.tsx
LJKaski: Forgot about that syntax. Thanks :)

src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,175 @@
+import React, { Component, ReactNode } from 'react';
+import classnames from 'classnames';
+import { HtmlInput, HtmlLabel, HtmlSpan, HtmlDiv } from '../../reset';
+
+import { idGenerator } from '../../utils/uuid';
+
+const baseClassName = 'fi-checkbox';
+
+export const checkboxBaseClassNames = {
```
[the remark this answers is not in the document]
LJKaski: This is now removed.

src/components/Form/Checkbox.tsx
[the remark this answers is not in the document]
LJKaski: Removed this.

src/components/Form/Checkbox.tsx
LJKaski: Replaced with the truthy syntax.

src/components/Form/Checkbox.tsx
LJKaski: Replaced as well.

src/components/index.ts
LJKaski: Not sure if that's necessary. The interface is used to specify which attributes are available to be passed to the input element when using the `checkboxInputProps` prop of the parent component.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
```diff
@@ -0,0 +1,175 @@
...
+      &:checked + .fi-checkbox_label::after {
+        content: '';
+      }
+    }
+    & .fi-checkbox_label {
+      position: relative;
+      display: inline-block;
+      padding-left: ${theme.spacing.l};
+      min-height: 30px;
+      min-width: 30px;
```
LJKaski, line 128: These are still a bit speculative. The element needs to have a size to show the checkbox itself, but the values could be tweaked maybe.

aappoalander:
Branch names should include feature/ fix/ release/ etc. No need to change this imo, but something to keep in mind for the future.

src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,175 @@
...
+  disabled: `${baseClassName}--disabled`,
+  container: `${baseClassName}_container`,
+  input: `${baseClassName}_input`,
+  label: `${baseClassName}_label`,
+  statusText: `${baseClassName}_status`,
+  hintText: `${baseClassName}_hintText`,
+};
```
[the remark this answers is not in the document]
aappoalander: Even better, create a getter that always includes the baseClassName?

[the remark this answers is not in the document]
aappoalander: import { getClassName } from 'components/checkbox';

getClassName().disabled; vs. checkboxBaseClassNames.disabled?

Maybe the latter is better. And I propose to change the object name to checkboxClassName, since the base refers to prefix fi-checkbox.

src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * 'small' | 'large'
+   * @default small
+   */
+  variant?: CheckboxVariant;
+  /**
+   * 'default' | 'error' | 'disabled'
+   * @default default
+   */
+  status?: CheckboxStatus;
```
aappoalander, line 56: Would it be better to have the disabled prop removed and that behaviour controlled by the status? Or would that be inconsistent with checkbox JSX props? Now one would have to do <Checkbox disabled status="disabled" />

src/components/Form/Checkbox.tsx
aappoalander: This approach needs to be changed. Will be depricated as of React 17 and name should be UNSAFE_componentWillReceiveProps as we use React 16.9 which deprecates the original name. In addition the documentation behind the link above advices to use a different approach instead of using this lifecycle method. I propose to find an alternative method.

src/components/Form/Checkbox.tsx
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * aria-label for the HTML input-element,
+   * alternatively you can define aria-labelledby with label-element id
+   */
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
+  /** Unique id
+   * @default uuidV4
+   */
+  id?: string;
```
aappoalander, line 76: Maybe add explanation that default is autogenerated uuidV4?

src/components/index.ts
aappoalander: I thing it's more a matter of consistency. If we have exported sub-types for the API elsewhere then we should do it here as well. It is also a matter of architecture, if we would be using composition then it would not be as necessary. With children as props it's more useful to provide the typings since w might want to do `const childProps: ChildProps = ...` and `<Comp childProps={childProps} />` if our code style so requires.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
aappoalander: Are we able to say what should a min-width be? If we set something the I think the sizes should come from design. What is the display style of the element and how have we done this with other components? Maximum size could also be limited to window size.

src/core/Form/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,89 @@
...
+    hintText={hint}
+    statusText={statusText}
+    data-testid={dataTestId}
+    id="test"
+  >
+    {labelText}
+  </Checkbox>
+);
+
+const RegularTestCheckbox = createTestCheckbox(
```
aappoalander, line 28: Looks good, maybe add disabled as well since that might be quite commonly used and depends on styles.

src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,75 @@
...
+const checkboxClassNames = {
+  error: `${baseClassName}--error`,
+  checked: `${baseClassName}--checked`,
+  large: `${baseClassName}--large`,
+};
+
+export interface CheckboxProps extends CompCheckboxProps, TokensProp {}
+export interface CheckboxInputProps
+  extends CompCheckboxInputProps,
+    TokensProp {}
```
aappoalander, line 24: This indentation looks weird, is this really what our prettier config now produces?

src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,75 @@
...
+    } = withSuomifiDefaultProps(this.props);
+    const { checkboxStatus } = this.state;
+    return (
+      <StyledCheckbox
+        disabled={disabled}
+        onClick={this.handleClick}
+        {...passProps}
+        className={classnames(className, {
+          [checkboxClassNames.error]: status === 'error',
+          [checkboxClassNames.checked]: checkboxStatus,
```
aappoalander, line 67: Not sure if this could be improved, but now it's a bit confusing since we have CheckboxStatus which is the interface for CheckboxProps.status. Now we have state.checkboxStatus which represents the checked state. Maybe call it `checkedState` etc. more descriptive?

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
```diff
@@ -0,0 +1,175 @@
...
+  SuomifiThemeProp,
+} from '../../theme';
+import { disabledCursor } from '../../../components/utils/css';
+import { element, font } from '../../theme/reset';
+
+/* stylelint-disable */
+const checkedStyles = ({ theme }: SuomifiThemeProp) => css`
+  &.fi-checkbox--checked .fi-checkbox_label {
+    &::before,
+    &::after {
```
aappoalander, line 14: Are these the styles that prevent using current focus styles with the component? Are these referring to the container? If so, shouldn't it be possible to target these styles without pseudo and use the current focus styles?

src/core/Form/Checkbox/Checkbox.md
```diff
@@ -0,0 +1,40 @@
...
+    This is a regular checkbox with a hint text and an error message
+  </Checkbox>
+
+  <Checkbox defaultChecked disabled>
+    This is a disabled checkbox
+  </Checkbox>
+
+  <Checkbox />
+
+  <Checkbox id="2" variant="large">
```
aappoalander, line 27: Do we support `<Checkbox.large>Large</Checkbox.large>` syntax? If so, should we also document it?

src/components/Form/Checkbox.tsx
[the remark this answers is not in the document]
aappoalander: And now that I read this PR more thoroughly, it's also a bit confusing to have checkboxBaseClassNames here with the --disabled style, then we have error and other derived styles elsewhere and the relationship between components and core styles and classnames is difficult to understand in my opinion. If these have an effect on the core styles as well or if core styles are built on top of these, I think everything that is put together in core should make sense / be consistent and clear in a way that it's easy to understand what comes from components, what from core and that there are no conflicts. Having e.g. baseClassName with the composed object here and then using baseClassName = "fi-checkbox..." in core is a bit confusing.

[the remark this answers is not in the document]
aappoalander: We have a lot of these elsewhere, but I think this could be improved.

src/components/Form/Checkbox.tsx
aappoalander: As discussed, it is also possible to check if disabled exists and dismiss other status props if so. Allowing e.g. disabled and checked would require styles to be defined for designs etc.

src/core/Form/Checkbox/Checkbox.md
aappoalander: Seems to be supported, so I'd add an example showing this as well.

src/components/Form/Checkbox.tsx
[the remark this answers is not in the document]
aappoalander: I noticed an issue with some of the classnames. We indicate status with some of the classnames which is ok. However, in some cases we apply the status indicating class to underlying elements, like svg in toggle and the svg does not have a reliable selector to use with testing. These classnames do indicate a status as well. I guess it would be ok to leave these as they are now to be consistent with older components. Is there a reason to have some of the classnames / styles in component files and other here?

src/components/Form/Checkbox.tsx
aappoalander: This should be doable with getDerivedStateFromProps using similar approach with Toggle and Expander.

src/components/Form/Checkbox.tsx
LJKaski: I'd say that's something we can leave for the users of the component. For example, in some contexts you might want to show a disabled checkbox with an error message if an action is required by a user to enable the checkbox. The disabled+checked variant can be found in the styleguide, so I'd leave that as it is.

aappoalander: I think the behavior of the API should be deterministic and easy to understand. Are these equivalent: `<Checkbox status='disabled'>` and `<Checkbox disabled>`? What happens if I have this `<Checkbox disabled status='error'>`? 

I propose the following:
1) Remove disabled from CheckboxStatus
2) Either dismiss the error style if provided with disabled, or define the style so that it works
3) If we support both disabled and error at the same time, ensure that there is matching design and we follow that.

aappoalander: The disabled prop is present in other components as well. As discussed, to keep the API's consistent let's keep it that way and just make sure we disable other status related styles when disabled is present.

src/components/Form/Checkbox.tsx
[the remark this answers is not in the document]
LJKaski: Classes other than base classes should now be in one place (core).

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
LJKaski: Actually, the max-width is yet to be set. Didn't see it set in other components either, so it's left for the user at this point.

src/components/Form/Checkbox.tsx
LJKaski: Other status styles are now turned off when the component is disabled.

src/core/Form/Checkbox/Checkbox.md
```diff
@@ -0,0 +1,37 @@
...
+    Checked large checkbox with a hint text
+  </Checkbox.large>
+
+  <Checkbox disabled>Disabled checkbox</Checkbox>
+
+  <Checkbox
+    checked={checked}
+    onClick={({ checkboxState }) => {
+      if (confirm('Change checkbox state?'))
+        setChecked(checkboxState);
```
ketsappi, line 31: Would like to see braces here to increase readability.

```jsx
if (confirm('Change checkbox state?')) {
  setChecked(checkboxState);
}
```

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
```diff
@@ -0,0 +1,165 @@
...
+      color: ${theme.colors.alertBase};
+      margin-top: ${theme.spacing.xxs};
+    }
+  }
+`;
+
+const largeVariantStyles = ({ theme }: SuomifiThemeProp) => css`
+  &.fi-checkbox--large {
+    & .fi-checkbox_label {
+      padding-left: 40px;
```
ketsappi, line 67: Would it be possible to use design token here? `spacing.xxl` is also 40px.

src/core/Form/Checkbox/Checkbox.tsx
aappoalander: I guess it does since it passes all linting.

LJKaski: I replied to this ages ago and even saw the answer here for a long while. Where the hell did it disappear? But yeah, that's what prettier does here.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
LJKaski: These are referring to the pseudo elements of the label. Before is the checkbox itself and after was the checkmark. After is now gone after the change to the checkmark icon, but  my suspicion is that applying the focus styles on pseudo elements breaks them. Design says the focus should be around the checkbox only. works fine if applied to the whole component. Need to investigate more now that one of the pseudo elements is gone.

src/components/Form/Checkbox.tsx
aappoalander: This seems to be fixed as well.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
aappoalander: Do I understand correctly by assuming that it makes no difference in this case if we don't use the pseudo focus styles since z-index should have no effect on visuals?

src/core/Form/Checkbox/Checkbox.md
aappoalander: Also, please add window.confirm to be more descriptive. This should be fixed for toggle example as well.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
LJKaski: Good find. This should now be fixed.

src/core/Form/Checkbox/Checkbox.md
LJKaski: Changed to window.confirm.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
LJKaski: In this case we are not able to use the pseudo focus styles. We can use the non-pseudo box-shadow variety just fine after it's implemented.

src/core/Form/Checkbox/Checkbox.md
aappoalander: Looks like the braces are fixes as well! :)

PR #318, opened 2020-04-20
[BugFix] skipLibCheck related type issues

aappoalander:
## Description

This PR enables libCheck for TS-build. In addition it replaces Styleguidist webpack build awesome-typescript-loader with ts-loader. Also some related type issues are now fixed.

Styleguidist build related typings are also added as dev dependencies. Some typings that are required with `skipLibCheck: false` configuration are added as peer dependencies.

## Related Issue

Based on feedback from test projects, but actual open issues.

## Motivation and Context

To allow strict type checking in projects using the ui-components library, also typings for dependencies have to be valid and available.

## How Has This Been Tested?

Tested with Suomi.fi design system site, Create React App TS project and React SSR project template.

### DIFF
--- a/package.json
+++ b/package.json
@@ -50,8 +50,9 @@
     "@types/jest-axe": "3.2.2",
     "@types/react": "16.9.32",
     "@types/react-dom": "16.9.6",
-    "@types/styled-components": "5.0.1",
+    "@types/styled-components": "4.1.8",
     "@types/uuid": "7.0.2",
+    "@types/warning": "3.0.0",
     "@typescript-eslint/eslint-plugin": "2.27.0",
     "@typescript-eslint/parser": "2.x",
     "@wessberg/rollup-plugin-ts": "1.2.23",
@@ -123,6 +124,8 @@
     "uuid": "7.0.3"
   },
   "peerDependencies": {
+    "@types/styled-components": "4.1.8",
+    "@types/warning": ">=3.0.0",
     "react": ">=16.8.0",
     "styled-components": ">=4.3.2"
   },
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -123,6 +125,8 @@
     "uuid": "7.0.3"
   },
   "peerDependencies": {
+    "@types/styled-components": "4.1.8",
+    "@types/warning": ">=3.0.0",
     "react": ">=16.8.0",
     "styled-components": ">=4.3.2"
```
ketsappi, line 131: I assume the `@types/styled-components` v4.18 is the last one without the issue, but is the `styled-components` v4.3.2 okay with this or is it already checked?

package.json
```diff
@@ -123,6 +125,8 @@
     "uuid": "7.0.3"
   },
   "peerDependencies": {
+    "@types/styled-components": "4.1.8",
+    "@types/warning": ">=3.0.0",
```
ketsappi, line 129: I get the error that `Could not find a declaration file for module 'warning'`

So if this package is needed when using `"skipLibCheck": false` maybe add that also to the documentation?

package.json
aappoalander: Should be ok, but has not actually been tested with current setup. Types are broken since v.4.1.9. Styled-components did work with current develop until all deps were updated and the API's have not changed.

package.json
aappoalander: Yes, this is expected behavior (this is what peerDependencies mean - you need to handle them on your project and not included in the bundle of referring package.json). Will improve documentation for this.

PR #321, opened 2020-04-27
Replace UNSAFE_componentWillReceiveProps with getDerivedStateFromProps

aappoalander:
## Description

This PR replaces UNSAFE_componentWillReceiveProps lifecycle methods with getDerivedStateFromProps for Toggle component. Functionality or API of the component does not change.

## Motivation and Context

UNSAFE_componentWillReceiveProps has been deprecated and should be replaced with other approaches. For Toggle component it should be relatively straight forward since rendering is always based on state. We are only interested on changes in checked prop when it is defined and differs from state (provided by parent component thus making the state controlled from outside). By using getDerivedStateFromProps we can achieve similar approach as was previously used with UNSAFE_componentWillReceiveProps.

## How Has This Been Tested?
Tested with Suomifi-design-system site, CRA TS project using controlled and uncontrolled approaches with checked.

### DIFF
--- a/src/core/Form/Toggle/Toggle.md
+++ b/src/core/Form/Toggle/Toggle.md
@@ -26,5 +26,12 @@ import { Toggle } from 'suomifi-ui-components';
   <Toggle onClick={({ toggleState }) => console.log(toggleState)}>
     Unchecked enabled using button
   </Toggle>
+
+  <Toggle
+    checked={false}
+    onClick={({ toggleState }) => console.log(toggleState)}
+  >
+    Controlled checked state.
+  </Toggle>
 </>;
 ```
--- a/src/core/Form/Toggle/Toggle.md
+++ b/src/core/Form/Toggle/Toggle.md
@@ -1,5 +1,10 @@
 ```js
+import { useState } from 'react';
+
 import { Toggle } from 'suomifi-ui-components';
+
+const [checked, setChecked] = useState(false);
+
 <>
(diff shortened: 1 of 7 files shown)


src/core/Form/Toggle/Toggle.md
```diff
@@ -26,5 +26,12 @@ import { Toggle } from 'suomifi-ui-components';
   <Toggle onClick={({ toggleState }) => console.log(toggleState)}>
     Unchecked enabled using button
   </Toggle>
+
+  <Toggle
+    checked={false}
+    onClick={({ toggleState }) => console.log(toggleState)}
+  >
+    Controlled checked state.
+  </Toggle>
```
ketsappi, line 35: Maybe add use case of controlled version, to see it in action. Something like this might be nice:

```jsx
import { useState } from 'react';
//...
const [toggleChecked, setToggleChecked] = useState(false);
//...
  <Toggle
    checked={toggleChecked}
    onClick={({ toggleState }) => console.log(toggleState)}
  >
    Controlled checked state.
  </Toggle>
  <button onClick={() => setToggleChecked(!toggleChecked)}>
    Toggle the controlled version
  </button>
```

aappoalander: Tried this but could not get it to work, perhaps due to a bug that is now fixed. Will add something like this to better describe how the component works.

PR #324, opened 2020-05-07
[Fix] Focus style align with style guide

aappoalander:
## Description

Update for current focus styles so that they match the new style guide focus styles.

## Motivation and Context

Current focus styles were following the design system specification which was never taken into use outside design system implementation. This was blocking ui-components usage in projects. Focus styles were redesigned to better match default styles that are in use in may services and applications.

## How Has This Been Tested?

Tested using styleguidist and Create React App TypeScript version with long contents for components.

### DIFF
--- a/src/core/theme/radius.ts
+++ b/src/core/theme/radius.ts
@@ -1,4 +1,4 @@
 export const radius = {
   basic: '2px',
-  focus: '4px',
+  focus: '2px',
 };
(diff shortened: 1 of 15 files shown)


src/core/theme/radius.ts
```diff
@@ -1,4 +1,4 @@
 export const radius = {
   basic: '2px',
-  focus: '4px',
+  focus: '2px',
```
LJKaski, line 3: I think the 4px radius is closer to the current design if that's a viable value otherwise.

aappoalander:
> Try it with the 4px radius. If it works, I'd say go with that value. If not, the current works just fine :)

The 4px radius does not work as well as in the designs. The distance from the component border does not remain constant so I'd stick with the current value.

PR #325, opened 2020-05-08
[Feature] Add non-pseudo focus style

LJKaski:
## Description

This PR implements a focus variant that only uses box shadow instead of a pseudo element. This can be used for inline elements to allow row wrapping focus styles as well as for pseudo elements such as the box in checkbox. This PR also implements the new focus style for checkbox.

## Motivation and Context

The previous focus styles didn't work correctly for inline elements and pseudo elements.

## How Has This Been Tested?

Tested by running locally and manually testing as well as with automated tests.

### DIFF
--- a/src/core/theme/utils/outline.ts
+++ b/src/core/theme/utils/outline.ts
@@ -4,17 +4,13 @@ import { radius } from '../radius';
 
 const boxshadow = ({
   borderRadius,
-  border,
   color,
 }: {
   borderRadius: string;
-  border: string;
   color: string;
 }) => css`
   border-radius: ${borderRadius};
-  border: ${border} solid ${color};
-  box-sizing: border-box;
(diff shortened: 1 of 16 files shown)


src/core/theme/utils/outline.ts
```diff
@@ -50,7 +46,7 @@ const afterBoxshadow = ({
 
 // TODO Refactor, create interfaces (and extend with Partial<>), add JSDOC for functions
 export const boxshadowOutline = ({
-  color = colors.accentBase,
+  color = colors.accentSecondary,
   borderColor = 'white',
```
ketsappi, line 50: Not changed in this PR, but would it be possible to get (`white`) it from theme/tokens?

aappoalander: That would be better, I forgot to change that after initial tests.

PR #328, opened 2020-05-12
[Feature] Align button styles with style guide

aappoalander:
## Description

Button styles now match the style guide. Adjustments to paddings, colors and font-size. No changes to component APIs. Includes improvements to gradient color names.

## Motivation and Context

Some components had not been updated when style guide updates were introduced. These components now have to be aligned with style guide.

## How Has This Been Tested?
Since APIs were not changed, testing with styleguidist should be sufficient. No issues found during testing locally with styleguidist or build.

### DIFF
--- a/src/core/Button/__snapshots__/Button.test.tsx.snap
+++ b/src/core/Button/__snapshots__/Button.test.tsx.snap
@@ -68,11 +68,12 @@ exports[`calling render with the same component on the same container does not r
   font-size: 16px;
   line-height: 1.5;
   font-weight: 600;
-  line-height: 1;
-  padding: 8px 16px;
+  font-size: 14px;
+  line-height: 20px;
+  padding: 10px 20px;
   min-height: 40px;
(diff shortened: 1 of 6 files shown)


src/core/Button/__snapshots__/Button.test.tsx.snap
```diff
@@ -68,11 +68,12 @@ exports[`calling render with the same component on the same container does not r
   font-size: 16px;
   line-height: 1.5;
   font-weight: 600;
-  line-height: 1;
-  padding: 8px 16px;
+  font-size: 14px;
```
aappoalander, line 71: This needs a new typography token before merge.

aappoalander: This typography definition will be replaced with a token later on. Design inventory and definition work needs to be done before adding the tokens and this would be blocking these changes for too long.

PR #329, opened 2020-05-13
[Bugfix] Align expander styles with the design

LJKaski:
## Description
This PR sets the expander component styles to match the latest design and adds a word break rule for the expander content to prevent it from stretching the component.

## Motivation and Context
Components need to be up-to-date with designs in order to be useful.

## How Has This Been Tested?
Tested manually in local environment

### DIFF
--- a/src/core/Expander/Expander.baseStyles.tsx
+++ b/src/core/Expander/Expander.baseStyles.tsx
@@ -34,7 +34,7 @@ export const baseStyles = withSuomifiTheme(
     display: block;
     width: 100%;
     font-size: ${theme.typography.bodySemiBold};
-    height: 60px;
+    min-height: 60px;
     &--no-tag {
       padding: 17px ${theme.spacing.xxxl} 16px ${theme.spacing.m};
       color: ${theme.colors.highlightBase};
@@ -58,6 +58,7 @@ export const baseStyles = withSuomifiTheme(
     display: block;
     height: 0;
     overflow: hidden;
+    word-break: break-word;
     transform: scaleY(0);
     transform-origin: top;
     transition: all ${`${theme.transitions.basicTime}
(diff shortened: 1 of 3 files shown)


src/core/Expander/Expander.baseStyles.tsx
```diff
@@ -33,16 +33,20 @@ export const baseStyles = withSuomifiTheme(
     position: relative;
     display: block;
     width: 100%;
+    font-size: ${theme.typography.bodySemiBold};
+    min-height: 60px;
```
ketsappi, line 37: Would it be possible to use `spacing.xxxl` from the tokens? And little later on for the `height` and `width` of the `.fi-expander_title-icon`

Or where is the fine line when not to use tokens? At least from the naming it might be confusing to use spacing, when we are talking about component height. 
But the ideal would be to use existing tokens is possible to keep different sizes consistent 🤔

LJKaski: That's what I was thinking with these values, and it is a good question indeed.

aappoalander: I would not use spacing tokens for size. Size should consist of content + borders + spacing. Which all can be defined by tokens if we play by the book. E.g. Typography token with line height set + border token or default style with constant width + token for spacing => always a deterministic set of possible heights.

src/core/Expander/Expander.baseStyles.tsx
```diff
@@ -54,6 +58,7 @@ export const baseStyles = withSuomifiTheme(
     display: block;
     height: 0;
     overflow: hidden;
+    word-break: break-word;
```
aappoalander, line 61: How is white-space defined for this component? Inherited from a reset etc?

LJKaski: I think it comes from resets at the moment. It's set to normal, so any sequences of whitespaces are collapsed into one whitespace.

src/core/Expander/Expander.baseStyles.tsx
ketsappi: Let's use the hardcoded pixel value for now and change to use tokens at later alignments.

PR #330, opened 2020-05-13
[Fix] Expander's defaultOpen, open and onClick properties not working when inside ExpanderGroup

ketsappi:
## Description

Expander components props `open`, `defaultOpen` and `onClick` was not working when Expander was inside ExpanderGroup.

When talking about **controlled** state it means that the Expander is given a `open` prop. User is then required to handle the open state of the Expander self.

- `open`, `defaultOpen` and `onClick` should now work as intended and documented.
- Added an example of the controlled state for the Expander. It can be seen on the Expander's styleguidist page.
- Opening state of the individual Expander is stored inside itself, not by parent.
  - Parent will keep own state to handle logic to show icon (*caret*) and text (*Open/Close all*)
- Added tests to verify the use of `open`, `defaultOpen` and `onClick`props when inside ExpanderGroup and when being just standalone Expander.

## Details

Some details about the props and what are they used for.

### open

`open` prop will tell that the component is in controlled state.

### defaultOpen

`defaultOpen` prop does not effect when in controlled state; user can change the state with the `open` prop if needed. Look the controlled example from the styleguidist.

### onClick

When in controlled state,  you can e.g use the Expander's `onClick` property to update the state of individual state of the Expander. Look the controlled example from the styleguidist.

## Motivation and Context

Will help to use the Expander and ExpanderGroup components as the props works as documented.

## How Has This Been Tested?

- Locally on the browser running the styleguidist
- Running tests and validation: `yarn validate`

### DIFF
--- a/src/components/Expander/Expander.tsx
+++ b/src/components/Expander/Expander.tsx
@@ -157,12 +157,13 @@ class ExpanderItem extends Component<ExpanderProps> {
 
   handleClick = () => {
     const { open, onClick } = this.props;
-    const openState = !this.state.openState;
+    const { openState } = this.state;
     const controlled = open !== undefined;
+    const newOpenState = controlled ? !!open : !openState;
 
-    this.setState({ openState: controlled ? !!open : openState });
+    this.setState({ openState: newOpenState });
     if (!!onClick) {
-      onClick({ openState: controlled ? !!open : openState });
+      onClick({ openState: newOpenState });
     }
   };
 
--- a/src/components/Expander/Expander.tsx
+++ b/src/components/Expander/Expander.tsx
@@ -140,9 +140,12 @@ class ExpanderItem extends Component<ExpanderProps> {
       }
     }
 
+    const { open } = this.props;
+    const { openState } = this.state;
+    const controlled = open !== undefined;
     if (
-      prevState.openState !== this.state.openState ||
-      prevProps.open !== this.props.open
+      (prevState.openState !== openState && !controlled) ||
+      (prevProps.open !== open && controlled)
     ) {
       const {
         expanderGroup,
@@ -152,7 +155,8 @@ class ExpanderItem extends Component<ExpanderProps> {
         },
       } = this.props;
       if (!!expanderGroup && !!onExpanderOpenChange && index !== undefined) {
-        onExpanderOpenChange(index);
+        const currentState = controlled ? !!open : openState;
+        onExpanderOpenChange(index, currentState);
       }
     }
   }
--- a/src/components/Expander/Expander.tsx
+++ b/src/components/Expander/Expander.tsx
@@ -163,7 +163,9 @@ class ExpanderItem extends Component<ExpanderProps> {
     const controlled = open !== undefined;
     const newOpenState = controlled ? !!open : !openState;
 
-    this.setState({ openState: newOpenState });
+    if (!controlled) {
+      this.setState({ openState: newOpenState });
+    }
     if (!!onClick) {
       onClick({ openState: newOpenState });
     }
--- a/src/components/Expander/Expander.tsx
+++ b/src/components/Expander/Expander.tsx
@@ -33,7 +33,7 @@ interface SharedExpanderProps {
 }
 
 interface StyledExpanderContentProps extends SharedExpanderProps {
-  openState: boolean;
+  openState?: boolean;
   hidden: boolean;
 }
 
@@ -108,31 +108,61 @@ const IfTitleTag = ({
   </Fragment>
 );
 
-export class ExpanderItem extends Component<ExpanderProps> {
+class ExpanderItem extends Component<ExpanderProps> {
   state: ExpanderState = {
     openState:
       this.props.defaultOpen !== undefined ? this.props.defaultOpen : false,
   };
 
+  componentDidUpdate(prevProps: ExpanderProps, prevState: ExpanderState) {
+    if (
+      !!this.props.consumer &&
+      !!prevProps.consumer &&
+      this.props.consumer.toggleAllExpanderState !==
+        prevProps.consumer.toggleAllExpanderState
+    ) {
+      const { openState } = this.state;
+      const {
+        expanderGroup,
+        index,
+        consumer: { toggleAllExpanderState },
+        open,
+      } = this.props;
+
+      if (
+        !!expanderGroup &&
+        index !== undefined &&
+        ((open === undefined &&
+          !!openState !== toggleAllExpanderState.toState) ||
+          (open !== undefined && !!open !== toggleAllExpanderState.toState))
+      ) {
+        this.handleClick();
+      }
+    }
+
+    if (
+      prevState.openState !== this.state.openState ||
+      prevProps.open !== this.props.open
+    ) {
+      const {
+        expanderGroup,
+        index,
+        consumer: { onClick: consumerOnClick } = { onClick: undefined },
+      } = this.props;
+      if (!!expanderGroup && !!consumerOnClick && index !== undefined) {
+        consumerOnClick(index);
+      }
+    }
+  }
+
   handleClick = () => {
-    const {
-      open,
-      onClick,
-      consumer: { onClick: consumerOnClick } = { onClick: undefined },
-      index,
-      expanderGroup,
-    } = this.props;
-    if (!!expanderGroup && !!consumerOnClick && index !== undefined) {
-      consumerOnClick(index);
-    } else {
-      const openState = !this.state.openState;
-      const notControlled = open === undefined;
-      if (notControlled) {
-        this.setState({ openState });
-      }
-      if (!!onClick) {
-        onClick({ openState: notControlled ? openState : !!open });
-      }
+    const { open, onClick } = this.props;
+    const openState = !this.state.openState;
+    const controlled = open !== undefined;
+
+    this.setState({ openState: controlled ? !!open : openState });
+    if (!!onClick) {
+      onClick({ openState: controlled ? !!open : openState });
     }
   };
 
@@ -148,18 +178,13 @@ export class ExpanderItem extends Component<ExpanderProps> {
       titleProps,
       index,
       expanderGroup: dissmissExpanderGroup,
-      consumer: { openExpanders } = { openExpanders: undefined },
+      consumer,
       contentProps: { className: contentClassName, ...contentPassProps } = {
         className: undefined,
       },
       ...passProps
     } = this.props;
-    const localOpenState = () =>
-      open !== undefined ? !!open : this.state.openState;
-    const openState =
-      index !== undefined && !!openExpanders
-        ? openExpanders.includes(index)
-        : localOpenState();
+    const openState = open !== undefined ? !!open : this.state.openState;
 
     return (
       <StyledDiv
--- a/src/components/Expander/Expander.tsx
+++ b/src/components/Expander/Expander.tsx
@@ -147,10 +147,12 @@ class ExpanderItem extends Component<ExpanderProps> {
       const {
         expanderGroup,
         index,
-        consumer: { onClick: consumerOnClick } = { onClick: undefined },
+        consumer: { onExpanderOpenChange } = {
+          onExpanderOpenChange: undefined,
+        },
       } = this.props;
-      if (!!expanderGroup && !!consumerOnClick && index !== undefined) {
-        consumerOnClick(index);
+      if (!!expanderGroup && !!onExpanderOpenChange && index !== undefined) {
+        onExpanderOpenChange(index);
       }
     }
   }
--- a/src/components/Expander/ExpanderGroup.tsx
+++ b/src/components/Expander/ExpanderGroup.tsx
@@ -143,14 +143,14 @@ export class ExpanderGroup extends Component<ExpanderGroupProps> {
     });
   };
 
-  isEveryExpanderOpened = () => {
+  allExpandersOpen = () => {
     return this.props.children.length > this.state.openExpanders.length;
   };
 
   handleAllToggleClick = () => {
     this.setState({
       toggleAllExpanderState: {
-        toState: this.isEveryExpanderOpened(),
+        toState: this.allExpandersOpen(),
       },
     });
   };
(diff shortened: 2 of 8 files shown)


src/components/Expander/Expander.tsx
```diff
@@ -108,32 +108,62 @@ const IfTitleTag = ({
...
-      expanderGroup,
-    } = this.props;
-    if (!!expanderGroup && !!consumerOnClick && index !== undefined) {
-      consumerOnClick(index);
-    } else {
-      const openState = !this.state.openState;
-      const notControlled = open === undefined;
-      if (notControlled) {
-        this.setState({ openState });
+  componentDidUpdate(prevProps: ExpanderProps, prevState: ExpanderState) {
```
aappoalander, line 117: 😅 This looks super complex conditional logic, but I guess there is no way around it. Is it possible that both blocks get executed on one update so that onClick and consumerOnClick are both executed? I was trying to figure out if an infinite update loop is possible in some condition, e.g. controlled open for one expander does not allow updating one item in group?

src/components/Expander/Expander.tsx
```diff
@@ -108,32 +108,62 @@ const IfTitleTag = ({
...
+      } = this.props;
+      if (!!expanderGroup && !!consumerOnClick && index !== undefined) {
+        consumerOnClick(index);
       }
     }
+  }
+
+  handleClick = () => {
+    const { open, onClick } = this.props;
+    const openState = !this.state.openState;
```
aappoalander, line 160: this could maybe be written so that openState is resolved once based on presence of open prop and used on both setState and onClick.

src/components/Expander/ExpanderGroup.tsx
```diff
@@ -34,7 +40,7 @@ export interface ExpanderGroupProps extends OpenCloseAll {
 
 export interface ExpanderProviderState {
   onClick: (index: number) => void;
```
aappoalander, line 42: Is this really onClick event? Feels a bit confusing to have click events mixed with internal value change events. It's completely fine this way if used for handling both, user interaction and something that is either programmatic or derived from user action. Could this be renamed to onExpanderOpenChange etc?

src/components/Expander/ExpanderGroup.tsx
```diff
@@ -82,43 +99,72 @@ const OpenAllButton = ({
   onClick,
```
aappoalander, line 99: What is the difference between onClick and onClickAll? Do we need both for the API? If we do, I would at least consider the naming to be better descriptive of functionality.

src/components/Expander/ExpanderGroup.tsx
```diff
@@ -82,43 +99,72 @@ const OpenAllButton = ({
...
-    openExpanders: [],
+    openExpanders: InitialStateOfExpanders(this.props.children),
+    toggleAllExpanderState: {
+      toState:
+        InitialStateOfExpanders(this.props.children).length ===
+          this.props.children.length && this.props.children.length > 0,
+    },
   };
 
   handleClick = (index: number = 0) => {
```
aappoalander, line 133: I'd rename this since we don't handle clicks with this anymore?

src/components/Expander/ExpanderGroup.tsx
```diff
@@ -82,43 +99,72 @@ const OpenAllButton = ({
...
+      const openExpanders = prevExpanderOpen
+        ? prevOpenExpanders.filter((value) => value !== index)
+        : Array.from(new Set([...prevOpenExpanders, index]));
+      return {
+        openExpanders,
+      };
+    });
+  };
+
+  isEvenOneExpanderClose = () => {
```
aappoalander, line 146: isEvenOneExpanderClose*d* or allExpandersOpen or allExpandersClosed?

src/core/Expander/Expander.md
```diff
@@ -15,3 +15,74 @@ import { Expander } from 'suomifi-ui-components';
...
+      title="Test expander 2"
+      open={expanderTwoOpen}
+      onClick={() => setExpanderTwoOpen(!expanderTwoOpen)}
+    >
+      Test expander content 2
+    </Expander>
+    <Expander
+      title="Test expander 3"
+      open={expanderThreeOpen}
+      onClick={() => setExpanderThreeOpen(!expanderThreeOpen)}
```
aappoalander, line 82: Could this example be simplified and aligned with controlled examples for other components? E.g. add window.confirm for one of the expanders and have others just change state when controlled like it is now. Or keep one button in addition to these above... and make it suomifi-button!

src/core/Expander/Expander.test.tsx
```diff
@@ -1,19 +1,27 @@
 import React from 'react';
```
aappoalander, line 1: Good work with tests! :)

src/core/Expander/Expander.tsx
```diff
@@ -61,15 +65,55 @@ export class Expander extends Component<ExpanderProps> {
       this.props.defaultOpen !== undefined ? this.props.defaultOpen : false,
   };
 
+  componentDidUpdate(prevProps: ExpanderProps, prevState: ExpanderState) {
```
aappoalander, line 68: We should really consider ditching the component level for this kind of components or at least somehow getting rid of the duplicate logic. I guess there is not much we can do with this for the time being / within this ticket, but this is something to keep in mind.

aappoalander: And please consider all comments in component level for core level as well.

src/components/Expander/Expander.tsx
ketsappi: This is now simplified and `openState` is resolved just once and used for both.

src/components/Expander/ExpanderGroup.tsx
ketsappi: This is now changed to `onExpanderOpenChange` and seems to be a lot better 👍

src/components/Expander/ExpanderGroup.tsx
ketsappi: This is now renamed to be `handleExpanderOpenChange`

src/components/Expander/ExpanderGroup.tsx
ketsappi: This is now changed to `isEveryExpanderOpened`.

src/core/Expander/Expander.tsx
ketsappi: Could not find any comments that I have added in this PR. Which one did you have in mind?

src/core/Expander/Expander.md
ketsappi: - Open/Close all with `window.confirm` didn't work. It was stuck in infinite loop if user presses OK to confirm. Would work if used just for toggling one by one.

- Changed to use Suomifi Button instead of the basic HTML button.

src/components/Expander/Expander.tsx
ketsappi: - It is complex and maybe could be simplified. But at the moment I have no thought what to change. Maybe move the logics to separate functions?

- On the controlled example both are executed at once; onClick and consumerOnClick (now renamed to _onExpanderOpenChange_), so it should work okay™

src/components/Expander/ExpanderGroup.tsx
ketsappi: - `onClickAll` is not intended to be shown to "outside". It is a way to tell from core to components that _"Open/Close"_ all was clicked.
  - The `onClick` one is the one that components-level is used for handling the _"Open/Close all"_ . And it is not shown outside either.

Naming could be improved for sure, currently not having better and more describing ones in mind.

src/components/Expander/Expander.tsx
aappoalander: Sounds reasonable. And it is true that simplifying the logic might not be possible or might make readability even worse.

src/components/Expander/ExpanderGroup.tsx
aappoalander: Ah, so the onClickAll is a prop needed for the button to call a handler for the click event, while the onClick is a handler provided by the library user? If onClickAll is not visible on public API it's fine to leave it as is. It would also be possible to use just one onClick prop and wrap the user provided function inside another which could then do our magic before calling the user provided onClick.

aappoalander: Is it currently still possible to provide a custom button for the open/close all? If it is, does this thing still work without the wrapper function as described above?

src/core/Expander/Expander.md
aappoalander: Why did this infinite loop occur? Sounds like there might be a corner case where this might happen by accident?

src/core/Expander/Expander.tsx
aappoalander: Not sure anymore. I guess it's fine.

src/core/Expander/Expander.md
ketsappi: Not yet knowing the details how `window.confirm` works like this, so it needs more investigating.

aappoalander: > And the infinite loops is occurring with the window.confirm dialog itself, as that keep asking it over and over. Until user presses Cancel.

This might suggest a bug in the flow since once user clicks open all button > event is triggered in the expanders > controlled expander calls onClick > user clicks ok > expander group receives the new state through provider > **expander group should update it's internal state and do nothing more**

The last step might be failing and triggering a recurring update in the controlled component since onClick is called again?

ketsappi: I found the problem that caused it, it was that the function version of the setState was not used when depending on previous state.
This is now fixed and using `window.confirm`s now on the controlled example.

src/components/Expander/ExpanderGroup.tsx
ketsappi: Good find. 
It seems that even before this change, by testing on the `develop` branch, when providing custom `OpenAll` or `CloseAll` button the opening/closing all won't do anything.

aappoalander: Let's leave it for another bug ticket. Feel free to resolve the discussion once the ticket is created.

src/core/Expander/Expander.md
```diff
@@ -33,54 +33,37 @@ const [expanderThreeOpen, setExpanderThreeOpen] = React.useState(
...
-      </Button>
-    </div>
-  </div>
-
   <Expander.group OpenAll="Open all" CloseAll="Close all">
     <Expander
       title="Test expander 1"
       open={expanderOneOpen}
-      onClick={() => setExpanderOneOpen(!expanderOneOpen)}
+      onClick={() => {
```
aappoalander, line 40: I'd propose to have the first two to use uncontrolled state and leave the last one as it is now? This was a good test, but I think it should be sufficient to have just one confirm?

src/components/Expander/ExpanderGroup.tsx
```diff
@@ -82,43 +99,72 @@ const OpenAllButton = ({
...
-      ? prevOpenExpanders.filter((value) => value !== index)
-      : Array.from(new Set([...prevOpenExpanders, index]));
-    this.setState({ openExpanders });
+  handleExpanderOpenChange = (index: number = 0) => {
+    this.setState((prevState: ExpanderGroupState) => {
+      const { openExpanders: prevOpenExpanders } = prevState;
+      const prevExpanderOpen = prevOpenExpanders.includes(index);
+      const openExpanders = prevExpanderOpen
+        ? prevOpenExpanders.filter((value) => value !== index)
+        : Array.from(new Set([...prevOpenExpanders, index]));
```
aappoalander, line 139: Is there a reason to use Set? Wouldn't it work to just by using [...prevOpenExpanders, index]? Using spread inside brackets should return a new array anyways? Or does it create a shallow copy so that change detection is not triggered?

src/core/Expander/ExpanderGroup.tsx
```diff
@@ -50,25 +54,116 @@ const OpenAllButton = ({
...
+  state: ExpanderGroupState = {
+    openExpanders: InitialStateOfExpanders(this.props.children),
+    toggleAllExpanderState: {
+      toState:
+        InitialStateOfExpanders(this.props.children).length ===
+          this.props.children.length && this.props.children.length > 0,
+    },
+  };
+
+  handleExpanderOpenChange = (index: number = 0) => {
```
aappoalander, line 113: Is the default parameter value really needed here? Should only apply changes when the index is really provided?

src/components/Expander/ExpanderGroup.tsx
ketsappi: Bug ticket is now created to our system.

src/components/Expander/ExpanderGroup.tsx
ketsappi: Not sure as this was part of the previous implementation. 

It might have been used to prevent duplicate index to be added as Set takes care of that, but on the other hand it is already checked before this if the index is already in the array.

src/core/Expander/ExpanderGroup.tsx
ketsappi: Default parameter is now removed as it was not really needed, maybe never had been.

src/core/Expander/Expander.md
ketsappi: This is now changed as suggested; I removed the controlled state of the first two Expanders and left it only for the third one.

src/core/Expander/Expander.md
```diff
@@ -15,3 +15,40 @@ import { Expander } from 'suomifi-ui-components';
...
+    <Expander title="Test expander 1">
+      Test expander content 1
+    </Expander>
+    <Expander title="Test expander 2">
+      Test expander content 2
+    </Expander>
+    <Expander
+      title="Test expander 3"
+      open={expanderThreeOpen}
+      onClick={() => {
```
aappoalander, line 44: Should we use the openState parameter to set the value, not just invert the state? onClick={({openState}) => ... setExpanderThreeOpen(openState)?

aappoalander:
Tested with CRA using the examples in the MD as well as with all expanders in controlled state.

src/core/Expander/Expander.md
ketsappi: Good find. This change is now committed.

PR #331, opened 2020-05-14
[Feature] Align dropdown styles with style guide

aappoalander:
## Description

Align dropdown component colors and spacing with style guide. Add missing shadow.

## Motivation and Context

Style guide has been updated without updating the ui-components implementation. Projects mostly follow the style guide, which prevents taking the library in use in the projects. This needs to be addressed by aligning the ui-components styles with the style guide.

## How Has This Been Tested?

Tested with styleguidist and create react app project using long texts for contents.

### DIFF
--- a/src/core/Dropdown/Dropdown.baseStyles.tsx
+++ b/src/core/Dropdown/Dropdown.baseStyles.tsx
@@ -55,6 +55,8 @@ export const menuPopoverStyles = withSuomifiTheme(
     border: 0;
     padding: 0;
     white-space: normal;
+    word-break: break-word;
+    overflow-wrap: break-word;
   }
 
   & [data-reach-menu-item].fi-dropdown_item {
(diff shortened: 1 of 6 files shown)


src/core/Dropdown/Dropdown.baseStyles.tsx
```diff
@@ -52,12 +54,14 @@ export const menuPopoverStyles = withSuomifiTheme(
   & [data-reach-menu-items] {
     border: 0;
     padding: 0;
+    white-space: normal;
```
aappoalander, line 57: this probably needs defs for word-break and overflow-wrap as well.

aappoalander: And there it is!

src/core/Dropdown/Dropdown.baseStyles.tsx
```diff
@@ -7,15 +7,17 @@ export const baseStyles = withSuomifiTheme(
...
+      line-height: 1.5;
       background-color: ${theme.colors.whiteBase};
+      box-shadow: ${theme.shadows.dropdownShadow};
       cursor: pointer;
       &:before {
         content: '';
         position: absolute;
         top: 50%;
-        right: 10px;
+        right: 16px;
```
ketsappi, line 20: Would it make sense to use `insetXl` token here to get the 16px?
Or is the idea to change this after the updated tokens are released?

PR #332, opened 2020-05-15
[Bugfix] Align textinput and searchinput styles with design

LJKaski:
## Description
This PR aligns the TextInput and SearchInput styles to the design. Some larger rebuilding still remains in the searchinput, but that's outside of the scope of this PR.

While adjusting styles, a couple of related matters got sorted. The statustext element is now conditionally rendered depending on if it has content or not. Also, the button in the styleguidist example was changed to suomifi button.

## Motivation and Context
Textinput and searchinput had styles that differed from the latest designs. 

## How Has This Been Tested?
Running locally and manually testing.

### DIFF
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -19,8 +19,8 @@ export const baseStyles = withSuomifiTheme(
     flex-direction: column;
 
     & .fi-text-input_statusText_span {
+      ${theme.typography.bodySemiBoldSmall}
       font-size: 14px;
-      font-weight: 600;
       line-height: 20px;
     }
   }
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -21,6 +21,7 @@ export const baseStyles = withSuomifiTheme(
     & .fi-text-input_statusText_span {
       font-size: 14px;
       font-weight: 600;
+      line-height: 20px;
     }
   }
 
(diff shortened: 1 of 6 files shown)


src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -1,11 +1,13 @@
...
-import { input, inputContainer } from '../../theme/reset';
+import { input, inputContainer, font } from '../../theme/reset';
 
 export const baseStyles = withSuomifiTheme(
   ({ theme }: TokensAndTheme) => css`
   & .fi-text-input_label-p {
-    margin-bottom: ${theme.spacing.insetXl};
+    margin-bottom: ${theme.spacing.insetL};
+    ${font({ theme })('actionElementInnerTextBold')};
+    color: ${theme.colors.blackBase};
```
ketsappi, line 10: This was not mentioned in the description but seems to fix the issue when parent container's `color` was changing the color of the label.

LJKaski: Counted this as a style fix, so didn't mention it separately. Maybe should've. But yeah, it should be fixed by this.

src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -17,21 +19,29 @@ export const baseStyles = withSuomifiTheme(
     flex-direction: column;
 
     & .fi-text-input_statusText_span {
-      ${theme.typography.bodySemiBoldSmall}
+      font-size: 14px;
```
aappoalander, line 22: Removing the typography token also removes the font-face definition. Does this allow font-face bleeding into status text from outside or will this be nested within a block that has more specific typography rules?

LJKaski: Good point. Better safe than sorry. Now fixed.

PR #333, opened 2020-05-19
[Feature] Add new instructions to Readme and FAQ

LJKaski:
## Description
Updates for github documentation to cover SSR and strict content security policy cases as well as some dependency version requirements.

## Motivation and Context
Having up-to-date information about implementing the library helps our users to implement it and reduces the need for support actions from our side.

## How Has This Been Tested?
Content changes, so no testing done.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -6,14 +6,34 @@
 
 2. ## Content-Security-Policy (CSP) problems?
 
-   You need to configure DEV env to allow
+   If strict content security policy is not required, you can simply use
 
    ```bash
    style-src: 'unsafe-inline'
    ```
 
-   And also might need to set `dangerously omit style nonce` to `true`.
+   to allow the use of dynamic inline styling.
+
+   You might also need to set `dangerously omit style nonce` to `true`.
+
+   ### Strict content security policies
+
+   Strict content security policy does not allow dynamic styling without extra safety measures. Styled components uses `__webpack_nonce__` to provide these measures.
+
+   Add `__webpack_nonce__ = '<some generated nonce>'` in the beginning of your entry file. It must be defined before any other scripts. One way to do this is to introduce the nonce as above in a new file and import that in your entry file. More information can be found in [webpack documentation.](https://webpack.js.org/guides/csp/)
+
+   Use [helmet](https://helmetjs.github.io/docs/csp/) or a similar tool to provide the generated nonce to the browser. See our [simplified example setup](https://github.com/ketsappi/strict-csp-expressed-with-suomifi-ui-components) for reference.
+
+   Styled components now has access to the compiled value of `__webpack_nonce__` through the window object, which allows styled-components to create necessary nonce values for changing styles, ensuring the integrity of the changes and meeting strict content security policy rules.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -41,6 +41,8 @@ You should also install the following dependencies, if your project does not alr
 yarn add styled-components
 ```
 
+- If using typescript, version 3.5 or above is required.
```
ketsappi, line 44: Maybe change to ` typescript` -> `TypeScript`

FAQ.md
```diff
@@ -1,19 +1,37 @@
...
+
+```bash
+style-src: 'unsafe-inline'
+```
+
+to allow the use of dynamic inline styling.
+
+### Strict content security policies
+
+Strict content security policy does not allow dynamic styling without extra safety measures. Styled components uses `__webpack_nonce__` to provide these measures.
```
aappoalander, line 19: Not sure how styled components should be written, but at least this way is used in their website: "As you can see, styled-components"

FAQ.md
```diff
@@ -1,19 +1,37 @@
...
+
+Add `__webpack_nonce__ = '<some generated nonce>'` in the beginning of your entry file. It must be defined before any other scripts. One way to do this is to introduce the nonce as above in a new file and import that in your entry file. More information can be found in [webpack documentation.](https://webpack.js.org/guides/csp/)
```
aappoalander, line 21: '<some generated nonce>' => maybe would be better to say "some generated uuid" or somehow point out already the nature of nonce.

FAQ.md
```diff
@@ -1,19 +1,37 @@
...
+
+Use [helmet](https://helmetjs.github.io/docs/csp/) or a similar tool to provide the generated nonce to the browser. See our [simplified example setup](https://github.com/ketsappi/strict-csp-expressed-with-suomifi-ui-components) for reference.
+
+Styled components now has access to the compiled value of `__webpack_nonce__` through the window object, which allows styled-components to create necessary nonce values for changing styles, ensuring the integrity of the changes and meeting strict content security policy rules.
```
aappoalander, line 25: Styled components => Styled-components

aappoalander: which allows styled-components to create necessary nonce values for changing styles => which allows styled-components to add nonce to dynamically inserted content/styles.

FAQ.md
```diff
@@ -1,19 +1,37 @@
...
+
+If you are using TypeScript, make sure you have the latest version of `@types/webpack-env` as it containts the `__webpack_nonce__`.
+
+Also notice that `__webpack_nonce__` does not work with hot reload in local development environment.
```
aappoalander, line 29: Hot Reloading aka Hot "Module" Reloading, but is also fine as is. Webpack related https://webpack.js.org/concepts/hot-module-replacement/

FAQ.md
LJKaski: Nonce itself already is both a regular English word for something intended for single-use, so I wouldn't introduce another one for that. I'll try to explain it somehow since it's not exactly the most common term to be used.

PR #335, opened 2020-05-20
[Feature] Update textinput styles to include inner shadow

LJKaski:
## Description
This PR adds inset box shadows to textinput and searchinput as well as fixes an issue with the placement of the searchinput icon.

Since the shadow being used in dropdown is exactly the same, it was renamed to inputBoxShadow and implemented in the textinput and searchinput as well as in dropdown.

## Motivation and Context
Implementations need to match styleguide in order to be usable in projects.

## How Has This Been Tested?
Manual testing plus automated tests.

### DIFF
--- a/src/core/theme/colors.ts
+++ b/src/core/theme/colors.ts
@@ -88,7 +88,9 @@ export const shadows = {
   panelShadow: `0 1px 2px 0 ${alphaHex(0.14)(
     colors.blackBase,
   )}, 0 1px 5px 0 ${alphaHex(0.12)(colors.blackBase)}`,
-  inputBoxShadow: `0 1px 2px 0 ${alphaHex(0.1)(colors.brandBase)} inset`,
+  actionElementBoxShadow: `0 1px 2px 0 ${alphaHex(0.1)(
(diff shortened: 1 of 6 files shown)


src/core/theme/colors.ts
```diff
@@ -88,7 +88,7 @@ export const shadows = {
   panelShadow: `0 1px 2px 0 ${alphaHex(0.14)(
     colors.blackBase,
   )}, 0 1px 5px 0 ${alphaHex(0.12)(colors.blackBase)}`,
-  dropdownShadow: `0 1px 2px 0 ${alphaHex(0.1)(colors.brandBase)} inset`,
+  inputBoxShadow: `0 1px 2px 0 ${alphaHex(0.1)(colors.brandBase)} inset`,
```
aappoalander, line 91: Is dropdown / select considered an input? Input box shadow might not be a good name for this since it's also used for dropdown. What about action element box shadow? Form element box shadow? I will leave it for you to decide.

LJKaski: Changed this to actionElementBoxShadow.

PR #336, opened 2020-05-25
[Feature] Align toggle styles with style guide

aappoalander:
## Description
This PR aligns Toggle component styles with style guide. Focus styles have now rounded corners and only focus on toggle button / input instead of label also. Positioning of the toggle knob has been slightly adjusted. A stronger shadow is also present.

Due to the changes, the component width changes slightly.

## How Has This Been Tested?
Tested with Create react app Typescript project and styleguidist build.

### DIFF
--- a/src/core/Form/Toggle/Toggle.baseStyles.tsx
+++ b/src/core/Form/Toggle/Toggle.baseStyles.tsx
@@ -50,7 +50,7 @@ export const baseStyles = withSuomifiTheme(
   & .fi-toggle_icon {
     width: ${iconWidth};
     height: ${iconHeight};
-    margin-right: ${theme.spacing.insetM};
+    margin-right: ${theme.spacing.insetL};
     vertical-align: bottom;
     overflow: visible;
     transform: translateY(-0.1em);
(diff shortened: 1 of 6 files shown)


src/core/Form/Toggle/Toggle.baseStyles.tsx
```diff
@@ -8,86 +8,113 @@ const svgPrefix = 'icon-toggle_svg__';
...
         outline: 0;
-        ${focus({ theme, noPseudo: true })}
+        & .fi-toggle_icon-container {
+          ${focus({ theme, noPseudo: true })}
+          position: absolute;
+          &:after {
+            border-radius: 14px;
+            right: -4px;
+            left: -4px;
+          }
```
LJKaski, line 61: Would it be cleaner to store this bit in a variable such as toggleFocusStyles and use that instead of repeating it for both variants? I think a case can be made for both approaches since it's only used here.

aappoalander: This should now be fixed. Left the focus call out of the variable since it's easier to read that way in my opinion and the typings are a mess - including it would require a long custom type for some reason.

aappoalander:
Tested with Safari, Chrome, Brave and Firefox and all seem to work in a consistent manner now.

PR #338, opened 2020-05-26
[Fix] Grouped expanders corners rounded and separator between

ketsappi:
## Description

- Fixed the Expander's corner rounding when inside ExpanderGroup
- Added separator between Expanders when inside ExpanderGroup

## Details

Showing the rounded corners only on top and bottom like in the [Material-UI](https://material-ui.com/components/expansion-panels/)

- This is not currently yet in the designs but was discussed for this change to be good to go
- Previously the rounding in the designs were all over the place and it was not possible to do it with just the CSS. It would have required also code changes to get it to match the design 100% and not really worth the effort as Material-UI way is simplistic and good looking.

## Motivation and Context

To better match with the design and component to look good.

## How Has This Been Tested?

- Locally in the Chrome

## Screenshots 

Explanation to the screenshot (with the changes of this PR)
- Rounded corners surrounded by circles
- Corners without rounding marked by rectangles 

expandergroup_kulmat

## Release notes for this PR
- Rounded top and bottom corners for Expander groups and separator between as detailed in the design. (#338)

### DIFF
--- a/src/components/Expander/Expander.tsx
+++ b/src/components/Expander/Expander.tsx
@@ -39,10 +39,7 @@ interface StyledExpanderContentProps extends SharedExpanderProps {
 
 export const StyledExpanderContent = styled(
   ({ openState, className, ...passProps }: StyledExpanderContentProps) => (
-    <StyledDiv
-      {...passProps}
-      className={classnames(className, baseClassName)}
-    />
+    <StyledDiv {...passProps} className={classnames(className)} />
   ),
 )`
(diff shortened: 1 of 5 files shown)


aappoalander:
Seems to be working as intended. Firefox on Mac has ugly bold fonts but that is probably unrelated.

ketsappi:
> Seems to be working as intended. Firefox on Mac has ugly bold fonts but that is probably unrelated.

It seems that it is not limited to this component. Almost every component and texts look different with the Firefox. Especially when using some `font-weight`.

PR #357, opened 2020-06-10
[Fix] Button style fixes

ketsappi:
## Description

- Fixes the invert text shadow to have opacity of 50%
- Disabled Button text-shadow to have opacity of 50%
- Disabled Button icon is now showing correct cursor, `not-allowed`
- Disabled inverted button opacity changed from 50% to 41% to match with the design
- Button icon color is now using the `currentColor` so that it will use same color as the text

## Motivation and Context

Wanted to match the Button styles with the design.

## How Has This Been Tested?

Locally ran in Styleguidist on Brave.

## Release notes

**Fix**
- Button text-shadow and icon color changes.

### DIFF
--- a/src/core/Button/Button.baseStyles.tsx
+++ b/src/core/Button/Button.baseStyles.tsx
@@ -2,6 +2,7 @@ import { css } from 'styled-components';
 import { withSuomifiTheme, TokensAndTheme, SuomifiThemeProp } from '../theme';
 import { ButtonProps } from './Button';
 import { element, focus, button } from '../theme/reset';
+import { alphaHex } from '../../utils/css';
 
 const invertedStyles = ({ theme }: SuomifiThemeProp) => css`
   &.fi-button--inverted {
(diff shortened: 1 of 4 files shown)


src/core/Button/Button.baseStyles.tsx
```diff
@@ -139,6 +140,9 @@ export const baseStyles = withSuomifiTheme(
       margin-left: ${theme.spacing.insetM};
     }
   }
+  &.fi-button--disabled > .fi-button_icon {
```
aappoalander, line 143: Could this be nested under disabled styles in starting in line 108?

src/core/Button/Button.tsx
```diff
@@ -125,11 +99,7 @@ class ButtonWithIcon extends Component<ButtonProps & InternalTokensProp> {
           <Icon
             mousePointer={true}
             icon={iconRight}
-            color={iconColor({
-              tokens,
-              disabled,
-              invert: secondaryOrTertiary,
-            })}
```
aappoalander, line 132: This makes a lot of code redundant! Great! 👍

src/core/Button/Button.baseStyles.tsx
ketsappi: It was first there but then it complained about the descendant style rule.

PR #361, opened 2020-06-15
[Bugfix] Checkbox statustext margins

LJKaski:
## Description
Checkbox status text element styles were off. The top margin was applied only with the error status. It now applies to the statustext element regardless of the actual status.

Also found a naming error in the tests while doing this and fixed it since it was a minor thing and didn't affect anything else. 

## Motivation and Context
We want the library to be as bug-free as possible. 

## How Has This Been Tested?
Running locally and automated tests.

## Release notes
Fixed a bug in checkbox status text spacing

### DIFF
--- a/src/core/Form/Checkbox/Checkbox.test.tsx
+++ b/src/core/Form/Checkbox/Checkbox.test.tsx
@@ -70,8 +70,8 @@ const DisabledTestCheckbox = createTestCheckbox(
 );
 
 test('Calling render with the same component on the same container does not remount', () => {
-  const toggleInputRendered = render(RegularTestCheckbox);
-  const { getByTestId, container, rerender } = toggleInputRendered;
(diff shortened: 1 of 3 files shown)


src/core/Form/Checkbox/Checkbox.test.tsx
```diff
@@ -70,8 +70,8 @@ const DisabledTestCheckbox = createTestCheckbox(
 );
 
 test('Calling render with the same component on the same container does not remount', () => {
-  const toggleInputRendered = render(RegularTestCheckbox);
-  const { getByTestId, container, rerender } = toggleInputRendered;
+  const checkboxRendered = render(RegularTestCheckbox);
+  const { getByTestId, container, rerender } = checkboxRendered;
```
ketsappi, line 74: Good fix to improve the test readability and to prevent confusion 👍

PR #363, opened 2020-06-15
[Feature] Dropdown value property

aappoalander:
### 🚨 Breaking Changes 🚨

This PR changes the underlying Reach-UI component for dropdown to Listbox and by doing so, adds support for e.g., value, defaultValue and onChange properties. Also controlled state is now supported. Action menu functionality is still supported by use of visual placeholder.

- Add name, value and defaultValue props
- Add visualPlaceholder, alwaysShowVisualPlaceholder props
- Add onChange prop
- dropdownButtonProps type changed to OptionalListboxButtonProps
- DropdownPopoverProps type changed to OptionalListboxPopoverProps
- menuPopoverComponent replaced by listboxPopoverComponent and prop type changed to OptionalListboxPopoverProps
- dropdownItemProps replaced by dropdownItemClassName and type changed to string.
- changeVisualPlaceholderToSelection prop removed
- Dropdown.item onSelect prop removed from DropdownItemProps.

## Motivation and Context

Previous implementation did not function as select component for form implementations.

## How Has This Been Tested?

Tested with React Styleguidist, Create React App TS project and SSR test project.

## Release notes
- Add support for dropdown form use cases. Multiple breaking changes, see #363


src/components/Dropdown/Dropdown.tsx
```diff
@@ -76,77 +88,81 @@ export interface DropdownProps {
...
 }
 
 export class Dropdown extends Component<DropdownProps> {
-  state: DropdownState = { selectedName: undefined };
+  state: DropdownState = {
+    selectedValue: !!this.props.defaultValue
+      ? this.props.defaultValue
+      : !!this.props.value
+      ? this.props.value
+      : undefined,
```
LJKaski, line 114: could this just be optionA || optionB || optionC ?

aappoalander: Nope, since also non-truthy values need to be supported e.g. empty string.

aappoalander: This was changed to honor value over defaultValue even if provided untruthy value e.g. value={undefined}.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -76,77 +86,79 @@ export interface DropdownProps {
...
+  listboxPopoverComponent?: React.ComponentType<OptionalListboxPopoverProps>;
+  dropdownItemClassName?: string;
   /** DropdownItems */
   children?:
-    | Array<ReactElement<DropdownPopoverItems>>
-    | ReactElement<DropdownPopoverItems>;
+    | Array<ReactElement<DropdownItemProps>>
+    | ReactElement<DropdownItemProps>;
+
+  onChange?(newValue: string): void;
```
LJKaski, line 102: Should this have a description as well?

src/components/Dropdown/Dropdown.tsx
```diff
@@ -179,13 +198,31 @@ export class Dropdown extends Component<DropdownProps> {
...
+      onChange,
+      name,
+      value: selectedValue || '',
+    };
+
+    const listboxDisplayValue = alwaysShowVisualPlaceholder
+      ? visualPlaceholder
+      : !!selectedValue
+      ? ''
+      : visualPlaceholder;
```
LJKaski, line 224: If alwaysShowVisualPlaceholder is false and selectedValue has a value, this seems to return an empty string. Why is that?

aappoalander: Reach handles empty string as no value and uses the selected item node as button display value. We still want to explicitly set the prop to make it work properly => if always show is set, we replace the button value with the placeholder alway, if it's not, we display the selected node (can be string or react element) and if nothing is selected, we show the placeholder.

src/components/Dropdown/Dropdown.tsx
aappoalander: Yes, have to add documentation for it.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -60,8 +60,18 @@ export interface DropdownProps {
    * @default uuidV4
    */
   id?: string;
+  /** Name used for input's form value. */
+  name?: string;
```
ketsappi, line 64: Should we explicitly show it like this on JSDoc for other form elements also? 
Now some supports this `name` prop but it is not shown in the API doc on styleguidist.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -31,25 +30,26 @@ export const dropdownClassNames = {
 export interface DropdownLabelProps extends HtmlLabelProps {}
 
 export interface DropdownItemProps {
-  /** Operation to run on select */
-  onSelect: () => void;
+  value: string;
   /** Item content */
   children: ReactNode;
+  className?: string;
```
ketsappi, line 36: Should these `value` and `className` props have also JSDoc?

src/components/Dropdown/Dropdown.tsx
aappoalander: Good point, not sure but since dropdown does not extend upon other props like e.g. div, would feel a bit weird not to show it in the public API. Extended prop interfaces show the definition as first thing in IDEs when inspected. But I'm open for both alternatives.

src/components/Dropdown/Dropdown.tsx
aappoalander: Yes, added.

src/components/Dropdown/Dropdown.tsx
aappoalander: Added.

src/components/Dropdown/Dropdown.tsx
ketsappi: Yes, let's show the prop in public API. We need just to make the other places to follow same logic.

PR #367, opened 2020-06-29
[Fix] Changes to examples

ketsappi:
## Description

Small fixes to different components examples:
- Dropdown
- Text
- Toggle

No changes to actual implementation of the components, just the examples shown in the styleguidist.

## Motivation and Context

Examples were not using the "rules" they should have.

The code in Markdown file for the styleguidist is little problematic as IDE:s doesn't seem to be able to find the issues so easily as with "normal" code. These problems were found when tested to use the example codes in a separate CRA TS project.

If someone tests now to copy these examples straight from the styleguidist, they have bigger change of success.

## How Has This Been Tested?

- Styleguidist run locally
- CRA TS project

## Release notes

- Fixed few component examples [Dropdown, Toggle, Text] (#367)

### DIFF
--- a/src/core/Text/Text.md
+++ b/src/core/Text/Text.md
@@ -3,7 +3,7 @@ import { Paragraph, Text } from 'suomifi-ui-components';
 
 <>
   <Paragraph>Paragraph text</Paragraph>
-  <Paragraph marginBottomSpacing="xl">
+  <Paragraph marginBottomSpacing="s">
     <Text.lead>Leading text</Text.lead>
   </Paragraph>
   <Paragraph>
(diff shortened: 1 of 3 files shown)


src/core/Text/Text.md
```diff
@@ -3,7 +3,7 @@ import { Paragraph, Text } from 'suomifi-ui-components';
 
 <>
   <Paragraph>Paragraph text</Paragraph>
-  <Paragraph marginBottomSpacing="insetXl">
+  <Paragraph marginBottomSpacing="xl">
```
LJKaski, line 6: This is quite a bit larger than insetXl. Works like this as well, but if you want this to stay about the same, then use s or m as the value.

ketsappi: Yes, I can change it to be `s` or `m`. Thought that it has no real purpose as it is just as an example which could use every possible token available.

ketsappi: This is now changed to use the `s` token. It was the closest size of the original author's wanted size.

PR #371, opened 2020-07-07
[Fix] Disable generated button with disabled-attribute

jteuho:
## Description

Disable generated button -element properly, if instantiating Button with disabled={true}.

## Related Issue

Closes #

## Motivation and Context

Current implementation doesn't actually disable button, it only makes the button look like disabled.

### DIFF
--- a/src/core/Button/Button.test.tsx
+++ b/src/core/Button/Button.test.tsx
@@ -19,6 +19,23 @@ test('calling render with the same component on the same container does not remo
   expect(getByTestId('nottub').textContent).toBe('Test two');
 });
 
+test('disabled button is disabled', () => {
+  const button = render(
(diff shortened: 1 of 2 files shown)


src/core/Button/Button.test.tsx
```diff
@@ -19,6 +19,23 @@ test('calling render with the same component on the same container does not remo
...
+  expect(getByTestId('button')).toBeDisabled();
+
+  rerender(
+    <Button data-testid="button" disabled={false}>
+      Test
+    </Button>,
+  );
+  expect(getByTestId('button')).toBeEnabled();
+});
+
```
ketsappi, line 38: Thanks also for the test 👍

PR #375, opened 2020-07-31
[Feature] name and value props for Checkbox

ketsappi:
## Description

- Adds `name` and `value` props for Checkbox.
  - Checkbox is now usable when using in form.
- Small refactoring to tests.
- More tests added.

## Motivation and Context

Checkbox was not working correctly when used in forms as the value and name was not available for form.

## How Has This Been Tested?
- Locally run on CRA-TS project
- Locally run on CRA-JS project with form

## Release notes

Feature
- Adds `name` and `value` props for Checkbox.

### DIFF
--- a/src/components/Form/Checkbox.tsx
+++ b/src/components/Form/Checkbox.tsx
@@ -65,6 +65,7 @@ export interface CheckboxProps {
    * @default uuidV4
    */
   id?: string;
+  name?: string;
 }
 
 export interface CheckboxState {
@@ -111,6 +112,7 @@ export class Checkbox extends Component<CheckboxProps> {
       hintText,
       status,
       statusText,
+      name,
       ...passProps
     } = this.props;
     const { checkedState } = this.state;
@@ -146,6 +148,7 @@ export class Checkbox extends Component<CheckboxProps> {
           {...newCheckboxInputProps}
           type="checkbox"
           aria-describedby={infoElementIds}
+          name={name}
         />
         <HtmlLabel
           htmlFor={id}
--- a/src/components/Form/Checkbox.tsx
+++ b/src/components/Form/Checkbox.tsx
@@ -141,7 +141,7 @@ export class Checkbox extends Component<CheckboxProps> {
       className: checkboxBaseClassNames.input,
       onChange: this.handleClick,
       name,
-      value,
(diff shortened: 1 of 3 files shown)


src/components/Form/Checkbox.tsx
```diff
@@ -65,6 +65,8 @@ export interface CheckboxProps {
    * @default uuidV4
    */
   id?: string;
+  name?: string;
+  value?: string;
```
LJKaski, line 69: Do we want to describe these as well?

ketsappi: Yeah, that is a good question. As those are "basic/default" attributes for form element I was not sure if it is needed to have explicit documentation. 

As also other component, maybe including checkbox might extends some HtmlxxxElement and supports all the attributes from there if it is given to as prop, if I read it correctly.

LJKaski: Yeah, I think it has the basic HMTL input props. I think either case (to document or not to document) could be argued. They are clearly something that a programmer using the component expects to be there. I guess the question is that do we want to signal specifically that we support these props.

ketsappi: True. Maybe it could be good that we explicitly tell these `name` and `value` props in our documentation by providing the JSDoc for them. 

Should 
```typescript
/** Name **/
  name?: string;
/** Value **/
  value?: string
```
be enough?

LJKaski: For that purpose I think those are just fine :)

ketsappi: This is now changed.

src/components/Form/Checkbox.tsx
```diff
@@ -136,6 +142,8 @@ export class Checkbox extends Component<CheckboxProps> {
       checked: !!checkedState,
       className: checkboxBaseClassNames.input,
       onChange: this.handleClick,
+      name,
```
aappoalander, line 145: Since name is optional, this could results into name="" in DOM. Probably no side effects though.

ketsappi: According to the snapshot test without the `name` prop, it does not end as `name=""` in the DOM.

PR #376, opened 2020-07-31
[Feature] Textarea component

ketsappi:
## Description

New Textarea component added.

## Motivation and Context

Component was created to satisfy the need of having possibility to enter multiline text. First idea was to make it just an variant of TextInput

## How Has This Been Tested?
- Tested locally with CRA-TS project
- Tested locally with CRA-JS project in form

## Screenshots

image

## Release notes

Feature
- Textarea component.


src/components/Form/Textarea.tsx
```diff
@@ -0,0 +1,139 @@
...
+    );
+  }
+}
+
+const StyledBaseTextarea = styled((props: TextareaProps) => (
+  <BaseTextarea {...props} />
+))`
+  & .${textareaClassNames.textarea} {
+    resize: ${(props) => props.resize || 'vertical'};
+  }
```
ketsappi, line 132: This was done like this because the `resize` need to be added with CSS to the textarea.

ketsappi: If we would like to add also `minWidth` and `minHeight` props for the **Textarea** component, this is the place to add those also.

src/components/Form/Textarea.tsx
```diff
@@ -0,0 +1,139 @@
...
+    const onClickProps = !!disabled ? {} : { onMouseDown: onClick };
+
+    const hideLabel = labelMode === 'hidden';
+    const id = idGenerator(propId);
+    const statusTextId = `${idGenerator(propId)}-statusText`;
+    const hintTextId = `${idGenerator(propId)}-hintText`;
+
+    return (
+      <HtmlLabel className={classnames(baseClassName, className, {})}>
+        {hideLabel ? (
```
aappoalander, line 89: While this is still very readable, would it be even better if these parts were introduced as consts before the render method?

src/components/Form/Textarea.tsx
```diff
@@ -0,0 +1,139 @@
...
+          <VisuallyHidden>{labelText}</VisuallyHidden>
+        ) : (
+          <Paragraph>
+            <HtmlSpan className={textareaClassNames.label}>
+              {labelText}
+            </HtmlSpan>
+            {optionalText && (
+              <HtmlSpan className={textareaClassNames.optionalText}>
+                {`(${optionalText})`}
```
aappoalander, line 98: Is string interpolation necessary here?

src/core/Form/Textarea/Textarea.baseStyles.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+        &:focus {
+          ${focus({ theme, noPseudo: true, variant: 'boxShadow' })}
+        }
+        ::placeholder {
+          font-style: italic;
+        }
+      }
+
+      & .fi-textarea_statusText {
```
aappoalander, line 48: Was it so that there is no token for this text type?

src/components/Form/Textarea.tsx
aappoalander: If I remember correctly, some cases have been handled by using props when generating the baseStyles, so this could be also done in baseStyles.tsx. Not sure which is better though.

src/components/Form/Textarea.tsx
ketsappi: Not necessarily, but when changing it to: `({optionalText})`. Then our eslint rules will complain: `{optionalText} must be placed on a new line eslintreact/jsx-one-expression-per-line`
And if trying to put them on new line, autoformat will change it back to the starting state `({optionalText})`

And in the DOM it will show like this:
image

But it does look same visually for the user though.

src/components/Form/Textarea.tsx
ketsappi: Yes, that is true. I guess it is doable to make the same logic in the baseStyles.

I picked one place where prop is looked in baseStyles for example:
https://github.com/vrk-kpa/suomifi-ui-components/blob/716065208128775fad8928ae5d0ef3bb5b196903/src/core/Text/Text.baseStyles.ts#L7-L11

src/components/Form/Textarea.tsx
ketsappi: Yeah, it might be even better to introduce those before render. Good idea 👍

src/components/Form/Textarea.tsx
aappoalander: Ok, we have bad linter rules 😓  In this case template literal solution is better in my opinion too.

src/components/Form/Textarea.tsx
```diff
@@ -0,0 +1,139 @@
...
+        )}
+        {hintText && (
+          <HtmlSpan className={textareaClassNames.hintText} id={hintTextId}>
+            {hintText}
+          </HtmlSpan>
+        )}
+        <HtmlTextarea
+          id={id}
+          className={textareaClassNames.textarea}
+          disabled={disabled}
```
aappoalander, line 111: How does this work with screen readers and keyboards? Especially tab-navigation?

src/components/Form/Textarea.tsx
aappoalander: In my opinion the place to add the classnames logic does not matter that much, but I'd like to keep the css in the baseStyles file.

src/core/Form/Textarea/Textarea.baseStyles.tsx
ketsappi: I think there was not matching token as also Checkbox does it this way.

LJKaski: Yeah, I think the token was left in the air waiting for implementation and styling details from both us and the designers.

src/components/Form/Textarea.tsx
ketsappi: It seems to work as expecting; when using Tab to move the focus to textarea it will read first the label (and status text and hint texts if provided) then the contents of the textarea itself.
To get out of the textarea, user can press Esc to exit "edit mode" to continue using Tab to navigate.

src/core/Form/Textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+    });
+  });
+
+  describe('resize', () => {
+    it('allows only vertical resizing', () => {
+      const { getByRole } = render(
+        <Textarea labelText="label" resize="vertical" />,
+      );
+      const textarea = getByRole('textbox') as HTMLTextAreaElement;
+      expect(textarea).toHaveStyle('resize: vertical');
```
aappoalander, line 180: Is it possible that there could be style like this (e.g. problematic future change):
`resize: vertical;
resize: both`
?

Not sure how the matcher works, does it look for applied styles or style definitions...

src/core/Form/Textarea/Textarea.test.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+    it('does not allow resizing', () => {
+      const { getByRole } = render(
+        <Textarea labelText="label" resize="none" />,
+      );
+      const textarea = getByRole('textbox') as HTMLTextAreaElement;
+      expect(textarea).toHaveStyle('resize: none');
+    });
+  });
+});
```
aappoalander, line 207: Excellent work with the tests overall! 🎉

src/core/Form/Textarea/Textarea.test.tsx
ketsappi: In current way this test checks if it has `resize: vertical` style, but does not check if the style is tried to applied with other value. 

So in this case, it could be wise to add new expects to make sure that only one exists, e.g

```js
 expect(textarea).not.toHaveStyle('resize: both');
// and others

```

src/components/Form/Textarea.tsx
ketsappi: I pushed the change were logic are now done in separate functions for labelText, hintText and statusText.

src/components/Form/Textarea.tsx
aappoalander: Even if disabled={true}?

src/core/Form/Textarea/Textarea.test.tsx
aappoalander: Maybe just test if it breaks if you override the desired style with another option?

src/components/Form/Textarea.tsx
ketsappi: No, just like **TextInput, Checkbox, Toggle and LanguageMenu (+ Radiobutton)** it can not have focus when it is disabled with Tab.
As this is also related to that if **Button** can receive focus when disabled, we should then adjust all to work the same way if so decided. 

Maybe do the change for all the components in the same PR to have them work the same when decisions have been made?

src/core/Form/Textarea/Textarea.test.tsx
ketsappi: Test seems to broke if trying to apply the style like this:

```jsx
<Textarea
  labelText="label"
  resize="vertical"
  style={{ resize: 'both' }}
/>
```

Jest:
```
    - Expected

    - resize: vertical;
    + resize: both;
```

Or did you have some other way in your mind for overriding the style?

src/components/Form/Textarea.tsx
aappoalander: Sure, just wanted to make sure. Maybe using the standard disabled is a better way to go since we cannot know all implications of using an alternative way.

src/components/Form/Textarea.tsx
```diff
@@ -0,0 +1,159 @@
...
+   */
+  status?: 'default' | 'error';
+  /** Status text to be shown below the component and hint text. Use e.g. for validation error */
+  statusText?: string;
+  /** Resize mode of the textarea
+      'both' | 'vertical' | 'horizontal' | 'none'
+      @default 'vertical' 
+   */
+  resize?: 'both' | 'vertical' | 'horizontal' | 'none';
+  optionalText?: string;
```
LJKaski, line 57: Took me a moment to grasp what this is. It needs to be documented at the very least. Maybe we should/could come up with a common naming and practice for this prop to use across the components.

src/components/Form/Textarea.tsx
```diff
@@ -0,0 +1,159 @@
...
+      ...passProps
+    } = this.props;
+
+    const onClickProps = !!disabled ? {} : { onMouseDown: onClick };
+    const id = idGenerator(propId);
+
+    return (
+      <HtmlLabel className={classnames(baseClassName, className, {})}>
+        {this.labelText()}
+        {this.hintText()}
```
LJKaski, line 130: This is a different approach than elsewhere, but I kinda like it especially with our plan to implement these as separate components 👍

src/components/Form/Textarea.tsx
ketsappi: The `resize` prop is now taken into account on the **baseStyles** instead in the recent commit.

LJKaski: I agree that it would be a clearer and cleaner practice to keep the css in one place, even if it requires some extra lines.

src/components/Form/Textarea.tsx
ketsappi: Good find, it seems to be missing JSDoc.

Yes, this **OptionalText** is one of the "internal components"-idea that is going to be common in multiple components.

ketsappi: JSDoc is now added.
Also changed that if the `labelMode` is **"hidden"** then the optional text is also added. Now it did not exist.

LJKaski:
In the designs the font inside the input seems to be a suomifi font, but in the current implementation it looks like the default font of the element. I think this should be addressed.

ketsappi:
> In the designs the font inside the input seems to be a suomifi font, but in the current implementation it looks like the default font of the element. I think this should be addressed.

This should be now fixed with one of the recent commits.

src/core/Form/Textarea/Textarea.test.tsx
aappoalander: This or a change in baseStyles, but I guess it does not matter where the override is coming from.

src/components/Form/Textarea.tsx
aappoalander: I propose we leave this decision to be made later on and stick with the current implementation for now.

src/components/Form/Textarea.tsx
LJKaski: I think this one can be resolved.

PR #377, opened 2020-07-31
[Feature] Radiobutton

ketsappi:
## Description

New component Radiobutton (+ RadiobuttonGroup and RadiobuttonDivider)

## Motivation and Context

To answer the need and designs of a radiobutton component.

## How Has This Been Tested?
- Locally in CRA-TS project
- Locally in CRA-JS project in form

## Screenshots
image

## Release notes

Feature
- Radiobutton component


.styleguidist/styleguidist.sections.js
```diff
@@ -8,26 +8,27 @@ const primitiveComponents = [
...
           name: component[1],
           underName: `${component[0]}/${component[1]}`,
         })
       : getComponent({ name: component }),
   );
-const getComponentWithVariants = component => variants => [
+const getComponentWithVariants = (component) => (variants) => [
   getComponent({ name: component }),
-  ...variants.map(variant =>
+  ...variants.map((variant) =>
```
ketsappi, line 31: These changes was done automatically by the Prettier I presume.

src/components/Form/Radiobutton.tsx
```diff
@@ -0,0 +1,156 @@
...
+  disabled?: boolean;
+  /**
+   * 'small' | 'large'
+   * @default small
+   */
+  variant?: RadiobuttonVariant;
+  radiobuttonGroup?: boolean;
+  consumer?: RadiobuttonGroupProviderState;
+  checked?: boolean;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
```
ketsappi, line 53: These props are for using the Radiobutton in group and therefore do not have public JSDoc. We should "enforce" of using RadiobuttonGroup instead of manually doing the same.

src/components/Form/RadiobuttonGroup.tsx
```diff
@@ -0,0 +1,159 @@
...
+import classnames from 'classnames';
+import { HtmlDiv, HtmlSpan } from '../../reset';
+import { VisuallyHidden } from '../Visually-hidden/Visually-hidden';
+import { idGenerator } from '../../utils/uuid';
+
+import { RadiobuttonProps } from './Radiobutton';
+
+const baseClassName = 'fi-radiobuttongroup';
+const radiobuttonGroupClassNames = {
+  container: `${baseClassName}_container`,
```
ketsappi, line 11: Is this `_container` class useless? Should we remove it?

src/components/Form/RadiobuttonGroup.tsx
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * Unique id
+   * If no id is specified, one will be generated using uuid
+   * @default uuidV4
+   */
+  id?: string;
+  /** Name for the group; this will be set to Radiobuttons */
+  name: string;
+  /** Value that is selected from group, should match with the value given to Radiobutton */
+  value?: string;
```
ketsappi, line 42: `value` is basically required for radiobutton to work correctly at least in the uses cases that I thought, but it is optional for now. We may want to change this if needed.

src/components/Form/RadiobuttonGroup.tsx
```diff
@@ -0,0 +1,159 @@
...
+      defaultValue: dismissDefaultValue,
+      ...passProps
+    } = this.props;
+    const { selectedValue } = this.state;
+    const hideLabel = labelMode === 'hidden';
+
+    const id = idGenerator(propId);
+    const labelId = `${idGenerator(propId)}-label`;
+    const hintTextId = `${idGenerator(propId)}-hintText`;
+    const labelledBy = [labelId, ...(hintText ? [hintTextId] : [])].join(' ');
```
ketsappi, line 120: Joining with help of spreading to prevent of adding empty class (basically just _space_) when hintText was not provided.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
```diff
@@ -0,0 +1,159 @@
+import { css } from 'styled-components';
+import { withSuomifiTheme, TokensAndTheme } from '../../theme';
+import { disabledCursor } from '../../../components/utils/css';
+import { element, font } from '../../theme/reset';
+import { focus } from '../../theme/utils/focus';
+
+/* stylelint-disable no-descending-specificity */
```
ketsappi, line 7: Had to add this `no-descending-specificity` after all different ways to prevent the error. 
If anyone knows the correct way to fix this and current functionality stays same and styles are visually intact, I would like to learn it.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
```diff
@@ -0,0 +1,159 @@
...
+        color: ${theme.colors.depthDark1};
+        ${theme.typography.bodyTextSmall};
+    }
+
+    & .fi-radiobutton_input {
+      opacity: 0;
+      position: absolute;
+      margin: 8px 0 0 3px;
+
+      + label {
```
ketsappi, line 26: Not using component specific names in every place on these scss rules (e.g **_fi-radiobutton_label_**), but should we?

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
```diff
@@ -0,0 +1,159 @@
...
+      }
+
+      &:disabled:checked {
+        + label:after {
+          background: ${theme.colors.depthBase};
+          width: 8px;
+          height: 8px;
+          top: 10px;
+          left: 5px;
+        }
```
ketsappi, line 102: Both variants, `small` and `large` uses this kind of styling. When radiobutton is disabled **and** selected, it is slightly different than in designs. I could not find a way to make it look exactly same and not worth the effort.
But if this is a real issue, this needs more investigation.

src/components/Form/RadiobuttonGroup.tsx
```diff
@@ -0,0 +1,159 @@
...
+    children,
+    (child: React.ReactElement<RadiobuttonProps>) => {
+      if (React.isValidElement(child) && child.props.value) {
+        return React.cloneElement(child, {
+          radiobuttonGroup: true,
+          checked: selectedValue === child.props.value,
+          name: groupName,
+        });
+      }
+      return child;
```
ketsappi, line 74: If we would want to pass `variant` prop also to **RadiobuttonDivider** this should be the place. Problem might be to detect it reliably.

src/components/Form/RadiobuttonGroup.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+const RadiobuttonGroupItems = (
+  children: Array<React.ReactElement<RadiobuttonProps>>,
+  groupName: string,
+  selectedValue?: string,
+) =>
+  React.Children.map(
+    children,
+    (child: React.ReactElement<RadiobuttonProps>) => {
+      if (React.isValidElement(child) && child.props.value) {
```
ketsappi, line 67: `&& child.props.value` condition is here to detect that the child is a Radiobutton and **not** a RadiobuttonDivider.
If there is more elegant and reliable way to make it better, we should change this.

src/components/Form/Radiobutton.tsx
```diff
@@ -0,0 +1,156 @@
...
+const radiobuttonClassNames = {
+  container: `${baseClassName}_container`,
+  input: `${baseClassName}_input`,
+  label: `${baseClassName}_label`,
+  hintText: `${baseClassName}_hintText`,
+};
+
+type RadiobuttonVariant = 'small' | 'large';
+
+export interface RadiobuttonProps {
```
aappoalander, line 22: No need for common input props?

src/components/Form/Radiobutton.tsx
```diff
@@ -0,0 +1,156 @@
...
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+}
+
+export interface RadiobuttonState {
+  checkedState: boolean;
+}
+
+class RadiobuttonItem extends Component<RadiobuttonProps> {
+  state = {
+    checkedState: !!this.props.checked,
```
aappoalander, line 62: What happens if user adds shorthand like this: `<RadiobuttonItem checked ...`? Just wondering if this truthy check is sufficient or should it actually check if the property exists in props like we had to do for the value for dropdown?

src/components/Form/RadiobuttonGroup.tsx
aappoalander: Seems to be the only classname that actually point to the div containing radiogroup? In the actual Radiobutton.tsx it seems to be redundant since it's only used for the same element with baseClassName?

src/components/Form/RadiobuttonGroup.tsx
aappoalander: https://stackoverflow.com/a/53138782 could this work for detecting child types?

src/components/Form/RadiobuttonGroup.tsx
```diff
@@ -0,0 +1,159 @@
...
+    prevState: RadiobuttonGroupState,
+  ) {
+    const { value } = nextProps;
+    if (value !== undefined && value !== prevState.selectedValue) {
+      return { selectedValue: value };
+    }
+    return null;
+  }
+
+  handleRadiobuttonChange = (value: string) => {
```
aappoalander, line 98: I presume this does not work well with "controlled state"? Let's say I want to add a confirmation box before selecting an option, and user clicks cancel => state is changed internally and the external state is ignored since it hasn't changed?

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
aappoalander: I think we had similar issues with other components. This rule does not provide much value with css-in-js solution in my opinion.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
aappoalander: I'd assume it's more risky to do it this way, but since we are only supporting strings as children I guess there is no issue?

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+      + label {
+        position: relative;
+        display: inline-block;
+        cursor: pointer;
+        min-height: 27px;
+        line-height: 1.5em;
+        padding-left: 26px;
+
+        &::before {
```
aappoalander, line 34: Some comments for neat css drawing might help in the future. E.g. radio-input circle etc.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
```diff
@@ -0,0 +1,159 @@
...
+      &:focus {
+        + label::before {
+          ${focus({ theme, noPseudo: true, variant: 'boxShadow' })}
+          border-radius: 50%;
+        }
+      }
+
+      &:disabled {
+        + label {
+          ${disabledCursor}
```
aappoalander, line 78: Are cursors automatically handled for other elements / no need to add this elsewhere when disabled?

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
aappoalander: Maybe check with PAMU team as well?

src/core/Form/Radiobutton/__snapshots__/Radiobutton.test.tsx.snap
```diff
@@ -0,0 +1,2033 @@
...
+  left: 7px;
+}
+
+<div>
+  <div
+    class="c0 fi-radiobutton_container fi-radiobutton c1 fi-radiobutton--disabled"
+  >
+    <input
+      class="c2 fi-radiobutton_input"
+      disabled=""
```
aappoalander, line 739: Should this be omitted completely if not provided? Or is this case disabled={false} in React code?

src/core/Form/Radiobutton/__snapshots__/Radiobutton.test.tsx.snap
```diff
@@ -0,0 +1,2033 @@
...
+      id="test-id"
+      type="radio"
+      value=""
```
aappoalander, line 742: Same goes here, having empty string as value might have side-effects.

src/components/Form/Radiobutton.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+export interface RadiobuttonProps {
+  /** Custom classname to extend or customize */
+  className?: string;
+  /**
+   * Label element content
+   */
+  children?: ReactNode;
+  /** Group multiple Radiobuttons to be inside same selection group */
+  name?: string;
```
aappoalander, line 30: There is a small chance for error when RadiobuttonGroup name is defined and name is also provided for the button as these can be different. Or does the name on the group have any effect on things?

aappoalander:
Looks good overall! Great tests once again!

src/components/Form/Radiobutton.tsx
ketsappi: Not sure. I used  **Checkbox** as an example to look guidance and to keep it similar.

If trying to use `extends Omit<HtmlInputProps, 'type'>` on the Radiobutton it complains of multiple problems, as currently the `...passProps` are passed to **HtmlLabel** as with the Checkbox.

But I'm not sure if we want **Radiobutton** and **Checkbox** to give those `...passProps` to **HtmlLabel** or **HtmlInput**. Do you remember @LJKaski was there some special reason to pass them to HtmlLabel?

aappoalander: That is actually a valid point, we do have additional stuff in addition to that actual input. I still think that passprops should go to the input ja there might be quite a few common input props that we may need to support later on unless support is added now. Omitting type from common props should do the trick, but there might be too much stuff in the common props, so not sure what would be the best approach.

src/components/Form/RadiobuttonGroup.tsx
ketsappi: That way didn't seem to be possible. But the following does work, but seems little wonky as required to import from core side. (_more circulars_)

```jsx
import { Radiobutton as CoreRadiobutton } from '../../core/Form/Radiobutton/Radiobutton';
...
if (React.isValidElement(child) && child.type === CoreRadiobutton) {
```

Should we change it like that ^^ ?

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
ketsappi: I think it is sufficient and working as of tested with Brave-browser.

`label` basically is the only visible element there as the` input` element is hidden.
That background and "selected" _circles_  are done with `::after` and `::before` pseudos of the `label`.

src/core/Form/Radiobutton/__snapshots__/Radiobutton.test.tsx.snap
ketsappi: Yes, good finds. These should be changed like it was done in other PR with the **Checkbox**'s value:  https://github.com/vrk-kpa/suomifi-ui-components/pull/375/commits/66d14aa9711ba03a8f4cb80ee653676ea5895d16

src/core/Form/Radiobutton/__snapshots__/Radiobutton.test.tsx.snap
ketsappi: This seems to be case when `disabled={true}`, and it will shown as `disabled=""` on snapshot.

src/components/Form/Radiobutton.tsx
ketsappi: I tested it now on the styleguidist and the `name` prop that **RadiobuttonGroup** do give will be used, as it happens later.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
ketsappi: Added comments to explain all those pseudo codes of radio input circles in the recent commit.

src/core/Form/Radiobutton/__snapshots__/Radiobutton.test.tsx.snap
ketsappi: This change is now committed to repository.

src/components/Form/RadiobuttonGroup.tsx
ketsappi: > Seems to be the only classname that actually point to the div containing radiogroup?

True that. It is the only classname that has the children, **Radiobuttons** or/and **RadiobuttonDividers**
So it can be left as now.

ketsappi: > In the actual Radiobutton.tsx it seems to be redundant since it's only used for the same element with baseClassName?

Do you mean that we should remove the `baseClassName` from the **components** side and just add that in the **core** side?

src/components/Form/Radiobutton.tsx
LJKaski: I think that might just be an error in Checkbox as well. Passing to the input makes more sense for sure. In general I think it is easier to just pass the input props and omit the ones with the most obvious clashing potential.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
LJKaski: I see no big threat either, but somehow would still feel safer with using specific classnames. Would be clearer for developers as well.

src/components/Form/Radiobutton.tsx
```diff
@@ -0,0 +1,156 @@
...
+  /** Unique value to be used inside Radiobutton group for this item */
+  value?: string;
```
LJKaski, line 32: Why is value optional here?

aappoalander: That's a good question! Not sure if there is a use case without value, but current implementation does not work if no value is provided. At least inside a radiogroup nothing is selected if no value is given.

src/components/Form/Radiobutton.tsx
aappoalander: Might be worth documenting this to the radioGroup/radiobutton. E.g. overriden by group name

src/components/Form/RadiobuttonGroup.tsx
aappoalander: Nope, I'd remove the container here since the element with baseClass name is the container.

src/components/Form/Radiobutton.tsx
aappoalander: onChange is not documented, but yet it is used in the examples. I'd either document the prop or remove it from the example.

aappoalander: Btw. Should there be an onChange event for the group? And about the type, is the whole event necessary here, or could we just pass the new value if it really changes?

src/components/Form/RadiobuttonGroup.tsx
LJKaski: I'd put it as required then. I see no use cases where you want a radiobutton without value, especially if it is required for it to function correctly.

ketsappi: I noticed that my original comment for this belongs actually to **Radiobutton** instead of **RadiobuttonGroup**.
Here the value should be optional as it acts like controlling the selected radio if using this.

src/components/Form/RadiobuttonGroup.tsx
ketsappi: So you want to remove the `fi-radiobuttongroup_container` then? Because if removed, then there would be no specific classname provided to catch the container for radios.
Or would it better to rename it to reflect it's use better, like `fi-radiobuttongroup_radios`?

Because the **baseClassName**,  `fi-radiobuttongroup`,  is only provided to root of the component.

src/components/Form/RadiobuttonGroup.tsx
LJKaski: Ah, meant my comment for the radiobutton as well.

src/components/Form/Radiobutton.tsx
ketsappi: Currently there is no `onChange` prop for group. But it is good idea to be added.

One could possibly use the existing https://github.com/vrk-kpa/suomifi-ui-components/blob/5a9f7a28d5748b56a3f92007fdb12fceb3ca7949/src/components/Form/RadiobuttonGroup.tsx#L47-L49

with the implementation. It is used to handle the case when one of the radios are selected in the
https://github.com/vrk-kpa/suomifi-ui-components/blob/5a9f7a28d5748b56a3f92007fdb12fceb3ca7949/src/components/Form/RadiobuttonGroup.tsx#L98-L100

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
ketsappi: These all are now changed to be explicitly named, so the example shown above is now

https://github.com/vrk-kpa/suomifi-ui-components/blob/f2c4662246ea8313f769cbafa2d9c9006732874f/src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx#L23-L26

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
ketsappi: Asked from the PAMU and to keep it consistent with e.g TextInput, they slight grey color should be with disabled and checked as in the design (background is now white when it is selected and the component is disabled)

src/components/Form/Radiobutton.tsx
aappoalander: I think it would make using these easier since only one listener would be sufficient.

aappoalander: Unless it messes the controlled state somehow when used mixed with the ones for radiobutton 😓

src/components/Form/RadiobuttonGroup.tsx
aappoalander: Ah, sorry for misunderstanding. For the buttonGroup current implementation seems to be solid, but for the components/Form/Radiobutton.tsx there is also a container class which in my opinion is redundant.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
aappoalander: So this should now be exactly as specified in designs?

src/components/Form/RadiobuttonGroup.tsx
aappoalander: This might become an issue later on, but I guess it's working for now. I'll leave it up to you to decide if we should leave it as is or pursue some other alternative.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
ketsappi: No, because currently it is white/transparent. Haven't yet made change to this part.

src/components/Form/Radiobutton.tsx
ketsappi: Value is now required prop.

src/core/Form/Radiobutton/Radiobutton.baseStyles.tsx
LJKaski: As discussed yesterday, this should be achieved by simply adding `background-color: ${theme.colors.depthLight3};` to the disabled `:before` style.

ketsappi: This is now changed, just with a little twist: using `background` to set the color to keep it consistent.

src/components/Form/Radiobutton.tsx
aappoalander: After digging through HtmlInputProps which is essentially HtmlInputElement from lib.dom.d.ts, it looks like there are way too many native attributes causing or enabling unwanted effects e.g. with styling. Considering this it was agreed to not add common input props.

src/components/Form/RadiobuttonGroup.tsx
aappoalander: Seems to be ok now. There hierarchy is as follows: fi-radiobuttongroup_container > fi-radiobutton_container > fi-radiobutton-input & fi-radiobutton_label and so on.

src/components/Form/RadiobuttonGroup.tsx
aappoalander: Currently there is no reliable way for checking the typing of child elements. Text and components etc. can be checked, but that does not provide additional value for now.

aappoalander: Left a comment to code to clarify this.

src/components/Form/RadiobuttonGroup.tsx
aappoalander: After testing it seems that things work ok, since Radiobuttons checked and name props are overridden / dismissed when nested in a group.

src/components/Form/Radiobutton.tsx
aappoalander: Documentation updated.

aappoalander: And onChange prop for the group added along with existence check for value in group props to prevent changing state accidentally in controlled state.

src/components/Form/Radiobutton.tsx
aappoalander: Seems to work as expected.

PR #378, opened 2020-07-31
[Feature] TextInput improvements

ketsappi:
## Description

- `hintText` prop for the TextInput
- Able to give `type` to TextInput: ```'text' | 'email' | 'number' | 'password' | 'tel' | 'url'```
- More tests added.

## Motivation and Context

These changes were made because those were wanted by the designs and users of the library. 

Now user is able to give some different **type** than the previous `text` if it is needed. Not every type is supported as it makes no sense to give e.g _checkbox_ or _button_ as types because we have separate component for those.

Tests were added to increase test coverage and to hopefully help in future to catch some possible bugs.

## Screenshots
image

## Release notes

Feature
- TextInput now has `hintText` and `type` props.

### DIFF
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -29,6 +29,8 @@ export interface TextInputLabelProps extends HtmlLabelProps {}
 
 type Label = 'hidden' | 'visible';
 
+type InputType = 'text' | 'email' | 'number' | 'password' | 'tel' | 'url';
+
 export interface TextInputProps extends Omit<HtmlInputProps, 'type'> {
   /** Custom classname for the input to extend or customize */
   className?: string;
@@ -61,6 +63,10 @@ export interface TextInputProps extends Omit<HtmlInputProps, 'type'> {
   hintText?: string;
   /** Status text to be shown below the component and hint text. Use e.g. for validation error */
   statusText?: string;
+  /** 'text' | 'email' | 'number' | 'password' | 'tel' | 'url'
+   * @default text
+   */
+  type?: InputType;
 }
 
 class BaseTextInput extends Component<TextInputProps> {
@@ -78,6 +84,7 @@ class BaseTextInput extends Component<TextInputProps> {
       hintText,
       visualPlaceholder,
       id: propId,
+      type = 'text',
       ...passProps
     } = this.props;
 
@@ -111,7 +118,7 @@ class BaseTextInput extends Component<TextInputProps> {
               id={propId}
               {...passProps}
               className={classnames(inputBaseClassName, inputClassName)}
-              type="text"
+              type={type}
               aria-describedby={[
                 generatedStatusTextId,
                 generatedHintTextId,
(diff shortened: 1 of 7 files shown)


src/components/Form/TextInput.tsx
```diff
@@ -92,20 +103,31 @@ class BaseTextInput extends Component<TextInputProps> {
...
               id={propId}
               {...passProps}
               className={classnames(inputBaseClassName, inputClassName)}
-              type="text"
-              aria-describedby={generatedId}
+              type={type}
+              aria-describedby={[
+                generatedStatusTextId,
+                generatedHintTextId,
+              ].join(' ')}
```
aappoalander, line 121: I'd keep this logic outside of the template. The classnames util might actually work for this as well, but wouldn't still use it for this use case.

ketsappi: Logic is now moved from the template.
_Also fixes that non existing ids are given to aria._

src/core/Form/TextInput/TextInput.md
```diff
@@ -60,3 +65,20 @@ const status = errorState ? 'error' : 'default';
...
 ```
+
+```js
+import { TextInput } from 'suomifi-ui-components';
+
+<>
+  <TextInput
+    labelText="TextInput with numbers"
+    type="number"
+    value={123}
```
LJKaski, line 76: I'd use defaultValue for these to allow playing around with the examples. With a set value editing is not possible.

ketsappi: Example is now using `defaultValue` instead of the previous `value`

src/core/Form/SearchInput/__snapshots__/SearchInput.test.tsx.snap
```diff
@@ -276,7 +295,7 @@ exports[`calling render with the same component on the same container does not r
       class="c4 fi-search-input_input-container fi-text-input_container"
     >
       <input
-        aria-describedby="test-id-statusText"
+        aria-describedby=""
```
aappoalander, line 298: is this due to test setup or something that occurs with real use cases? Could empty aria-describedby have undesired side-effects?

ketsappi: Good find.

If I read correctly, previously the generated id for `statusText` was always added to `aria-describedby` even if the were no `statusText` given. Now, after changes the generated id's (_statusText and the new hintText_) are only added when they are given.

Empty `aria-describedby` might cause small pause when screenreader reads it according to this discussion:
https://developer.paciellogroup.com/blog/2018/09/describing-aria-describedby/#div-comment-4779

LJKaski: I came across that discussion earlier too. How I understood it was that the delay was caused by the element referred to by the `aria-describedby` being empty. Still always better to have a clean solution :+1:

aappoalander: Seems to be clean now, no attribute at all if no related descriptions are given.

PR #386, opened 2020-08-25
[Feature] Add name prop to form component APIs

LJKaski:
## Description
This PR adds name property to form components that didn't previously have them.
ToggleInput is not really a form component, but is a checkbox under the hood, so name prop was added for that as well.

Also added documentation for the children prop in TextInput.

## Motivation and Context
Form elements should support different ways of using them, and those ways should be documented. With the addition of documentation for the name prop the support for a classic way of building a form is effectively conveyed.

## How Has This Been Tested?
Running locally & automated tests.

## Release notes
Added name property to form component APIs.

### DIFF
--- a/src/core/Form/TextInput/TextInput.test.tsx
+++ b/src/core/Form/TextInput/TextInput.test.tsx
@@ -99,8 +99,7 @@ describe('props', () => {
       const namedInput = getByTestId('input-name') as HTMLInputElement;
 
       it('has the given name attribute', () => {
-        fireEvent.change(namedInput, { target: { name: 'newName' } });
-        expect(namedInput.name).toBe('newName');
+        expect(namedInput.name).toBe('test-name');
       });
     });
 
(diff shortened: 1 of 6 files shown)


aappoalander:
I'd like to see some tests for these props to ensure they are passed on to the actual inputs.

LJKaski:
> I'd like to see some tests for these props to ensure they are passed on to the actual inputs.

Good point. Checked that on code level and tested manually but could also add automated tests.

src/core/Form/TextInput/TextInput.test.tsx
```diff
@@ -87,6 +87,23 @@ describe('props', () => {
...
+          labelText="Test input"
+          name="test-name"
+          data-testid="input-name"
+        />
+      );
+      const { getByTestId } = render(textInput);
+      const namedInput = getByTestId('input-name') as HTMLInputElement;
+
+      it('has the given name attribute', () => {
+        fireEvent.change(namedInput, { target: { name: 'newName' } });
```
aappoalander, line 102: Why test for name change? Should at least have an expect for the originally given name prop.

aappoalander (approved):

PR #388, opened 2020-09-07
[Feature] Aria-invalid attributes

aappoalander:
## Description
This PR adds aria-invalid attributes to inputs that already have the status prop. Aria-invalid is set to true when status equals "error".

Additionally, redundant aria-describedby attributes have been removed TextInput and those are now only read once by screenreaders.

## Motivation and Context
Cases where there are errors are now natively announced by screen readers. This is a better practice than only relying on error messages.

## How Has This Been Tested?
Tested with Mac OS Safari, Chrome and Firefox browsers using VoiceOver and Windows Chrome, Firefox and Edge browsers using NVDA.

## Release notes
- Form components with status now also use aria-invalid to communicate errors.
- TextInput labels and texts are not read twice any more.

### DIFF
--- a/src/components/Form/TextInput.tsx
+++ b/src/components/Form/TextInput.tsx
@@ -30,6 +30,8 @@ type Label = 'hidden' | 'visible';
 
 type InputType = 'text' | 'email' | 'number' | 'password' | 'tel' | 'url';
 
+type TextInputStatus = 'default' | 'error' | 'success';
+
 export interface TextInputProps extends Omit<HtmlInputProps, 'type'> {
   /** Custom classname for the input to extend or customize */
   className?: string;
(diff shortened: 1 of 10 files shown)


src/components/Form/TextInput.tsx
```diff
@@ -130,8 +126,8 @@ class BaseTextInput extends Component<TextInputProps> {
               {...passProps}
               className={classnames(inputBaseClassName, inputClassName)}
               type={type}
-              {...getDescribedBy()}
```
LJKaski, line 133: this change works as it should, though the jury is still out on which approach is more appropriate.

LJKaski (approved):
Changes work fine. The aria-describedby vs. 'wrap everything inside the label' approaches still need to be assessed, but that's a different task.

src/components/Form/TextInput.tsx
aappoalander: I'd argue that there might not be one best approach for this that fits all cases. Some things can be wrapped inside a label while other can't or may not work as desired. Might be better to use similar approach for everything since screen readers and browsers do not the same way => many different approaches with currently similar end result are not guaranteed to work the same way in the future.

LJKaski: Luckily these are things we can adjust later if need be without breaking the API. I like the 'inside the label' approach since the read order is the most logical in that one at the moment.

PR #390, opened 2020-09-16
[Feature] Chip component

LJKaski:
## Description
This PR adds a Chip component to the library. Chip can be used as a button or simply as a styled way to present information. It is going to be used at least inside the upcoming Combobox component.

There are two variants of the chip. One is semantically a button, with an onClick prop and styles to present hover and active states. The other one is simply a styled span with no interactivity.

## Motivation and Context
Chip is a component needed for some features in the Suomi.fi services. It can also be used as a part of other components and modules such as Combobox.

## How Has This Been Tested?
So far tested by running locally and testing manually with firefox, chrome and edge.

image

## Release notes
Added Chip component


src/core/Chip/Chip.baseStyles.tsx
```diff
@@ -0,0 +1,79 @@
+import { css } from 'styled-components';
+import { withSuomifiTheme, TokensAndTheme, SuomifiThemeProp } from '../theme';
+import { element, font, focus } from '../theme/reset';
+import { disabledCursor } from '../../components/utils/css';
+
+const removableStyles = ({ theme }: SuomifiThemeProp) => css`
+  &.fi-chip--removable {
```
aappoalander, line 7: Are these really only for removable chips?

src/core/Chip/Chip.baseStyles.tsx
```diff
@@ -0,0 +1,79 @@
...
+      border-radius: 16px;
+      }
+    }
+
+    &.fi-chip {
+      border-radius: 14px;
+      padding: ${theme.spacing.insetXxs} ${theme.spacing.insetL};
+      color: ${theme.colors.whiteBase};
+      background: ${theme.colors.highlightBase};
+      font: ${theme.typography.actionElementInnerText};
```
aappoalander, line 43: Is the bodyText above for something else or could we just use this one with the font function? And should the font function be used here as well? I think it applies e.g. word-breaks in addition to text design token.

src/core/Chip/Chip.baseStyles.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+      &:active {
+        background: ${theme.colors.highlightDark1};
+        }
+      }
+
+    & .fi-chip--icon{
+      height: 12px;
+      width: 12px;
+      transform: translateY(-0.35em);
```
aappoalander, line 62: Does this work well with browser zoom?

LJKaski: Tried it with the regular zoom function in Firefox and Chrome and it worked fine on both. Since we use pixel values for the fonts and elements, the text size setting doesn't really make a difference either way.

src/core/Chip/Chip.md
```diff
@@ -0,0 +1,30 @@
...
+    </Chip>
+  </div>
+  <div>
+    <Chip onClick={removeAction}>
+      Clickable chip without remove icon
+    </Chip>
+  </div>
+
+  <Chip.static>
+    Static chip with a long content that exceeds the size limit
```
aappoalander, line 25: This is a good example. Maybe even add the default max width to label? "exceeds the default maximum width of xx px." Since size limit is a bit ambiguous and could refer to character limit as well.

src/core/Chip/Chip.baseStyles.tsx
LJKaski: Yes. When the icon is shown, the max width for the content has to be lower for the overflow styles and positioning to work correctly.

src/core/Chip/Chip.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { default as styled } from 'styled-components';
+import { baseStyles } from './Chip.baseStyles';
+import { HtmlButton, HtmlSpan } from '../../reset';
+import { TokensProp, InternalTokensProp } from 'core/theme';
+import { withSuomifiDefaultProps } from '../theme/utils';
+import { Icon } from '../Icon/Icon';
+import { logger } from '../../utils/logger';
+import { VisuallyHidden } from '../../components/Visually-hidden/Visually-hidden';
+
+const baseClassName = 'fi-chip';
```
aappoalander, line 12: We had this classnames object somewhere where you could just use e.g. classnames.disabled in the code. I personally preferred that over this more traditional solution.

src/core/Chip/Chip.tsx
```diff
@@ -0,0 +1,126 @@
...
+const disabledClassName = `${baseClassName}--disabled`;
+const iconClassName = `${baseClassName}--icon`;
+const contentClassName = `${baseClassName}--content`;
+const removableClassName = `${baseClassName}--removable`;
+const buttonClassName = `${baseClassName}--button`;
+
+type ChipVariant = 'static' | 'default';
+
+interface InternalChipProps {
+  /** Chip element content */
```
aappoalander, line 22: should we add styles to allow overriding e.g. max-width without using the attached classnames? Not sure if we should, just wondering.

src/core/Chip/Chip.md
```diff
@@ -0,0 +1,30 @@
+```js
+import { Chip } from 'suomifi-ui-components';
+
+const removeAction = () => {
+  alert('Selection removed');
+};
+
+<>
+  <div styles={{ display: 'inline' }}>
```
aappoalander, line 9: This ends up in DOM as styles=[object, object]

LJKaski: Ah, it doesn't seem to be necessary after all. Just the div is enough 👍

src/core/Chip/Chip.tsx
LJKaski: I think that this is a simple enough component that if you really want a chip that wide you can quite easily build your own or do some extra tricks with this one. I don't think it's worthwhile to support that out of the box.

src/core/Chip/Chip.tsx
LJKaski: I think I used that in Checkbox. Didn't remember which one we agreed on. I'll implement it here as well.

src/core/Chip/Chip.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  it('should match snapshot', () => {
+    const { container } = render(customClassChip);
+    expect(container).toMatchSnapshot();
+  });
+});
+
+describe('variants', () => {
+  const defaultChip = (
+    <Chip removable={true} onClick={() => null} actionLabel="Unselect">
```
aappoalander, line 72: Unselect sounds weird, should be deselect or remove?

src/core/Chip/Chip.tsx
```diff
@@ -0,0 +1,126 @@
...
+    if (variant === 'static') {
+      return (
+        <HtmlSpan
+          className={classnames(baseClassName, className, {
+            [disabledClassName]: !!disabled,
+            [removableClassName]: !!removable,
+          })}
+          {...passProps}
+        >
+          <HtmlSpan className={contentClassName}>{children}</HtmlSpan>
```
aappoalander, line 76: Is this additional span necessary?

src/core/Chip/Chip.tsx
```diff
@@ -0,0 +1,126 @@
...
+      <HtmlButton
+        className={classnames(baseClassName, buttonClassName, className, {
+          [disabledClassName]: !!disabled,
+          [removableClassName]: !!removable,
+        })}
+        disabled={disabled}
+        onClick={onClick}
+        {...passProps}
+      >
+        <HtmlSpan className={contentClassName}>{children}</HtmlSpan>
```
aappoalander, line 90: Same for this one, does the behaviour change if this span is removed and styles are applied directly to the parent element? This way content classname would not be needed either.

src/core/Chip/Chip.md
```diff
@@ -0,0 +1,30 @@
...
+    <Chip removable actionLabel="Unselect" onClick={removeAction}>
```
aappoalander, line 10: Unselect sounds weird, should be deselect or remove?

src/core/Chip/Chip.tsx
LJKaski: I think it does. We need to be able to limit the size of the text content element in order to handle the overflow of the content correctly when an icon is present.

src/core/Chip/Chip.tsx
LJKaski: This one might actually be redundant. I'll look into it.

aappoalander:
Tests look good! Good work with those.

src/core/Chip/Chip.tsx
LJKaski: We could maybe position the icon absolutely and handle this one without the span, but I'm not sure if that's a cleaner solution.

aappoalander: Sure, makes sense!

src/core/Chip/Chip.md
LJKaski: I think deselect would be better. I'll change it.

src/core/Chip/Chip.md
LJKaski: Then I need to rephrase the content, since the max width of the content depends on the variant while the max width of the component stays constant (290px). I'd be quite happy keeping this as it is.

src/core/Chip/Chip.baseStyles.tsx
LJKaski: Just using actionElementInnerText in the main level font function seems to be enough, since there are no external labels or other things to consider here. Good point.

src/core/Chip/Chip.tsx
LJKaski: Removing this would require some duplicate code and restructuring in the styles, so this might be a cleaner way to control the size and overflow.

src/core/Chip/Chip.md
aappoalander: I would rephrase so that it says that max-width of the component is 290px. It's relevant information and helps devs when documented. Maybe something like "Static chip with long content that does not fit into components max-width of 290px."

src/core/Chip/Chip.tsx
LJKaski: now changed.

src/core/Chip/Chip.baseStyles.tsx
LJKaski: This is now fixed.

src/core/Chip/Chip.md
LJKaski: Good suggestion. Put it into use.

src/core/Chip/Chip.md
LJKaski: Should now be fixed.

src/core/Chip/Chip.md
LJKaski: Now changed.

src/core/Chip/Chip.tsx
aappoalander: Sounds reasonable, so let's keep it as it is even if not immediately obvious why it's there.

aappoalander:
What about side margins? Looks kind of bad when chips are placed next to each other and there is no whitespace between them. Especially with the focus ring.

aappoalander (approved):

aappoalander:
> What about side margins? Looks kind of bad when chips are placed next to each other and there is no whitespace between them. Especially with the focus ring.

As discussed in other forum, an example with margins is sufficient at this time and providing global margins would not be in line with other components.

PR #392, opened 2020-09-22
[Feature] Popover component

aappoalander:
## Description

This PR adds a low level React Portal Popover component for use with other components in the future. Especially with upcoming combobox. The implementation is built on top of popper.js / React Popper for the parts that handle the popover container placement. This component is internal for the library and there is no plan for making it available in the public API.

## Motivation and Context

This solution is low level when compared to existing Reach UI implementations and leave more room for desired custom functionality and should reduce undesired breaking changes. This should also enable getting rid of Reach UI dependency completely in the future.

## How Has This Been Tested?

The implementation was tested by creating a test component with a button which opened a popup underneath and moved focus to other buttons in the popup. The test component was removed before introduction of this PR.

Screenshot 2020-09-21 at 14 16 11
Screenshot 2020-09-21 at 14 16 21

## Release notes

- Popover component


LJKaski (approved):
Seems to work well and support what is needed at this point.

PR #393, opened 2020-09-23
[Feature] Text Input and Search Input style improvements

aappoalander:
## Description
- Text and Search Inputs have a default width of 290px. This can be overridden when necessary.
- Text and Search Inputs have fullWidth prop which allows the inputs to reserve all available space.
- Search Input has status prop to support error status styling and aria-invalid.
- Search Input error styles and aria-invalid are now working

## Breaking Changes
- inputClassName is removed from the API
- inputContainerProps are now applied to the outermost div element (vs. input element wrapper div).
- inputContainerProps type now omits className from HtmlDivProps. className prop can be used instead.
 
## Motivation and Context
The default width was set to 100% which made the input always take all available space. This was not the default use case and was therefore changed. Styling of the component was difficult due to label-element being the outermost element. Font-size was bleeding to the component from outside which broke the styles. Search Input didn't handle error states properly. 

## How Has This Been Tested?
Tested with Styleguidist build, SSR test and create react app. Style overrides tested with css, css-in-js and style attribute.

## Release notes
- Default width of 290px and fullWidth prop added for TextInput and SearchInput.
- Status prop added for Search Input.
- Multiple style fixes for Text and Search Input.
- Breaking changes to TextInput and SearchInput API and functionality. inputClassName removed, inputContainerProps type omits className, inputContainerProps are applied to top most div. See #393 for further details.

### DIFF
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -35,7 +35,7 @@ export const baseStyles = withSuomifiTheme(
   & .fi-text-input_input {
     ${input({ theme })}
     background-color: ${theme.colors.whiteBase};
-    width: 100%;
+    width: 290px;
     min-height:40px;
     padding-left: ${theme.spacing.insetL};
     ::placeholder{
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -28,7 +28,7 @@ export const baseStyles = withSuomifiTheme(
     color: ${theme.colors.blackBase};
   }
 
-  & .fi-text-input_container {
+  & .fi-text-input_input-element-container {
     ${inputContainer({ theme })}
   }
 
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -35,10 +35,10 @@ export const baseStyles = withSuomifiTheme(
   & .fi-text-input_statusText_container {
     display: flex;
     flex-direction: column;
+    font-size: 14px;
 
     & .fi-text-input_statusText {
       ${theme.typography.bodySemiBoldSmall}
-      font-size: 14px;
       line-height: 20px;
     }
   }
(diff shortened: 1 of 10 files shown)


src/core/Form/TextInput/TextInput.md
```diff
@@ -2,6 +2,7 @@
 import { TextInput } from 'suomifi-ui-components';
 <>
   <TextInput
+    className="testClassName"
```
LJKaski, line 5: does this serve a purpose or is it a relic from manual testing?

src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -1,40 +1,59 @@
...
 
-  & .fi-text-input_container {
+  & .fi-text-input_input-element-container {
     ${inputContainer({ theme })}
   }
 
   & .fi-text-input_statusText_container {
     display: flex;
     flex-direction: column;
+    font-size: 14px;
```
LJKaski, line 38: Moving this here seems to result in hint text having wrong text size. It gets 16px as the size from somewhere.

src/core/Form/TextInput/TextInput.md
```diff
@@ -66,6 +67,23 @@ const status = errorState ? 'error' : 'default';
...
+<>
+  <TextInput
+    labelText="Test TextInput with fixed custom width of 250px"
+    status={status}
+    inputContainerProps={{ style: { width: '250px' } }}
+  />
+
+  <TextInput
+    labelText="Test TextInput with 100% width"
+    status={status}
```
LJKaski, line 81: This seems unnecessary, since the example doesn't use status.

src/core/Form/TextInput/TextInput.baseStyles.tsx
aappoalander: Good catch! Fixed!

src/core/Form/TextInput/TextInput.md
aappoalander: True, was left there by accident. Should be now removed.

src/core/Form/TextInput/TextInput.md
aappoalander: It was there to demonstrate custom styling, but decided to remove it since it has no effect here.

LJKaski (approved):
Seems to be in order. Good work! :)

PR #394, opened 2020-09-25
[Feature] Text input icon support

LJKaski:
## Description
This PR implements support for using icons inside the text input component. User can give the icon name as well as any preferred icon props to the component. Those get passed to the corresponding elements. The necessary styling is also applied when an icon is given.

## Motivation and Context
Text input with an icon is part of the designs for this component, but it hasn't been implemented until now.

## How Has This Been Tested?
So far only tested by running locally.

## Release notes
Added support for icons inside text inputs.

### DIFF
--- a/src/core/Form/SearchInput/__snapshots__/SearchInput.test.tsx.snap
+++ b/src/core/Form/SearchInput/__snapshots__/SearchInput.test.tsx.snap
@@ -244,6 +244,22 @@ exports[`calling render with the same component on the same container does not r
   box-shadow: 0 1px 2px 0 rgba(0,53,122,0.1) inset;
 }
 
+.c2.fi-text-input_with-icon .fi-text-input_container {
+  position: relative;
+}
+
+.c2.fi-text-input_with-icon .fi-text-input_input {
+  padding-right: 40px;
+}
+
+.c2.fi-text-input_with-icon .fi-icon {
+  position: absolute;
+  width: 18px;
+  height: 18px;
+  top: 10px;
+  right: 10px;
+}
+
 .c2.fi-text-input--error .fi-text-input_input {
   border-color: hsl(3,59%,48%);
 }
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -1,6 +1,7 @@
 import { css } from 'styled-components';
 import { withSuomifiTheme, TokensAndTheme } from '../../theme';
 import { input, inputContainer, font } from '../../theme/reset';
+import { math } from 'polished';
 
 export const baseStyles = withSuomifiTheme(
   ({ theme }: TokensAndTheme) => css`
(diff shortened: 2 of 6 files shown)


LJKaski:
Just realized I should probably add a test for this as well. I'll do it asap.

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -64,8 +71,31 @@ const StyledTextInput = styled(
...
+      className,
+      ...passProps
+    } = withSuomifiDefaultProps(this.props);
+
+    const prioritizedIcon = icon || iconProps?.icon;
+
+    const newIconProps = {
+      ...iconProps,
+      icon: prioritizedIcon,
+    };
```
LJKaski, line 87: This part is not the cleanest, but without destructuring icon and iconProps out of the props an empty iconProps object was passed to the component.

src/core/Form/SearchInput/__snapshots__/SearchInput.test.tsx.snap
```diff
@@ -244,6 +244,22 @@ exports[`calling render with the same component on the same container does not r
   box-shadow: 0 1px 2px 0 rgba(0,53,122,0.1) inset;
 }
 
+.c2.fi-text-input_with-icon .fi-text-input_container {
```
aappoalander, line 247: If it has no effect for other use, I'd prefer to set the container always `position: relative`

src/core/Form/SearchInput/__snapshots__/SearchInput.test.tsx.snap
```diff
@@ -244,6 +244,22 @@ exports[`calling render with the same component on the same container does not r
...
+  position: relative;
+}
+
+.c2.fi-text-input_with-icon .fi-text-input_input {
+  padding-right: 40px;
```
aappoalander, line 252: No token available / not suitable for this use?

src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -44,6 +45,26 @@ export const baseStyles = withSuomifiTheme(
...
     }
+    
+  &.fi-text-input_with-icon {
+    & .fi-text-input_container {
+    position: relative;
+    }
+
+    & .fi-text-input_input{
+    padding-right: ${math(
+      `${theme.spacing.insetXl} * 2 + ${theme.spacing.insetM}`,
```
aappoalander, line 56: We should not use spacing tokens and inset tokens together, internal spacings inside components should use inset-tokens. This has probably been done elsewhere already, but I'd still avoid it.

src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -44,6 +45,26 @@ export const baseStyles = withSuomifiTheme(
...
+    )};
+    }
+
+    & .fi-icon{
+        position: absolute;
+        width: 18px;
+        height: 18px;
+        top: ${theme.spacing.insetL};
+        right: ${theme.spacing.insetL};
```
aappoalander, line 65: Same as above. In my opinion, we should stick with inset tokens when defining stuff inside the components.

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -64,8 +71,31 @@ const StyledTextInput = styled(
...
-    const { ...passProps } = withSuomifiDefaultProps(this.props);
+    const {
+      children,
+      icon,
+      iconProps,
+      className,
+      ...passProps
+    } = withSuomifiDefaultProps(this.props);
+
+    const prioritizedIcon = icon || iconProps?.icon;
```
aappoalander, line 82: prioritizedIcon sound misleading to me. If not possible to use icon, I'd prefer resolvedIcon or something like that.

src/core/Form/TextInput/TextInput.tsx
aappoalander: I think this is completely fine, just don't like the name of the icon that gets resolved from props and IconProps.

src/core/Form/TextInput/TextInput.tsx
LJKaski: Your suggestion sounds like a better name. I'll go with that.

src/core/Form/TextInput/TextInput.tsx
LJKaski: Changed the name to resolvedIcon.

aappoalander (approved):

PR #399, opened 2020-10-09
[Feature] Create separate styling for selected items in dropdown

LJKaski:
## Description
This PR adds a styling for the currently selected item in dropdown.

## Motivation and Context
Previously the styling was tied to the hover styles, which meant that hovering over an option in the dropdown made the selected style disappear. The styling also didn't match the current designs. This PR fixes both of those issues.

## How Has This Been Tested?
Tested by running locally in styleguidist.

image

## Release notes
-Improve styling of selected element in dropdown

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -25,7 +25,7 @@ export const dropdownClassNames = {
   button: `${baseClassName}_button`,
   popover: `${baseClassName}_popover`,
   item: `${baseClassName}_item`,
-  actionMenuItem: `${baseClassName}_actionMenuItem`,
+  noSelectedStyles: `${baseClassName}_noSelectedStyles`,
 };
 
 export interface DropdownLabelProps extends HtmlLabelProps {}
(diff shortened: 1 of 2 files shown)


src/core/Dropdown/Dropdown.baseStyles.tsx
```diff
@@ -45,51 +45,62 @@ export const baseStyles = withSuomifiTheme(
 
 export const listboxPopoverStyles = withSuomifiTheme(
   ({ theme }: TokensAndTheme) => css`
```
aappoalander, line 47: What if you would destructure alwaysShowVisualPlaceholder here and use it to conditionally place styles on active items?

aappoalander: Then there would be no need for additional template logic or classes.

LJKaski: Could do it that way as well I guess. Is there a benefit compared to the class approach? The class approach is used in most places in the library.

aappoalander: I guess the end result is the same no matter how it's done, but after discussion I would also stick to this more pure class based approach.

src/components/Dropdown/Dropdown.tsx
```diff
@@ -25,6 +25,7 @@ export const dropdownClassNames = {
   button: `${baseClassName}_button`,
   popover: `${baseClassName}_popover`,
   item: `${baseClassName}_item`,
+  actionMenuItem: `${baseClassName}_actionMenuItem`,
```
aappoalander, line 28: I'd rather call this hideActiveSelection etc. so that it's not too strongly connected to action menu use case. The functionality itself looks good!

LJKaski: Good point. Changed the name.

aappoalander (approved):

PR #400, opened 2020-10-13
[Feature] Search input clear and submit

aappoalander:
## Description
This PR aligns SearchInput implementation with current designs and add a clear and a submit button to the component. In addition this PR introduces a major refactor with breaking changes for the component making it only available with Suomi.fi styles and disconnected from TextInput.

In addition, this PR adds new internal components that should be taken into use for other components as well. LabelText component is for input labels above input elements and StatusText is for input status descriptions below the inputs.

## Motivation and Context

Designs had a search button and a clear button for SearchInput. In addition, SearchInput was heavily dependent on TextInput implementation which made maintenance difficult.

## How Has This Been Tested?
Tested with Styleguidist, CRA TS, SSR app and with Windows and Mac browsers / screen readers.

While testing a couple of bugs were found. First one causing inputs with content to read previous element's texts when focused. This bug only concerns Chromium based browsers with VoiceOver and has been reported to Apple. The second one causing following elements to be read after typing to input and clearing the contents using tab+enter. Only concerns Firefox and VoiceOver and has been reported to Mozilla.

## Release notes
- Breaking Change: Multiple API and functionality changes for SearchInput (see #400 ).


src/core/Form/SearchInput/SearchInput.baseStyles.tsx
```diff
@@ -1,44 +1,158 @@
...
-  }: TokensAndTheme & Omit<TextInputProps, 'labelText' | 'status'>) => css`
+  }: TokensAndTheme &
+    Omit<
+      SearchInputProps,
+      'labelText' | 'clearButtonLabel' | 'searchButtonLabel'
+    >) => css`
     &.fi-search-input {
       ${font({ theme })('bodyText')}
       display: inline-block;
-      width: ${fullWidth
```
aappoalander, line 16: Tämä muutos pitäisi tehdä myös TextInputiin.

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -1,47 +1,265 @@
...
+      conditionalSetState('');
+      setTimeout(() => {
+        if (this.inputRef.current) {
+          this.inputRef.current.focus();
+        }
+      }, 100);
+    };
+
+    const onKeyPress = (event: React.KeyboardEvent<HTMLInputElement>) => {
+      if (!!this.state.value && event?.key === 'Enter') {
```
aappoalander, line 164: Should we allow searching with empty strings?

aappoalander: Actually, we can't since the design pattern does not allow it either.

LJKaski: That's something I'd leave to whoever writes the search function anyway.

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -1,47 +1,265 @@
...
+  searchButton: `${baseClassName}_button-search`,
+  searchIcon: `${baseClassName}_button-search-icon`,
+  clearButton: `${baseClassName}_button-clear`,
+  clearIcon: `${baseClassName}_button-clear-icon`,
+};
 
-export interface SearchInputProps extends Omit<TextInputProps, 'status'> {
-  status?: 'default' | 'error';
+interface SearchInputState {
+  value: string | number | string[] | undefined;
```
ketsappi, line 85: Could the `StateValue` type  used here? 
It seems to have the same possible values listed and name indicates that it is the same thing.

src/core/Form/StatusText/StatusText.tsx
```diff
@@ -0,0 +1,62 @@
...
+      })}
+    >
+      {children}
+    </HtmlSpan>
+  ),
+)`
+  ${(tokens) => baseStyles(withSuomifiDefaultProps(tokens))}
+`;
+
+export class StatusText extends Component<StatusTextProps> {
```
ketsappi, line 52: Would be good to mention these new components, StatusText and LabelText, in the PR description somewhere.
Better if those could have been introduced in separate PR, but no need to split them now.

src/core/Form/LabelText/LabelText.baseStyles.tsx
```diff
@@ -0,0 +1,18 @@
+import { css } from 'styled-components';
+import { LabelTextProps } from './LabelText';
+import { withSuomifiTheme, TokensAndTheme } from '../../theme';
+import { font } from '../../theme/reset';
+import { Omit } from '../../../utils/typescript';
+
+export const baseStyles = withSuomifiTheme(
+  ({ theme }: TokensAndTheme & Omit<LabelTextProps, 'children'>) => css`
```
ketsappi, line 8: Are there any need to Omit anything as there seems to be no props in use from the `LabelTextProps` so it might be dropped totally?

src/core/Form/LabelText/LabelText.tsx
```diff
@@ -0,0 +1,76 @@
...
+  children: ReactNode;
+  /** Custom class name for styling and customizing  */
+  className?: string;
+  /** Hide or show label. Label element is always present, but can be visually hidden.
+   * @default visible
+   */
+  labelMode?: LabelMode;
+  /** Label span props */
+  labelSpanProps?: HtmlSpanProps;
+  /** Render the wrapping element as anohter element */
```
ketsappi, line 26: Typo `anohter` --> `another`

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -1,47 +1,265 @@
...
 const baseClassName = 'fi-search-input';
-const iconBaseClassName = `${baseClassName}_icon`;
+const searchInputClassNames = {
+  fullWidth: `${baseClassName}--full-width`,
+  error: `${baseClassName}--error`,
+  notEmpty: `${baseClassName}--not-empty`,
+  inputElement: `${baseClassName}_input`,
+  inputElementContainer: `${baseClassName}_input-element-container`,
+  functionalityContainer: `${baseClassName}_functionality-container`,
+  button: `${baseClassName}_button`,
```
ketsappi, line 77: Do buttons need this common classname? 
Could it then also be nice to have one common for icons, like `button-icon`?

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -1,47 +1,265 @@
...
+      }
+    };
+
+    const onClear = () => {
+      conditionalSetState('');
+      setTimeout(() => {
+        if (this.inputRef.current) {
+          this.inputRef.current.focus();
+        }
+      }, 100);
```
ketsappi, line 160: Is the chosen `100ms` some standard or why that amount is chosen? 
Just thought that are there some cases that it should be higher to work correctly 🤔

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -1,47 +1,265 @@
...
+                conditionalSetState(event.currentTarget.value);
+              }}
+              onKeyPress={onKeyPress}
+            />
+          </HtmlDiv>
+          <HtmlButton {...clearButtonProps}>
+            <VisuallyHidden>{clearButtonLabel}</VisuallyHidden>
+            <Icon
+              aria-hidden={true}
+              icon="close"
```
ketsappi, line 237: Not issue of this PR, but should we have separate icons for clear and close?

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: Cannot be done. Design blocks it since we show the activated button and clear button only after content has been added. Or we have to activate the search button onFocus and disable it onBlur which does not leave the decision for the users/developers either.

src/core/Form/StatusText/StatusText.tsx
aappoalander: Agreed, but did it this way because of the work process. Didn't actually know what was needed until everything was pieced together.

src/core/Form/LabelText/LabelText.baseStyles.tsx
aappoalander: I'll check if it works.

aappoalander: Probably legacy since it had the fullWidth prop in use.

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: Shared button styles are applied this way, I guess it would be possible to arrange the styles in a different way as well by extending. Icons do not share styles so common classname would only serve the purpose of custom styling which in my opinion should not be allowed in this case.

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: It's just a guess. So small that it's barely noticeable, but leaves enough time for re-rendering the component after state change (clear) before moving focus. This way screen readers read the new value rather than what was in the input before clearing.

There would probably be a better way for doing this, but that involves managed state transitions and probably some boolean logic to move focus to input with useLayoutEffect if a clear flag has been set => This kind of state transitioning would require thorough planning and a uniform way for handling for the whole library and is not worth it for this case in my opinion. If something fails, the focus is either not moved, or the old value could be read - so nothing should break in any case.

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: Maybe, but I guess it's the same one in the designs as well.

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: Yeah, should also probably rename it since it's exported with the interface. Maybe SearchInputStateValue etc.

src/core/Form/StatusText/StatusText.tsx
aappoalander: Will add to PR description.

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: Renamed to SearchInputValue and also used for the line above.

src/core/Form/StatusText/StatusText.tsx
aappoalander: Added.

src/core/Form/LabelText/LabelText.baseStyles.tsx
aappoalander: Seems to work without it as well, removed.

src/core/Form/LabelText/LabelText.tsx
aappoalander: Fixed.

src/core/Form/SearchInput/SearchInput.tsx
ketsappi: Okay, it sounds good to leave as it is now.

src/core/Form/SearchInput/SearchInput.tsx
ketsappi: We also discussed this case separately and we did not see reasons to make changes to the current implementation for now.

src/core/Form/SearchInput/SearchInput.tsx
ketsappi: Yes, let's leave those like currently and adjust later if new icons will come for both.

ketsappi (approved):

src/core/Form/SearchInput/SearchInput.tsx
LJKaski: Good point. No searching with empty strings then.

src/core/Form/StatusText/StatusText.tsx
```diff
@@ -0,0 +1,62 @@
...
+    className,
+    children,
+    tokens,
+    status,
+    ...passProps
+  }: StatusTextProps & InternalTokensProp) => (
+    <HtmlSpan
+      {...passProps}
+      className={classnames(className, baseClassName, {
+        [statusTextClassNames.error]: status === 'error',
```
LJKaski, line 42: Should the success class be added as well, just to account for all the possible values of the InputStatus?

src/core/Form/SearchInput/SearchInput.baseStyles.tsx
aappoalander: Ticket added to backlog.

src/core/Form/StatusText/StatusText.tsx
aappoalander: I'd leave this for the following issues when this is taken into use for components supporting this. Currently only used for SearchInput and it has a limited API only supporting error and default statuses.

LJKaski: That's fine for me for now. It's already good progress that it exists now :)

LJKaski (approved):

PR #401, opened 2020-10-15
[Feature] Dropdown disabled state

LJKaski:
## Description
This PR implements disabled prop and the corresponding styling for the dropdown component.

## Motivation and Context
Being able to disable a component is one of the key functions of any input component and needs to be supported.

## How Has This Been Tested?
Only tested by running locally in styleguidist.

image

## Release notes
-Added disabled prop for Dropdown component

### DIFF
--- a/src/core/Dropdown/Dropdown.test.tsx
+++ b/src/core/Dropdown/Dropdown.test.tsx
@@ -50,6 +50,16 @@ describe('Basic dropdown', () => {
     expect(input).toHaveAttribute('name', 'dropdown-test');
   });
 
+  it('should have disabled styles when disabled', async () => {
+    const { findByRole } = render(
+      <Dropdown labelText="Dropdown" disabled>
+        <Dropdown.item value={'item-1'}>Item 1</Dropdown.item>
+      </Dropdown>,
+    );
+    const button = await findByRole('button');
+    expect(button).toHaveClass('fi-dropdown--disabled');
+  });
+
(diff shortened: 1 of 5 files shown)


src/core/Dropdown/Dropdown.test.tsx
```diff
@@ -50,6 +50,16 @@ describe('Basic dropdown', () => {
     expect(input).toHaveAttribute('name', 'dropdown-test');
   });
 
+  it('should have disabled styles when disabled', async () => {
+    const { findByRole } = render(
+      <Dropdown labelText="Dropdown" disabled>
+        <Dropdown.item value={'item-1'}>Item 1</Dropdown.item>
+      </Dropdown>,
+    );
+    const button = await findByRole('button');
```
aappoalander, line 59: Are these queries really async by nature? Or is this the infamous useLayoutEffect etc. error message thing?

aappoalander:
Looks very good overall! :) Maybe we should still add an example to md?

src/core/Dropdown/Dropdown.test.tsx
LJKaski: To be honest, I just saw all the other tests being written with the async and followed suit 😅

aappoalander: This is probably the problematic case that causes a lot of warnings when executing, so maybe better to leave as is.

aappoalander (approved):

PR #403, opened 2020-10-21
[Feature] Export VisuallyHidden from the library

LJKaski:
## Description
This PR exports VisuallyHidden from the library and offers basic examples of its use.

## Motivation and Context
This component was requested by the developers using the library, and since we already had an internal implementation of it, it is easy to comply.

## How Has This Been Tested?
Tested manually in styleguidist.

## Release notes
-Add VisuallyHidden component

### DIFF
--- a/.styleguidist/styleguidist.sections.js
+++ b/.styleguidist/styleguidist.sections.js
@@ -6,6 +6,7 @@ const primitiveComponents = [
   'Button',
   'Heading',
   'Chip',
+  'VisuallyHidden',
   ['Form', 'TextInput'],
   ['Form', 'Toggle'],
   ['Form', 'SearchInput'],
@@ -77,7 +78,7 @@ module.exports = {
               ]),
             },
             {
-              name: 'Icons',
+              name: 'Icon',
               components: getComponents(['Icon', 'StaticIcon']),
             },
           ],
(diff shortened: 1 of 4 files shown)


src/core/VisuallyHidden/VisuallyHidden.md
```diff
@@ -0,0 +1,19 @@
...
+```js
+import { VisuallyHidden, SearchInput } from 'suomifi-ui-components';
+<>
+  <span>There is a hidden element under this text.</span>
+  <VisuallyHidden>
+    This element will be read by screen readers even though it's
+    visually hidden
+  </VisuallyHidden>
+
+  <SearchInput
```
aappoalander, line 12: Could we replace this with TextInput? It will be more simple after the upcoming breaking for SearchInput.

.styleguidist/styleguidist.sections.js
```diff
@@ -77,7 +78,7 @@ module.exports = {
               ]),
             },
             {
-              name: 'Icons',
+              name: 'Icon',
```
aappoalander, line 81: Why this change? It has two separate components under the heading.

src/core/VisuallyHidden/VisuallyHidden.md
LJKaski: Search input was a common example I saw people use online when talking about where VisuallyHidden styles are used. Text input not so much. That's why I opted for this.

.styleguidist/styleguidist.sections.js
LJKaski: It was to match it with the others. For example, Text and Link also contain several components, but they are in singular form in the navigation. This stood out.

src/core/VisuallyHidden/VisuallyHidden.md
aappoalander: Let's keep this as discussed elsewhere. Just have to remember to update when introducing changes to SearchInput.

aappoalander (approved):

PR #405, opened 2020-10-27
[Feature] Fix and rewrite focus styling

LJKaski:
## Description
This PR rewrites the inner structure of applying focus to components within the library. 

## Motivation and Context
The old structure got broken as some new features became available in browsers. Focus styles needed fixing.

## Release notes
-Rewrote the internal focus styling system

### DIFF
--- a/src/core/theme/utils/focus.ts
+++ b/src/core/theme/utils/focus.ts
@@ -1,6 +1,10 @@
 import { css } from 'styled-components';
 import { themeOrTokens, TokensOrThemeProps } from '../utils';
 
+import { radius } from '../radius';
+import { colors } from '../colors';
+import { zindexes } from '../zindexes';
+
 export interface FocusConfig {
   variant?: 'outline' | 'afterPseudo' | 'boxShadow';
   outline?: string;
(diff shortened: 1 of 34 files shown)


src/core/theme/utils/focus.ts
```diff
@@ -27,3 +31,25 @@ export const focus = ({
         }
       `;
 };
+
+export const absoluteFocus = css`
```
LJKaski, line 35: I'd rename this to something else as we discussed earlier. Maybe offsetFocus or pseudoFocus or something along those lines. Let's discuss.

ketsappi: We discussed this further and decided to not take actions on this at this time. Leaving it as it is now.

ketsappi (dismissed):

src/core/theme/utils/focus.ts
aappoalander: I prefer this name after all since this focus style cannot be used without absolute positioning. A longer name e.g. absolutelyPositionedFocus might be more descriptive but is already a bit cumbersome. Two hard things...

ketsappi (approved):

PR #408, opened 2020-11-02
[Fix] LanguageMenu text wrapping

ketsappi:
## Description

Long text will now wrap more nicely in the **LanguageMenu**
* In the button that opens the menu
* ...and in the opened menu

## Motivation and Context

Long text was not wrapping as wanted on the LanguageMenu.

## How Has This Been Tested?

- Locally in styleguidist

## Screenshots:
LanguageMenu_wrap

_In the screenshot the area circled shows the gap that is intentionally left on the left side, so that menu is not "touching" the edge._

## Release notes

Fix wrapping of long text in the LanguageMenu (#408)

### DIFF
--- a/src/core/LanguageMenu/LanguageMenu.tsx
+++ b/src/core/LanguageMenu/LanguageMenu.tsx
@@ -86,7 +86,7 @@ const LanguageMenuPopoverPosition = (
     top: `${targetRect.top + targetRect.height + window.pageYOffset}px`,
     maxWidth: `${Math.max(
       targetRect.width,
-      targetRect.width + targetRect.left - 20,
+      targetRect.width + targetRect.left - 30,
     )}px`,
     minWidth: `${targetRect.width - 2}px`,
   };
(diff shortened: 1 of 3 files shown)


src/core/LanguageMenu/LanguageMenu.tsx
```diff
@@ -84,6 +84,7 @@ const LanguageMenuPopoverPosition = (
     left: `${targetRect.left - popoverRect.width + targetRect.width}px`,
     // eslint-disable-next-line no-undef
     top: `${targetRect.top + targetRect.height + window.pageYOffset}px`,
+    maxWidth: `${targetRect.width + targetRect.left - 20}px`,
```
ketsappi, line 87: Currently using **20px** as a gap and this seemed to be ok for PAMU also.

But should we use token or somewhat create local variable that gives the explanation, like:
```typescript
const gap = 20
//...
maxWidth: `${targetRect.width + targetRect.left - gap}px`,
```

aappoalander (dismissed):

src/core/LanguageMenu/LanguageMenu.tsx
aappoalander: As proposed via chat, we could also use that min. of 20px margin or targetRect width, which ever is bigger at the moment. This way the min size would be the size of the targetRect (-2px) and in would allow the popover to grow to max 20px from left side.

ketsappi: _As discussed further, this gap is now 30px instead._

aappoalander (approved):

PR #409, opened 2020-11-04
[Feature] Upgrade dependencies to latest

aappoalander:
## Description
This PR upgrades all but webpack and TSDX css/scss related plugins to latest available version. Also related peer dependencies have been upgraded to the latest proven to work versions. These upgrades introduce a breaking change for styled-components types to add support for TS 4 and also gain performance and security improvements. These types include a known issue with skipLibCheck true compiler option, but a workaround is described in Readme. 

## Motivation and Context
To allow use of TypeScript 4 and later with skipLibCheck true compiler option, some dependencies had to be upgraded. Previously used types version 4.1.8 for styled-components had to be updated to latest available version, which has a known issue of including conflicting react-native types, but fortunately there is a workaround for the issue. Another reason for upgrade is that Styled Components v5 performs better than v4 and does not introduce breaking changes.

Jest related libraries are manually configured to use latest version in contrast to what is officially supported by TSDX. This is because of security issues in older versions. 

Styleguidist does not allow upgrading webpack to version 5+ and related plugins also need to remain 4.x compatible. TSDX css/scss related plugins cannot be upgraded either.

## How Has This Been Tested?

Tested with Styleguidist, CRA TS and SSR test projects and using multiple browsers.

## Release notes
- Breaking Change: Styled components and related types upgraded to version 5 and require configuration to allow skipLibCheck true. Also peer dependencies have been updated accordingly. Additional non breaking dependency upgrades.


LJKaski (approved):
Seems to work fine with the updated dependencies.

ketsappi (approved):

PR #411, opened 2020-11-16
[Feature] TextInput improvements

ketsappi:
## Description

- Add example how to use `optionalText` with TextInput
- **Breaking change**: Remove the `labelProps`, `labelTextProps` and `children` props from TextInput
- **Breaking changes**: Classname changes because of taking internal components in to use (HintText, LabelText, StatusText)
- Internal changes
  - Remove the TextInput from components level
  - Create internal component, HintText
  - Change TextInput to use HintText
  - Add `optionalText`-prop for LabelText
  - Add tests for LabelText
  - Fix small typo in SearchInput
  - Change SearchInput to use `InputStatus` type
  - Add tests for StatusText
  - Fix small typos in StatusText JSDoc

## Motivation and Context

To increase developer experience and simplify the internal structures.

* We no longer need to repeat same labelText, hintText and statusText all over project as those can now be reused.

* By removing the components level we decrease the difficulty of adding new features.

* Added tests might cause some errors in future to get caught early.

  

## How Has This Been Tested?

* Locally in styleguidist, CRA-TS, next.js (SSR) with VoiceOver.
* In IE 11 with BrowserStack.

## Release notes

- **Breaking changes**: TextInput - Remove `labelProps`, `labelTextProps` and `children` props + classname changes because of taking internal components in to use


src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -18,10 +18,11 @@ import { StatusText } from '../StatusText/StatusText';
 import { LabelText, LabelMode } from '../LabelText/LabelText';
 import { Icon } from '../../Icon/Icon';
 import { baseStyles } from './SearchInput.baseStyles';
+import { InputStatus } from '../types';
 
 type SearchInputValue = string | number | undefined;
 
-type SearchInputStatus = 'default' | 'error';
+type SearchInputStatus = Exclude<InputStatus, 'success'>;
```
aappoalander, line 25: Interesting, didn't know there was such a thing as Exclude. Seems to have similar behavior with Omit for interfaces.

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -287,6 +288,6 @@ const StyledTextInput = styled(
  */
 export class SearchInput extends Component<SearchInputProps> {
   render() {
-    return <StyledTextInput {...withSuomifiDefaultProps(this.props)} />;
+    return <StyledSearchInput {...withSuomifiDefaultProps(this.props)} />;
```
aappoalander, line 291: Good catch, this was probably due to old implementation that used TextInput.

src/core/Form/LabelText/LabelText.tsx
```diff
@@ -25,13 +25,16 @@ interface InternalLabelTextProps extends HtmlDivProps {
   labelSpanProps?: HtmlSpanProps;
   /** Render the wrapping element as another element */
   asProp?: asPropType;
+  /** Optional text that is shown after labelText. Will be wrapped in parentheses. */
```
LJKaski, line 28: Should this be described more like "An optional input indicator text that is..." or "text to mark a field optional --" so that the purpose is clearer from the description.

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -1,74 +1,117 @@
...
+  onBlur?: (event: FocusEvent<HTMLInputElement>) => void;
+  /** Label */
+  labelText: string;
+  /** Hide or show label. Label element is always present, but can be visually hidden.
+   * @default visible
+   */
+  labelMode?: LabelMode;
+  /** Placeholder text for input. Use only as visual aid, not for instructions. */
+  visualPlaceholder?: string;
+  /** A custom element to be passed to the component. Will be rendered after the input */
```
LJKaski, line 57: What's the use case for the children prop nowadays? I think it was used for the search icon in the old structure, but what are the other ones?

ketsappi: Good question.
I have wondered the same in the past, but can't remember that there was any other specific use case.

I would suggest that we can get rid of this prop also.

LJKaski: I guess it's good for general flexibility. Maybe for tooltips or such, but not sure if something like that needs to be supported inside the component. Might be something we can get rid of, but let's think it through well before doing anything.

src/core/Form/LabelText/LabelText.tsx
aappoalander: I find the latter better understandable "text to mark a field optional --", could be also phrased like "Text to indicate an optional field"

src/core/Form/TextInput/TextInput.tsx
aappoalander: I propose we remove it. We are moving away from unstyled versions anyways, so why support something that probably break the layout? And there will be breaking changes to the API anyways, so now would be a good time in my opinion.

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -78,15 +121,62 @@ export class TextInput extends Component<TextInputProps> {
...
+        <LabelText
+          htmlFor={this.id}
+          labelMode={labelMode}
+          as="label"
+          optionalText={optionalText}
+        >
+          {labelText}
+        </LabelText>
+        <HintText id={this.hintTextId}>{hintText}</HintText>
+        <HtmlDiv className={textInputClassNames.functionalityContainer}>
```
aappoalander, line 143: Functionality container is probably unnecessary complexity for text input. This was introduced for search input because of visual glitches in shadows when focus styles were applied to multiple components (input, clear button, search button). Should be possible to combine functionalityContainer and inputElementContainer for this component.

src/core/Form/LabelText/LabelText.tsx
ketsappi: This is now changed to:
```
/** Text to mark a field optional. Will be wrapped in parentheses and shown after labelText. */
```

src/core/Form/TextInput/TextInput.tsx
ketsappi: Good find.
This `functionalityContainer` is now removed.

src/core/Form/TextInput/TextInput.tsx
ketsappi: `children` prop is now removed. Updated the PR description accordingly.

LJKaski (approved):
Tested with CRA, DS-site with and without NVDA. Also tested with IE and it seems to work as it should. Good work :)

aappoalander (approved):

PR #412, opened 2020-11-23
[Feature] Checkbox adjustments

ketsappi:
## Description

* Use internal components for `hintText` and `statusText` props
  * which causes this to be breaking change as the possible uses of customisation based on the classnames of these props
* Use new AutoId component that is created based on [reach-ui Auto ID](https://reach.tech/auto-id/)
  * Using that to generate automatic id's to better work with SSR setup to have more consistent id's versus client & server
* Add more tests for Checkbox
* Remove Checkbox from components level

### AutoId component

Here is the example how to use the **AutoId** component. It uses the render prop logic.
If the own id is given, `propId`, it will use that. In other case, it will generate one automatically.

https://github.com/vrk-kpa/suomifi-ui-components/blob/1fdd87d0afedd3168278564a91247df1eb4211f2/src/core/Form/Checkbox/Checkbox.tsx#L235-L237

## Motivation and Context

To better work with SSR setup and to increase developer experience by removing duplicate code _(by using internal components)_
And by removing from the components level, we decrease the complexity of future changes to the component.

## How Has This Been Tested?

* Locally in styleguidist with VoiceOver
* In next.js (SSR) with VoiceOver

## Release notes

**Breaking changes**: Styles classnames changes for Checkbox's `hintText` and `statusText`


src/utils/AutoId.tsx
```diff
@@ -0,0 +1,51 @@
+/** Based heavily on https://github.com/reach/reach-ui/tree/develop/packages/auto-id */
```
ketsappi, line 1: Is this enough mention of the use of idea of reach-ui's Auto ID? 

Basically we have just few minor changes to it get it working in our setup and our own wrapper component `<AutoId>`.

src/utils/AutoId.tsx
```diff
@@ -0,0 +1,51 @@
...
+import React, { useState, useEffect } from 'react';
+
+let serverHandoffComplete = false;
+let theId = 0;
+const genId = () => {
+  theId += 1;
+  return theId;
+};
```
ketsappi, line 9: This was originally like this, but linters complained about it ([np-plus-plus](https://eslint.org/docs/rules/no-plusplus) and [no-return-assign](https://eslint.org/docs/rules/no-return-assign)). 
```tsx
const genId = () => ++theId;
```

Current implementation satisfies the linter and was not happy to add ignore comment for linter just for this.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
```diff
@@ -139,20 +136,15 @@ export const baseStyles = withSuomifiTheme(
...
 
-    & .fi-checkbox_status {
+    & .fi-status-text {
       display: block;
-      margin-top: ${theme.spacing.xxs};
-      color: ${theme.colors.blackBase};
-      font-size: 14px;
       line-height: 18px;
-      font-weight: 600;
     }
```
ketsappi, line 148: Left the styles that was required to keep the visuals to left untouched.

src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -27,17 +34,62 @@ const iconClassnames = {
...
+   * Hint text to be displayed under the label.
+   */
+  hintText?: string;
+  /**
+   * aria-label for the HTML input-element,
+   * alternatively you can define aria-labelledby with label-element id
+   */
+  'aria-label'?: string;
+  'aria-labelledby'?: string;
+  'aria-describedby'?: string;
```
ketsappi, line 80: `aria-describedby` was added to interface as it was required to work as the Checkbox is not extending any interface that brings it.

For example, TextInput extends`HtmlInputProps`

https://github.com/vrk-kpa/suomifi-ui-components/blob/dccf99d269c4e9f1e9c0df1901aac8fa32288fce/src/core/Form/TextInput/TextInput.tsx#L33-L35

So, should **Checkbox** also extend it or was there reason why it is not already doing this?

src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -66,50 +118,136 @@ class DefaultCheckbox extends Component<CheckboxProps> {
...
+        return {
+          'aria-describedby': [
+            ...(statusText ? [statusTextId] : []),
+            ...(hintText ? [hintTextId] : []),
+            ...(ariaDescribedBy ? [ariaDescribedBy] : []),
+          ].join(' '),
+        };
+      }
+      return {};
+    };
```
ketsappi, line 172: This is done like this to prevent empty spaces in classname if some are not given.

src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -66,50 +118,136 @@ class DefaultCheckbox extends Component<CheckboxProps> {
...
+          {!!checkedState && (
+            <Icon
+              icon="check"
+              className={classnames(iconBaseClassName, {
+                [iconClassnames.checked]: checkedState && !disabled,
+                [iconClassnames.error]: status === 'error' && !disabled,
+                [iconClassnames.disabled]: !!disabled,
+              })}
+            />
+          )}
```
ketsappi, line 217: Would it be nice if we try to minimise the logic in the "main return"? 

So e.g the logic for showing the icon would be like this:

```
    const CheckedIcon = () => (
      <>
        {!!checkedState && (
          <Icon
            icon="check"
            className={classnames(iconBaseClassName, {
              [iconClassnames.checked]: checkedState && !disabled,
              [iconClassnames.error]: status === 'error' && !disabled,
              [iconClassnames.disabled]: !!disabled,
            })}
          />
        )}
      </>
    );
```

And using it:
```
          <CheckedIcon />
```

src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -66,50 +118,136 @@ class DefaultCheckbox extends Component<CheckboxProps> {
...
 
+const StyledCheckbox = styled(
+  ({
+    tokens,
+    id: propId,
+    ...passProps
+  }: CheckboxProps & InternalTokensProp) => (
+    <AutoId id={propId}>
+      {(id) => <BaseCheckbox id={id} {...passProps} />}
+    </AutoId>
```
ketsappi, line 237: Added the use of AutoId here instead of the exporting of the class. 
Now it only needs to be told once. In export part it would have been be told twice; for render and for the static.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
aappoalander: Is this different from other use cases?

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
```diff
@@ -85,7 +82,7 @@ const largeVariantStyles = ({ theme }: SuomifiThemeProp) => css`
         top: 4px;
       }
     }
-    & .fi-checkbox_hintText {
+    & .fi-hint-text {
```
aappoalander, line 85: These are only for Checkbox? Not sure if it matters, but this implementation relies on internal structure of hint text rather than defining the padding on the container of the element. The components are of course closely coupled anyways, but this would probably be an easy change? If it requires hacky solutions then I'd also prefer using this one.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
aappoalander: Or is this something that could / should be applied to status-text styles instead?

src/core/Form/Checkbox/Checkbox.tsx
aappoalander: In my opinion components should extend the interface of the element where passprops are given. E.g. if everything is wrapped inside a div and passProps go there, it should extend HtmlDiv props (Or low level API if no resets are used). Or if passprops go to input element, that is what should be included in the API as well.

This is a bit difficult topic over all, but I think a good rule of thumb is to extend the thing the component is mostly about, input, text (can be div etc), button and so on.

aappoalander: And in this case if Checkbox would extend HtmlInputProps or similar, we could provide containerProps: HtmlDivProps if necessary. In some cases it's also possible that extending would provide too much undesired or conflicting functionality and can be done the other way around by picking only necessary/needed props. So has to be considered case by case.

src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -66,50 +118,136 @@ class DefaultCheckbox extends Component<CheckboxProps> {
...
+            ...(statusText ? [statusTextId] : []),
+            ...(hintText ? [hintTextId] : []),
+            ...(ariaDescribedBy ? [ariaDescribedBy] : []),
+          ].join(' '),
+        };
+      }
+      return {};
+    };
+
+    const newCheckboxInputProps = {
```
aappoalander, line 174: Id prefer to call these derived or extendedCheckboxInputProps rather than new. Would indicate the use case better.

src/core/Form/Checkbox/Checkbox.tsx
aappoalander: In this case I'd prefer showing the logic for hiding in the template. If CheckedIcon is introduced inside render then props won't be an issue - not worth moving outside in my opinion. How about:
```
const CheckedIcon = () => (
          <Icon
            icon="check"
            className={classnames(iconBaseClassName, {
              [iconClassnames.checked]: checkedState && !disabled,
              [iconClassnames.error]: status === 'error' && !disabled,
              [iconClassnames.disabled]: !!disabled,
            })}
          />
    );
```
And
`{!!checkedState && (<CheckedIcon />)}`
in the template

src/utils/AutoId.tsx
aappoalander: Might be sufficient as is. Could also be beneficial to add some explanation on why an how, not the whole story from the link.

src/utils/AutoId.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+function canUseDOM() {
+  return !!(
+    typeof window !== 'undefined' &&
+    window.document &&
+    window.document.createElement
+  );
+}
+
+const useIsomorphicLayoutEffect = canUseDOM()
```
aappoalander, line 19: I'd use more descriptive name like enhancedEffect (this name is from Material UI). More importantly, I think this should be moved outside AutoId to maybe a separate util or to common TS utils. This is used already at least in Popover/Portal implementation and is not only related to Id's. Should also be updated to Popover.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
ketsappi: That is good question, not sure which way is the most used way at the moment: which "wins".

These seems to be needed also when doing similar changes for _**Textarea**_ component.

**HintText**
` margin-bottom: 0;`

**StatusText**
`display: block;` and `line-height: 18px;`

src/core/Form/Checkbox/Checkbox.tsx
ketsappi: Yes, that might be better; seeing the reason in the template why it might not be showing (even though it is quite logical from the name of the component.. but still).

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
ketsappi: It could be done like this, **IF**, the status text wouldn't have different padding (0px) than the label and hint.

```
    & > * {
      padding-left: ${theme.spacing.xxl};
    }
```

I don't know other ways to add the style without specifying different style for hint/status text element, like the current implementation.

src/core/Form/Checkbox/Checkbox.tsx
ketsappi: Or should we get rid of this and told it in directly where the component is?

```
<HtmlInput type="checkbox" 
  disabled,
  id,
  'aria-label': ariaLabel,
  'aria-labelledby': ariaLabelledBy,
  'aria-invalid': status === 'error',
  checked: !!checkedState,
  className: checkboxClassNames.input,
  onChange: this.handleClick,
  name,
  ...getDescribedBy(),
  ...(value ? { value } : {}),
/>
```

As these props are only for input and not used for anything else.

src/core/Form/Checkbox/Checkbox.tsx
ketsappi: This is now changed as suggested.

src/utils/AutoId.tsx
ketsappi: Added some explanation/reasoning as a comment.

src/utils/AutoId.tsx
ketsappi: This is now changed to `useEnhancedEffect` and is located in `suomifi-ui-components/src/utils/common.ts`. 
We can name the util better if needed.

src/core/Form/Checkbox/Checkbox.tsx
ketsappi: We discussed about this and we leave it now as it is.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
aappoalander: That's a good question. Optimally, only styles shared by all components should be in status/hintText. In my opinion this is also a design issue, not sure why we should have context specific styles for the same component. Another option, if possible, is to use the styles in the wrapping component instead and perhaps the components should not introduce any outside margins etc. and leave that to wrapping components?

src/core/Form/Checkbox/Checkbox.tsx
aappoalander: As discussed elsewhere, this is ok for now. Using HtmlInputProps as passProps would introduce quite a few conflicts and would unnecessarily pollute the API without providing any mandatory functionality in addition to what is now available.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
aappoalander: Ah, my bad. This is only to apply outer margins / padding to the component. As proposed earlier, maybe we should change the styles of hint text and leave the outer padding / margin to wrapping components like done here.

src/core/Form/Checkbox/Checkbox.tsx
aappoalander: Makes sense! Leaving the complex getDescribedBy etc. logic outside of the template is good, but separating the props completely from the context is unnecessary.

src/utils/AutoId.tsx
aappoalander: As this is so heavily based on Reach-UI implementation, I think we should include the Reach MIT Licence to our LICENCE file. Otherwise the implementation could have similar idea, but should be really our own. https://github.com/reach/reach-ui/blob/develop/LICENSE

ketsappi: **reach-ui**'s MIT license is now added to our license.

src/core/Form/Checkbox/Checkbox.tsx
ketsappi: This is now changed.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
aappoalander: Decided to fix these with another issue / branch.

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
aappoalander: Decided to fix these with another issue / branch.

aappoalander (approved):

PR #413, opened 2020-11-27
[Feature] Textarea adjustments

ketsappi:
## Description

**Textarea**
* Remove from components level
* Change to use AutoId to generate id, instead of idGenerator
* Change to use internal components for label-, hint- and status texts
* Fix the empty space bug with `aria-describedby`
* Add more tests
* Add `fullWidth` prop
* Add `containerProps` prop
* Add default width for Textarea, 290px (_same as for TextInput_)
* Change `placeholder` attribute tone omitted as we have the `visualPlaceholder` prop for this.

**Other**
* Add missing color tokens to `suomifi-ui-components/src/core/theme/colors.ts`
  * depthDark2
  * depthDark3
* Add `getAriaDescribedByProp` helper to `src/utils/aria.ts`
* Move existing aria utils from `src/components/utils/aria.ts` level to `src/utils/aria.ts`

## Motivation and Context

To better work with SSR setup and to increase developer experience by removing duplicate code (by using internal components)

## How Has This Been Tested?

* Locally in styleguidist with VoiceOver
* In next.js (SSR) with VoiceOver

## Release notes

* **Breaking changes**: Style classnames changes for Textarea's `labelText`,  `hintText` and `statusText`
* Fix the empty space bug with `aria-describedby` on Textarea
* Add `fullWidth` and `containerProps` props for Textarea


src/core/Form/Textarea/Textarea.tsx
```diff
@@ -1,52 +1,161 @@
...
-
 const textareaClassNames = {
+  textareaContainer: `${baseClassName}_textarea-element-container`,
+  textarea: `${baseClassName}_textarea`,
   disabled: `${baseClassName}--disabled`,
   error: `${baseClassName}--error`,
 };
 
-export interface TextareaProps extends CompTextareaProps, TokensProp {}
+type TextareaStatus = Exclude<InputStatus, 'success'>;
```
ketsappi, line 22: Now only TextInput is having all statuses: **_default, error and success_**

Textarea and Checkbox is excluding the **success** out from the InputStatus. Should we change it other way around as it seems that only TextInput will use the success?

https://github.com/vrk-kpa/suomifi-ui-components/blob/de3ba041127f51eb2df5c75ff152933eefc16e13/src/core/Form/Checkbox/Checkbox.tsx#L37

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -1,52 +1,161 @@
...
+
+const StyledTextarea = styled(
+  ({
+    tokens,
+    id: propId,
+    ...passProps
+  }: TextareaProps & InternalTokensProp) => (
+    <AutoId id={propId}>
+      {(id) => <BaseTextarea id={id} {...passProps} />}
+    </AutoId>
```
ketsappi, line 151: Added **AutoId** here to be consistent as it is added here with _Checkbox_ also.

src/core/Form/Textarea/Textarea.baseStyles.tsx
```diff
@@ -16,20 +16,12 @@ export const baseStyles = withSuomifiTheme(
       flex-direction: column;
```
ketsappi, line 16: Should we have width for Textarea, like we are having for TextInput?

https://github.com/vrk-kpa/suomifi-ui-components/blob/de3ba041127f51eb2df5c75ff152933eefc16e13/src/core/Form/TextInput/TextInput.baseStyles.tsx#L7-L13

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -1,52 +1,162 @@
...
+  /** Optional text that is shown after labelText. Will be wrapped in parentheses. */
+  optionalText?: string;
+  /**
+   * Unique id
+   * If no id is specified, one will be generated automatically
+   */
+  id?: string;
+  /** Input name */
+  name?: string;
+}
```
ketsappi, line 68: Do we need `fullWidth` prop for Textarea like with TextInput?

https://github.com/vrk-kpa/suomifi-ui-components/blob/de3ba041127f51eb2df5c75ff152933eefc16e13/src/core/Form/TextInput/TextInput.tsx#L71-L72

src/core/Form/Textarea/Textarea.tsx
aappoalander: I wouldn't change is as it is a part of status text component functionality. Changing it would lead to "customization" and move parts of implementation (styles) outside the component.

src/core/Form/Textarea/Textarea.tsx
aappoalander: Wouldn't hurt I guess. Not sure if there is a design need for this, but would improve usability at least and cover more use cases (e.g. switching to full available width in mobile view).

src/core/Form/Textarea/Textarea.baseStyles.tsx
aappoalander: That is question to ask from the Design Team I guess? Would be good to have something and when thinking about the component usage - one should be able to set the display and size attributes without breaking the component.

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -1,52 +1,162 @@
...
+      children,
+      onClick,
+      labelMode,
+      labelText,
+      hintText,
       status,
-      className,
+      statusText,
+      visualPlaceholder,
+      resize: dismissResize,
```
aappoalander, line 84: This pattern of renaming to indicate discarding a prop is used inconsistently in this repo. My proposal would be not to use it since it's not a commonly used best practice and is easy to forget.

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -1,52 +1,162 @@
...
+    const statusTextId = `${id}-statusText`;
+    const hintTextId = `${id}-hintText`;
+
+    const getDescribedBy = () => {
+      if (statusText || hintText || ariaDescribedBy) {
+        return {
+          'aria-describedby': [
+            ...(statusText ? [statusTextId] : []),
+            ...(hintText ? [hintTextId] : []),
+            ...(ariaDescribedBy ? [ariaDescribedBy] : []),
```
aappoalander, line 100: This is purely cosmetic, but in case there is only one element in the array, resulting string will contain an empty space after the id. I'd propose we write an ariaIdsTool to do this work nicely. So that you can pass in an array of string | undefined and the tool returns you a string.

aappoalander: The tool should be used in multiple places already, not worth doing for only this use case.

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -1,52 +1,162 @@
...
+   * Unique id
+   * If no id is specified, one will be generated automatically
+   */
+  id?: string;
+  /** Input name */
+  name?: string;
+}
 
-export class Textarea extends Component<TextareaProps> {
+class BaseTextarea extends Component<TextareaProps> {
```
aappoalander, line 70: Should probably omit some props here? E.g. type and if something else is limited by the component structure or functionality.

src/core/Form/Textarea/Textarea.tsx
LJKaski: I think it would be a welcome addition.

src/core/Form/Textarea/Textarea.tsx
ketsappi: As discussed, this will be just destructured without renaming, so it is now:
`resize,`

LJKaski (approved):

src/core/Form/Textarea/Textarea.tsx
ketsappi: This is now done with the util located `src/utils/aria.ts`

src/core/Form/Textarea/Textarea.tsx
ketsappi: `fullWidth` prop is now added to Textarea.

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -1,52 +1,168 @@
...
+   * Unique id
+   * If no id is specified, one will be generated automatically
+   */
+  id?: string;
+  /** Input name */
+  name?: string;
+  /** Set components width to 100% */
+  fullWidth?: boolean;
+  /** Textarea container div props */
+  containerProps?: Omit<HtmlDivProps, 'className'>;
```
ketsappi, line 78: Name is `containerProps` as it was silly to repeat the name, e.g `textareaContainerProps`.

src/utils/aria.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+/**
+ * Set element ability to be focusable based on aria-label
+ * @param {String} ariaLabel optional aria-label
+ */
+export const ariaFocusableNoLabel = (ariaLabel?: string) => {
+  return ifAriaNoLabel(ariaLabel) ? {} : { focusable: false };
+};
+
```
ketsappi, line 16: These are the moved utils from `src/components/utils/aria.ts`

src/utils/aria.ts
```diff
@@ -0,0 +1,45 @@
...
+ * />
+ * @param describedByIds Array of id-strings
+ * @returns Object with 'aria-describedby' if there is atleast one value that is not undefined. Otherwise returns empty Object.
+ */
+export const getAriaDescribedByProp = (
+  describedByIds: (string | undefined)[],
+):
+  | { 'aria-describedby': string }
+  | {
+      'aria-describedby'?: undefined;
```
aappoalander, line 35: should this return type rather be just `{}` if that is not allowed by linters, isn't the latter option enough, since it covers both cases: an empty object and aria-describedby string?

ketsappi: If using the latter, it will complain:
`Type 'string' is not assignable to type 'undefined'.ts(2322)`

If returning type is `{}` it does work, but then JSDoc obviously can not show the return type that it will have that `aria-describedby` inside of it if that is the case. But this tool is for internal use, so we can live without this information too.

src/core/Form/Textarea/Textarea.tsx
ketsappi: Not sure what props those could be and there was no change to that part regarding this PR, but I'm open for suggestions to improve it.

It seems that the textarea [has no](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea) `type` attribute and when tested, it has therefore no effect if given.

src/core/Form/Textarea/Textarea.baseStyles.tsx
ketsappi: This is asked from the design team, waiting for response.

src/core/Form/Textarea/Textarea.tsx
aappoalander: I guess it's not mandatory if there really is no functionality that would break if some props like readonly or some event listeners are given. This is completely possible in this case, just seemed weird that there would be nothing to omit.

aappoalander: lib.dom.d.ts has a loads of properties for HTMLTextAreaElement which is the underlying type.

src/utils/aria.ts
aappoalander: Yeah, was just interested why the typing was done this way.

src/core/Form/Textarea/Textarea.tsx
ketsappi: Okay, I will go through those and check if causing any troubles.

src/core/Form/Textarea/Textarea.baseStyles.tsx
ketsappi: They said to use same default width as with TextInput, ` width: 290px;` if it does not cause any problems.

src/core/Form/Textarea/Textarea.tsx
ketsappi: `placeholder` attribute is now omitted as we have `visualPlaceholder` prop for it.

src/utils/aria.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    expect(expectation).toEqual(data);
+  });
+
+  it('has the given id', () => {
+    const data = getAriaDescribedByProp(['other-id']);
+    const expectation = { 'aria-describedby': 'other-id' };
+    expect(expectation).toEqual(data);
+  });
+
+  it('has all the given ids', () => {
```
aappoalander, line 16: Good to have tests for this as well even though the util is quite simple. I'd add a fourth undefined id here in the middle of the array to ensure no additional whitespace is entered / everything is working as expected.

In addition, I think we should start considering dividing the tests into sets. Like CI = all, Hooks = smoke

ketsappi: Added the suggested `undefined` between the existing items to this test.

Should I rename this test to `aria.unit.test.ts`?
It can be then run as `yarn test unit`. _Or just like before, `yarn test`_

src/core/Form/Textarea/Textarea.test.tsx
```diff
@@ -276,4 +276,16 @@ describe('props', () => {
       expect(container.firstChild).toHaveClass('fi-textarea--full-width');
     });
   });
+
+  describe('containerProps', () => {
+    it('has the given props on the container', () => {
+      const { container } = render(
+        <Textarea
+          labelText="label"
+          containerProps={{ style: { width: 100 } }}
```
aappoalander, line 285: Is this without unit intentionally?

src/core/Form/Textarea/Textarea.md
```diff
@@ -64,5 +64,13 @@ import { Textarea } from 'suomifi-ui-components';
     Lorem ipsum dolor sit amet, consectetur adipiscing elit. Duis
     vestibulum iaculis augue, sit amet tincidunt ipsum.
   </Textarea>
+
+  <Textarea
+    labelText="Textarea with fixed width of 250px"
+    containerProps={{ style: { width: 250 } }}
```
aappoalander, line 70: Is this without unit intentionally?

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -98,6 +106,7 @@ class BaseTextarea extends Component<TextareaProps> {
 
     return (
       <HtmlDiv
+        {...containerProps}
```
aappoalander, line 111: Is className in the main props? I assume it is applied below, but now also containerProps includes a classname. This needs to be applied below or omitted from props in my opinion.

src/core/Form/Textarea/Textarea.test.tsx
ketsappi: Yes, it is. It is now done like React assumes the unit to be `px` in this case. 
It can be done also like this:
`containerProps={{ style: { width: '100px' } }}`

src/core/Form/Textarea/Textarea.md
ketsappi: See the reply from above.

src/core/Form/Textarea/__snapshots__/Textarea.test.tsx.snap
```diff
@@ -129,6 +129,7 @@ exports[`snapshot default structure should match snapshot 1`] = `
   -ms-flex-direction: column;
   flex-direction: column;
   color: hsl(0,0%,16%);
+  width: 290px;
```
aappoalander, line 132: Have you tested that it's still possible to override this with css? It should be, but just to make sure.

src/core/Form/Textarea/Textarea.tsx
ketsappi: It should be already be omitted from the TextareaProps's `containerProps`:

`containerProps?: Omit<HtmlDivProps, 'className'>;`

src/core/Form/Textarea/__snapshots__/Textarea.test.tsx.snap
ketsappi: It works with the following example:

```css
.fi-textarea.ta--custom.ta--custom { 
  width: 500px;
}
```
And giving the `ta--custom` className for Textarea:
```jsx
className="ta--custom"
```

src/utils/aria.test.ts
aappoalander: I wouldn't do the dividing / modularization now and the need for it comes more from reducing test running time for commonly repeating tasks like push rather than anything else.

src/core/Form/Textarea/Textarea.test.tsx
aappoalander: Which would be more common use case? We have used values with units previously I think.

src/core/Form/Textarea/Textarea.tsx
aappoalander: So it seems, missed it while doing the review. Sorry!

src/utils/aria.test.ts
ketsappi: Okay, let's leave it for now. 

It really would be nice to have the tests run by git-hooks to be faster.

src/core/Form/Textarea/Textarea.test.tsx
ketsappi: `TextInput` and `SearchInput` seems to be the only places where is similar situation and it does tell the width as string to explicitly tell the unit to be `px`.

I will change these occurrences to use the string format also.

aappoalander (approved):

PR #414, opened 2020-11-30
[Feature] Expander improvements

aappoalander:
## Description
This PR introduces composition pattern for Expander.
- Expander content is now described using ExpanderTitle and ExpanderContent components that can contain ReactNodes.
- Expander groups can now contain ReactNodes in addition to expanders

In addition, multiple internal improvements are included:
- Components level implementation is removed
- AutoId is now used when generating id's automatically
- Aria support has been improved according to: https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html
- Focus indicators now use focus-visible
- Handling of children is now completely based on context rather than child manipulation

## Related Issue

In addition multiple bugfixes are included:
- Content is no longer focusable/focused after closing an expander.
- Support for custom open/close all button is removed
- Aria attributes to support screen readers with open/close all button added
- Open/Close button stays in sync

## Motivation and Context

Expander and ExpanderGroup implementation was no longer up-to-date with design specification and had quite a few issues. It was better to do a larger refactor rather than trying to address all issues with small fixes.

## How Has This Been Tested?

Tested with TBA (CRA-TS, SSR, Styleguidist, DS-Site) using supported browsers and devices

- No changes to visuals

## Release notes
- **Breaking changes:** Complete API redesign and switch to content composition pattern. See #414
  - Expander API
    - Removed: titleTag, title, titleProps, contentProps and internally used expanderGroup and index
    - Renamed: onClick => onOpenChange
    - Added:  id
  - ExpanderGroup API
    - Static version Expander.group is no longer available. Use ExpanderGroup instead
    - Removed: onClickAll
    - Renamed: OpenAll => OpenAllText, CloseAll => CloseAllText
    - Added: AriaCloseAllText, AriaCloseAllText, toggleAllButtonProps, className

Screenshot 2020-11-30 at 16 47 00


src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
```diff
@@ -0,0 +1,190 @@
...
+import { baseStyles } from './ExpanderGroup.baseStyles';
+import { VisuallyHidden } from '../../../components';
+
+const baseClassName = 'fi-expander-group';
+const openClassName = `${baseClassName}--open`;
+const expandersContainerClassName = `${baseClassName}_expanders`;
+const openAllButtonClassName = `${baseClassName}_all-button`;
+
+type ToggleAllExpanderState = {
+  toState: boolean;
```
LJKaski, line 16: I think this might be clearer with a more descriptive name. Now it's named kinda like a function even though it isn't one. Could it be "allOpen" or something along those lines?

aappoalander: This is the original name and indicates the target state that the expanders should switch to unless already there. Maybe we could use `targetState` or `targetOpenState`? Any thoughts @ketsappi ?

LJKaski: Ah, remembered the logic a bit differently. Still I'd maybe rather have the the name tell what state the value represents and handle what to do with it elsewhere. It feels likea a double edged sword in this case though. I'd be down with `targetState` or `targetOpenState` as well, since they are quite discriptive already.

ketsappi:
As the static version, `Expander.group` is no longer available because the API changes it might be good to mention that in release notes & PR description.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: That would be ok, since it does not represent the current state nor necessarily the coming future state since some of the expanders might be controlled and refuse to change their state.

jest.setup.js
```diff
@@ -5,3 +5,8 @@ require('jest-axe/extend-expect');
 require('@testing-library/jest-dom/extend-expect');
 
 require('jest-styled-components');
+
+jest.mock('react', () => ({
+  ...jest.requireActual('react'),
+  useLayoutEffect: jest.requireActual('react').useEffect,
+}));
```
ketsappi, line 12: Would this be good place for those rare comments? To reason why this is done.

src/core/Expander/Expander/Expander.baseStyles.tsx
```diff
@@ -0,0 +1,28 @@
+import { css } from 'styled-components';
+import { withSuomifiTheme, TokensAndTheme } from '../../theme';
+import { element, font } from '../../theme/reset';
+
+import { ExpanderProps } from './Expander';
+
+/* stylelint-disable no-descending-specificity, no-duplicate-selectors */
```
ketsappi, line 7: I think this ignore comment is unneeded. Couldn't produce any linter issue when removed the line.

src/core/Expander/Expander/Expander.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+  it('classnames will be removed when clicked', () => {
+    const mockClickHandler = jest.fn();
+    const { getByTestId, getByRole } = render(
+      DefaultOpenExpander({ onOpenChange: mockClickHandler }),
+    );
+    const buttonToClick = getByRole('button');
+    const titleDiv = getByTestId('expander-open-by-default-title');
+    fireEvent.click(buttonToClick);
+    expect(titleDiv).toHaveClass('fi-expander_title');
+    expect(titleDiv).not.toHaveClass('fi-expander_title--open');
```
ketsappi, line 109: When using the `not.toHaveClass` it might be good to first check **before** the click to make sure that it existed in the first place using `toHaveClass`.

But in this test though, the existence is checked in the previous test. Maybe this is just note to myself :)

src/core/Expander/Expander/Expander.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+    expect(button.querySelector('svg')).toHaveClass(
+      'fi-expander_title-icon--open',
+    );
+  });
+});
+
+test('CSS export', () => {
+  const css = cssFromBaseStyles(baseStyles);
+  expect(css).toEqual(expect.stringContaining('background-color'));
+});
```
ketsappi, line 171: This have been here before this PR, but should we remove it as it seems quite useless.

src/core/Expander/Expander/Expander.tsx
```diff
@@ -0,0 +1,226 @@
...
+        {...passProps}
+        className={classnames(className, baseClassName, {
+          [openClassName]: !!openState,
+        })}
+      >
+        <ExpanderProvider
+          value={{
+            open: openState,
+            contentId,
+            titleId,
```
ketsappi, line 178: Would just the `id` be enough to be here? 
Then using it in later places to make those `{id}_content` and `{id}_title`.

src/core/Expander/Expander/Expander.baseStyles.tsx
aappoalander: Will remove. It was needed before breaking down the styles.

jest.setup.js
aappoalander: Let's see if I can come up with reasonable comments without getting too much into details.

src/core/Expander/Expander/Expander.tsx
```diff
@@ -0,0 +1,226 @@
...
+   */
+  children: ReactNode;
+  /** Custom classname to extend or customize */
+  className?: string;
+  /**
+   * Id for expander, must be unique. Duplicate id's break ExpanderGroup functionality.
+   * Autogenerated if not provided
+   */
+  id?: string;
+  /** Default status of expander open
```
ketsappi, line 53: > Expander is opened by default

Would this be better? As the possible values are `true/false`

src/core/Expander/Expander/Expander.test.tsx
aappoalander: These are expected in the previous test which uses the same component and props. If this is not sufficient, maybe we could combine the tests, but I'd rather not add duplicate expects.

src/core/Expander/Expander/Expander.tsx
```diff
@@ -0,0 +1,226 @@
...
+}
+
+export interface ExpanderProps extends InternalExpanderProps, TokensProp {}
+
+export interface ExpanderTitleBaseProps {
+  /** Custom classname to extend or customize */
+  className?: string;
+  /** Expander consumer for open state and toggle open callback */
+  consumer: ExpanderProviderState;
+}
```
ketsappi, line 71: Should this interface be here instead
`suomifi-ui-components/src/core/Expander/ExpanderTitle/ExpanderTitle.tsx` ?

src/core/Expander/Expander/Expander.test.tsx
aappoalander: Will remove.

src/core/Expander/Expander/Expander.tsx
```diff
@@ -0,0 +1,226 @@
...
+
+export interface ExpanderContentBaseProps {
+  /** Custom classname to extend or customize */
+  className?: string;
+  /** Expander consumer for open state */
+  consumer: ExpanderProviderState;
+}
```
ketsappi, line 78: And this one to be in
`suomifi-ui-components/src/core/Expander/ExpanderContent/ExpanderContent.tsx`?

src/core/Expander/Expander/Expander.tsx
aappoalander: I have no preference for either solution in this case since there are so few props that everything fits nicely into one view with both ways and the consts are only used once. The proposed solution has fewer lines so will change.

src/core/Expander/Expander/Expander.tsx
aappoalander: In my opinion no. It is the constraint for the expander interface and is required to be implemented by all title components. E.g. ExpanderTitleButton and future ExpanderTitle need to implement / extend this interface (in addition to some implementation details) to work.

src/core/Expander/Expander/Expander.tsx
aappoalander: See above comment reply.

src/core/Expander/Expander/Expander.tsx
aappoalander: Expanders are actually closed by default and this can change it. Would be good to come up with better wording though... "Expanders initial open state" with current @default false?

ketsappi: Yes, they are closed by default and tried to describe the situation where the boolean answers it when reading it
> Expander is opened by default: false

 as the false is default value in this case.

But the suggested is better:
> Expanders initial open state

src/core/Expander/Expander/Expander.tsx
ketsappi: Then it is good to have it here. 

By reviewing the PR, it is hard to know all the future use cases and therefore hard to see reason why it is **now** here. As we try to prevent making features and such that are _maybe_ used in the future.

But as discussed, that you will soon in separate PR create those other possible title components, then this is much clearer.

src/core/Expander/Expander/Expander.test.tsx
```diff
@@ -0,0 +1,205 @@
...
+  };
+
+  const TestExpander = TestExpanderWithProps({
+    ariaCloseText: 'click to close expander',
+    ariaOpenText: 'click to open expander',
+    children: 'Test expander',
+  });
+
+  it('render with the same component on the same container does not remount', () => {
+    const { getByTestId, rerender } = render(TestExpander);
```
ketsappi, line 31: Is this `TestExpander` really useful as it just used once? 
Could it be directly in the render as done with the rerender?

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
```diff
@@ -0,0 +1,189 @@
...
+const openClassName = `${baseClassName}--open`;
+const expandersContainerClassName = `${baseClassName}_expanders`;
+const openAllButtonClassName = `${baseClassName}_all-button`;
+interface InternalExpanderGroupProps {
+  /** Expanders and option other components */
+  children: ReactNode;
+  /** 'Open all' button text */
+  OpenAllText: string;
+  /** 'Close all'-component (Button) */
+  CloseAllText: string;
```
ketsappi, line 20: Should the JSDoc have same logic?

> 'Open all' button text
> 'Close all' button text

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
```diff
@@ -0,0 +1,189 @@
...
+  /** Open/Close all button props */
+  toggleAllButtonProps?: Omit<
+    HtmlButtonProps,
+    | 'onClick'
+    | 'onMouseDown'
+    | 'onMouseUp'
+    | 'onKeyPress'
+    | 'onKeyUp'
+    | 'onKeyDown'
+  >;
```
ketsappi, line 36: Would it be more clear if this is extracted outside of the interface?

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+interface ExpanderGroupState {
+  /** Expanders that are open */
+  expanders: ExpanderOpenStates;
+  /** State change transition request */
+  expanderGroupOpenState: ExpanderGroupTargetOpenState;
+}
+
+export interface ExpanderGroupProviderState {
+  onExpanderOpenChange: (id: string, toState: boolean | undefined) => void;
```
ketsappi, line 55: Maybe using the `targetState` here also, instead of `toState`?

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
```diff
@@ -0,0 +1,189 @@
...
+      delete expanders[id];
+      return { expanders };
+    });
+  };
+
+  expandersOpenState = () => {
+    const expanderCount = Object.keys(this.state.expanders).length;
+    const openExpanderCount = Object.values(this.state.expanders).filter(
+      (value) => value === true,
+    ).length;
```
ketsappi, line 97: Would it be possible to use
`(value) => !!value` or `(value) => value` instead?

Or would it cause some implications?

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: This is no changed to targetOpenState along with the type name and other interface name changes.

jest.setup.js
aappoalander: There should be some kind of description now. 😅

src/core/Expander/Expander/Expander.baseStyles.tsx
aappoalander: Removed

src/core/Expander/Expander/Expander.test.tsx
aappoalander: Removed

src/core/Expander/Expander/Expander.tsx
aappoalander: Changed to use id directly since needed only in one place per id.

src/core/Expander/Expander/Expander.tsx
aappoalander: Changed it to "Initial open state" since repeating Expander didn't feel necessary.

src/core/Expander/Expander/Expander.tsx
aappoalander: Yes, this is exported now, but used only in one place. This will change shortly after this PR gets merged. Wasn't sure about the pattern my self, but could not come up with anything better since it really is a constraint for the children of Expander.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: Fixed. Original wording was left there by mistake.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: It does not matter documentation wise (tried it out) and this type is not used elsewhere. Moving it out from the interface increases the amount of code by a little and doesn't provide any additional value in this case, at least in my opinion.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: Should not have any implications as values are booleans here. Changed it to the latter proposed option. It is shorter and a bit neater, but also more difficult to read as it requires one to know or check the type of value (expanders).

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: Changed it, but it should actually be "newState" instead of toState. newState was already used in handleExpanderOpenChange function and reflects the use better in my opinion, because it's not a request to change the state, but rather just informing that the child expander state has already changed.

jest.setup.js
ketsappi: I think those comments are good addition to clear things up 👍

src/core/Expander/Expander/Expander.baseStyles.tsx
ketsappi: It seems that it got removed from other baseStyles, but were still left in this file.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
ketsappi: Then let's leave it as it is now.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
ketsappi: Then if that is the case, should the `value` be more describing? 
Like using the same `targetOpenState` to explicitly tell what it is.

src/core/Expander/Expander/Expander.baseStyles.tsx
aappoalander: So it seems. Probably did one too many undos while testing out something. Will fix shortly.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: What about `isOpen`? It would maybe better describe what the value actually is. The value has no name in the interface (`[key: string]: boolean`).

ketsappi: That is better or maybe even shorter, `open.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
```diff
@@ -0,0 +1,189 @@
...
+import { baseStyles } from './ExpanderGroup.baseStyles';
+import { VisuallyHidden } from '../../../components';
+
+const baseClassName = 'fi-expander-group';
+const openClassName = `${baseClassName}--open`;
+const expandersContainerClassName = `${baseClassName}_expanders`;
+const openAllButtonClassName = `${baseClassName}_all-button`;
+interface InternalExpanderGroupProps {
+  /** Expanders and option other components */
+  children: ReactNode;
```
ketsappi, line 16: > Expanders and **_option_** other components

Was this wanted to be, `optionally`?

src/core/Expander/ExpanderContent/ExpanderContent.tsx
```diff
@@ -0,0 +1,92 @@
...
+const baseClassName = 'fi-expander';
+const contentBaseClassName = `${baseClassName}_content`;
+const contentOpenClassName = `${contentBaseClassName}--open`;
+const noPaddingClassName = `${contentBaseClassName}--no-padding`;
+
+export interface ExpanderContentProps extends Omit<HtmlDivProps, 'id'> {
+  /** Content for expander */
+  children: ReactNode;
+  /** Remove padding from expandable content area (for background usage with padding in given container etc.) */
+  noPadding?: boolean;
```
ketsappi, line 22: Maybe adding the `@default false` to see the value in the styleguidist.

src/core/Expander/Expander/Expander.tsx
```diff
@@ -0,0 +1,225 @@
...
+const {
+  Provider: ExpanderProvider,
+  Consumer: ExpanderConsumer,
+} = React.createContext(defaultProviderValue);
+
+interface InternalExpanderProps {
+  /**
+   * Children, extend type ExpanderTitleBaseProps or ExpanderContentBaseProps
+   * ExpanderProviderState context is used to communicate between title, content and expander
+   */
```
ketsappi, line 44: **Not to change anything regarding this PR**

Just to add note for future, if it is ever going to be supported in JSDoc, support for e.g `@see`.
So that these comments that will refer to these props/interfaces are not just text, but actual "link".
Something like:

```
Children, extend type {@see ExpanderTitleBaseProps} or {@see ExpanderContentBaseProps}
```

Open issue for this.
https://github.com/microsoft/TypeScript/issues/35524

src/core/Expander/Expander/Expander.baseStyles.tsx
aappoalander: Fixed.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: Changed it to `isOpen` as it suggests a boolean value.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
aappoalander: Changed it to `Expanders and optionally, other ReactNodes`

src/core/Expander/ExpanderContent/ExpanderContent.tsx
aappoalander: Added and also added `!!` for the condition to explicitly change undefined to false when not provided.

src/core/Expander/Expander/Expander.tsx
aappoalander: That's a good idea, maybe we should test it with styleguidist and apply where necessary. Parts of our API documentation could use improvements.

src/core/Expander/ExpanderGroup/ExpanderGroup.tsx
```diff
@@ -0,0 +1,189 @@
...
+  [key: string]: boolean;
+}
+
+type ExpanderGroupTargetOpenState = {
+  targetOpenState: boolean;
+};
+
+interface ExpanderGroupState {
+  /** Expanders that are open */
+  expanders: ExpanderOpenStates;
```
ketsappi, line 49: Just a small thing, but wasn't it so that this will always hold the state of the expanders not matter closed or opened?
If so, then the comment might be little misguiding.

aappoalander: Changed to `Expanders by id's with current open state`

ketsappi (approved):

LJKaski (approved):

PR #415, opened 2020-12-10
Feature/expander title checkbox

aappoalander:
## Description

This PR introduces ExpanderTitle as a variant for ExpanderTitleButton. ExpanderTitle only has open / close button on the far right of the title to support focusable content on the remaining title area. This makes it possible to use the expander with common use cases like checkbox in the expander title.

In addition, most of the Expander related tests are rearranged or refactored so that each component now have separate test files.

## Motivation and Context

There was a clear need to support focusable content on the expander title and the variant with whole title are as button did not work with these use cases.

## How Has This Been Tested?

Tested with Styleguidist, CRA TypeScript, SSR application by using VoiceOver with Chrome, Firefox and Safari and by using NVDA with Chrome, Firefox, Edge and IE.

Screenshot 2020-12-10 at 7 49 47

## Release notes
- ExpanderTitle component to support interactive content on expander titles.

### DIFF
--- a/src/core/Expander/Expander/Expander.md
+++ b/src/core/Expander/Expander/Expander.md
@@ -20,7 +20,10 @@ import {
 } from 'suomifi-ui-components';
 
 <Expander className="expander-test" defaultOpen={true}>
-  <ExpanderTitle>
+  <ExpanderTitle
+    ariaOpenText="open expander"
+    ariaCloseText="close expander"
+  >
     <Checkbox hintText="Checkbox hint text">Checkbox label</Checkbox>
   </ExpanderTitle>
   <ExpanderContent>Test expander</ExpanderContent>
--- a/src/core/Expander/Expander/Expander.md
+++ b/src/core/Expander/Expander/Expander.md
@@ -6,7 +6,9 @@ import {
 } from 'suomifi-ui-components';
 
 <Expander className="expander-test" defaultOpen={true}>
-  <ExpanderTitleButton>Test expander</ExpanderTitleButton>
+  <ExpanderTitleButton asHeading="h3">
+    Test expander
+  </ExpanderTitleButton>
   <ExpanderContent>Test expander</ExpanderContent>
 </Expander>;
 ```
--- a/src/core/Expander/Expander/Expander.md
+++ b/src/core/Expander/Expander/Expander.md
@@ -14,6 +14,22 @@ import {
 ```jsx
 import {
   Expander,
+  ExpanderTitle,
+  ExpanderContent,
+  Checkbox
+} from 'suomifi-ui-components';
+
+<Expander className="expander-test" defaultOpen={true}>
+  <ExpanderTitle>
+    <Checkbox hintText="Checkbox hint text">Checkbox label</Checkbox>
+  </ExpanderTitle>
+  <ExpanderContent>Test expander</ExpanderContent>
+</Expander>;
+```
+
+```jsx
+import {
+  Expander,
   ExpanderGroup,
   ExpanderTitleButton,
   ExpanderContent
@@ -59,16 +75,7 @@ const [expanderThreeOpen, setExpanderThreeOpen] = React.useState(
   false
 );
 
-const [showExpander, setShowExpander] = React.useState(false);
-
 <>
-  <button
-    onClick={() => {
-      setShowExpander(!showExpander);
-    }}
-  >
-    {showExpander ? 'Hide' : 'Show'}
-  </button>
   <ExpanderGroup
     OpenAllText="Open all"
     AriaOpenAllText="Open all expanders"
@@ -83,12 +90,6 @@ const [showExpander, setShowExpander] = React.useState(false);
       <ExpanderTitleButton>Test expander 2</ExpanderTitleButton>
       <ExpanderContent>Test expander content 2</ExpanderContent>
     </Expander>
-    {!!showExpander && (
-      <Expander>
-        <ExpanderTitleButton>Test expander X</ExpanderTitleButton>
-        <ExpanderContent>Test expander content X</ExpanderContent>
-      </Expander>
-    )}
     <Expander
       open={expanderThreeOpen}
       onOpenChange={(open) => {
--- a/src/core/Expander/ExpanderTitle/ExpanderTitle.tsx
+++ b/src/core/Expander/ExpanderTitle/ExpanderTitle.tsx
@@ -1,0 +1,112 @@
+import React, { Component, ReactNode } from 'react';
+import { default as styled } from 'styled-components';
+import classnames from 'classnames';
+import { withSuomifiDefaultProps } from '../../theme/utils';
+import { TokensProp, InternalTokensProp } from '../../theme';
+import {
+  HtmlDiv,
+  HtmlButton,
+  HtmlButtonProps,
+  HtmlDivProps,
+} from '../../../reset';
+import { expanderTitleBaseStyles } from './ExpanderTitle.baseStyles';
+import { Icon } from '../../Icon/Icon';
+import { ExpanderConsumer, ExpanderTitleBaseProps } from '../Expander/Expander';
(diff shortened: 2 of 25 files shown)


src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.tsx
```diff
@@ -4,19 +4,21 @@ import classnames from 'classnames';
...
 import { HtmlDiv, HtmlButton, HtmlButtonProps } from '../../../reset';
-import { expanderTitleButtonBaseStyles } from './ExpanderTitle.baseStyles';
+import { expanderTitleButtonBaseStyles } from './ExpanderTitleButton.baseStyles';
 import { Icon } from '../../Icon/Icon';
 import { ExpanderConsumer, ExpanderTitleBaseProps } from '../Expander/Expander';
 import { VisuallyHidden } from '../../../components';
 
 const baseClassName = 'fi-expander';
-const titleClassName = `${baseClassName}_title`;
+const titleClassName = `${baseClassName}_title-button`;
```
ketsappi, line 13: Could the `titleClassName` to be the `baseClassName` instead. If the `baseClassName` is used to refer to the this component in this file.

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+      providerProps: {
+        ...providerProps,
+        open: false,
+      },
+    });
+    const titleDiv = getByTestId('expander-open-by-default-title');
+    expect(titleDiv).toHaveClass('fi-expander_title-button');
+    expect(titleDiv).not.toHaveClass('fi-expander_title-button--open');
+    expect(buttonToClick.querySelector('svg')).not.toHaveClass(
+      'fi-expander_title-button-icon--open',
```
ketsappi, line 198: In general, should we have `--closed` kind of styles? E.g `fi-expander_title-button--closed`
Then we do not have to rely on `not.toHaveClass` and then it is also explicitly told.

src/core/Expander/ExpanderTitle/ExpanderTitle.tsx
```diff
@@ -0,0 +1,112 @@
...
+    | 'onKeyPress'
+    | 'onKeyUp'
+    | 'onKeyDown'
+  >;
+}
+
+export interface InternalExpanderTitleProps
+  extends ExpanderTitleProps,
+    ExpanderTitleBaseProps {}
+
```
ketsappi, line 48: Are there need to export these internal props? (_same with ExpanderTitleButton_)

src/core/Expander/Expander/Expander.md
```diff
@@ -11,6 +11,22 @@ import {
...
+```jsx
+import {
+  Expander,
+  ExpanderTitle,
+  ExpanderContent,
+  Checkbox
+} from 'suomifi-ui-components';
+
+<Expander className="expander-test" defaultOpen={true}>
+  <ExpanderTitle>
```
ketsappi, line 23: As the `ariaCloseText` and `ariaOpenText` props are required, it might be good idea to add them to the example.

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.tsx
aappoalander: Probably should be, will test to ensure there are no side-effects.

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.test.tsx
aappoalander: That's one way to do it. Would be a change to current pattern which is used in multiple places. Having it that way would be more verbose though and make things a bit more readable. Wouldn't change it just for the tests since expects are virtually free - setting up the tests seems to take most of the time consumed.

src/core/Expander/ExpanderTitle/ExpanderTitle.tsx
aappoalander: Probably not, those were reused originally and were probably left there by accident. Will check and remove is possible.

src/core/Expander/Expander/Expander.md
aappoalander: 😅  Why bother?

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.test.tsx
ketsappi: Not just for the tests, but if it would also help otherwise for users to give custom style based on that 🤔 .
And as `not.toHaveClass` are so easily mistaken as discussed earlier, therefore could also help there.

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+        ...renderOptions
+      }: { providerProps: ExpanderProviderState },
+    ) =>
+      customRender(ui, {
+        providerProps,
+        container: rendered.container,
+        ...renderOptions,
+      }),
+  };
+};
```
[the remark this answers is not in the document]
aappoalander: This is now fixed. Tried to keep the call signature unchanged, because rerender should take same props with actual customRender.

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.test.tsx
aappoalander: I wouldn't change this logic for the time being. Maybe there is a need for further discussion with the team if we need to change the overall pattern for this.

src/core/Expander/ExpanderTitle/ExpanderTitle.tsx
aappoalander: Exports removed.

src/core/Expander/Expander/Expander.md
aappoalander: Fixed.

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.tsx
aappoalander: Fixed.

src/core/Expander/Expander/Expander.md
aappoalander: Note that these props are only mandatory for ExpanderTitle, not ExpanderTitleButton.

ketsappi (approved):

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.tsx
```diff
@@ -0,0 +1,118 @@
...
+    } = this.props;
+
+    return (
+      <HtmlDiv
+        {...passProps}
+        className={classnames(className, baseClassName, {
+          [titleOpenClassName]: !!consumer.open,
+        })}
+      >
+        <HtmlSpan {...(!!asHeading ? { as: asHeading } : {})}>
```
ketsappi, line 67: Is it necessary to wrap to `span` when `asHeading` is not told?
Or was there some other benefit by doing like this?

ketsappi:
Looking good.
Comprehensive tests added which is nice 🚀

src/core/Expander/ExpanderTitleButton/ExpanderTitleButton.tsx
aappoalander: Not necessary, but would keep the wrapping structure uniform if done this way. I prefer not to use conditional wrapping structures with complex structures as the logic would be either in two places (both opening and closing tags) or the contents would have to be defined as const and have two completely separate blocks.

LJKaski (approved):
Looks good to me. Good work! :)

ketsappi (approved):

PR #417, opened 2020-12-14
[Feature] Debounce component

LJKaski:
## Description
This PR introduces an internal component that provides a debounce functionality to its children. This PR also implements said component to SearchInput and TextInput to provide users with a simple debounce functionality and related API.

## Motivation and Context
When working with asynchronous code, e.g. when fetching search results from the server, it is often necessary to use debounce to avoid sending the fetch request on every single keystroke. Using debounce will put less strain on the server and help apps and services to run smoother.

## How Has This Been Tested?
Tested manually in styleguidist and with automated tests.

## Release notes
-Debounce functionality in TextInput and SearchInput


src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -228,43 +231,57 @@ class BaseSearchInput extends Component<SearchInputProps> {
...
-          <HtmlButton {...searchButtonDerivedProps}>
-            <VisuallyHidden>{searchButtonLabel}</VisuallyHidden>
-            <Icon
-              aria-hidden={true}
-              icon="search"
-              className={searchInputClassNames.searchIcon}
-            />
-          </HtmlButton>
-        </HtmlDiv>
+        <Debounce waitFor={this.props.debounce || 0}>
```
ketsappi, line 234: Wasn't there discussion that there is no need to tell the time as zero, as undefined would be handled by the Debounce util?

So the use will be like:
`<Debounce waitFor={this.props.debounce}>`

aappoalander: Truthy check should be sufficient if used in Debounce component. !!waitFor should cover all cases where waitFor is 0, "", undefined or null => not present and can be handled the same way.

src/core/utils/Debounce/Debounce.tsx
```diff
@@ -0,0 +1,49 @@
...
+    this.timeout = setTimeout(() => {
+      callback(value);
+    }, waitFor);
+  };
+
+  cancelDebounce = () => {
+    this.clearTimeout();
+  };
+
+  clearTimeout() {
```
ketsappi, line 34: Would it possible to have different name for this method?
It is now little confusing to read, when the `clearTimeout` is called and it will call the JS "internal" one and not this.

So `this.clearTimeout` and `clearTimeout` are too close to each other when looking this.

src/core/Form/TextInput/TextInput.md
```diff
@@ -101,5 +101,11 @@ import { TextInput } from 'suomifi-ui-components';
     icon="mapLocation"
     iconProps={{ fill: 'red' }}
   />
+  <TextInput
+    onBlur={(event) => console.log(event.target.value)}
```
ketsappi, line 105: I would remove the `onBlur` to not distract what is done with debounce.

src/core/Form/TextInput/TextInput.test.tsx
```diff
@@ -257,4 +257,71 @@ describe('props', () => {
...
+        <TextInput labelText="Test input" icon="close" />,
+      );
+      const icon = container.querySelector('.fi-icon');
+      expect(container.contains(icon)).toBeTruthy();
+    });
+  });
+
+  describe('debounce', () => {
+    it('delays the running of onChange by the given time', () => {
+      jest.useFakeTimers();
```
ketsappi, line 293: This fake timer is great find 👍   
Didn't know it existed before this PR.

src/core/Form/SearchInput/SearchInput.test.tsx
```diff
@@ -199,6 +199,46 @@ describe('props', () => {
...
+          clearButtonLabel="clear"
+          searchButtonLabel="search"
+          data-testid="debounced-search"
+          debounce={1000}
+          onChange={mockOnChange}
+        />
+      );
+      const { getByTestId } = render(searchInput);
+
+      const inputElement = getByTestId('debounced-search') as HTMLInputElement;
```
aappoalander, line 218: no need to change this, but getByRole, getByText etc. should be preferred over test id and other arbitrary queries. See https://testing-library.com/docs/guide-which-query

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: In addition, I would either put Debounce as high up in the component hierarchy as possible (e.g. wrap the exported SearchInput) or right next to the component it is used for, in this case wrap just the functionalityContainer div since components outside it have no use for the functionality.

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -151,9 +154,6 @@ class BaseSearchInput extends Component<SearchInputProps> {
       if (!('value' in this.props)) {
```
aappoalander, line 154: I think debounce should not be applied in controlled state because this leads to weird behaviour of not updating the input value immediately. Not sure if we should tackle this or leave it up for the teams using this?

src/core/Form/TextInput/TextInput.md
```diff
@@ -101,5 +101,11 @@ import { TextInput } from 'suomifi-ui-components';
...
+    labelText="TextInput with debounced onChange event"
+    onChange={(value) => console.log(value)}
```
aappoalander, line 107: Should we have both onBlur and onChange applied to the same example?

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -68,10 +71,14 @@ export interface TextInputProps
...
   name?: string;
+  /** Controlled value */
+  value?: TextInputValue;
   /** Set components width to 100% */
   fullWidth?: boolean;
   /** Text to mark a field optional. Will be wrapped in parentheses and shown after labelText. */
   optionalText?: string;
+  /** Debounce time in milliseconds for onChange function. No debounce is applied if no value is given. */
+  debounce?: number;
   icon?: BaseIconKeys;
```
aappoalander, line 82: Not related to this PR, but why is the icon here if it is not even documented?

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -151,15 +159,25 @@ class BaseTextInput extends Component<TextInputProps> {
...
+              <HtmlInput
+                {...passProps}
+                id={this.id}
+                className={textInputClassNames.inputElement}
+                type={type}
+                placeholder={visualPlaceholder}
+                {...{ 'aria-invalid': status === 'error' }}
+                {...getDescribedBy()}
+                onChange={(event: ChangeEvent<HTMLInputElement>) => {
+                  const eventValue = event.currentTarget.value;
```
aappoalander, line 173: this could actually be inside the if statement so when there is no propOnChange the const never gets created.

src/core/utils/Debounce/Debounce.test.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { render, fireEvent } from '@testing-library/react';
+import { Debounce } from '../Debounce/Debounce';
+import { HtmlInput } from '../../../reset/HtmlInput/HtmlInput';
+
+describe('debounce', () => {
+  it('runs the children function wrapped by it', () => {
+    const mockFunction = jest.fn(() => 12);
+    render(
+      <Debounce data-testid="abc" waitFor={1000}>
+        {mockFunction}
```
aappoalander, line 11: This looks like it should always work regardless of context => useless test. The mockFunction is immediately called on render, so the wrapping it inside debounce does nothing. Rather put it so that some event triggers a callback that uses the debounce without delay, e.g. like below but without the waitFor.

aappoalander:
Looks very good overall, well done! 💪

src/core/Form/SearchInput/SearchInput.tsx
LJKaski: I think that's something that can be left to the project teams. Might be good to mention it in the prop description though.

src/core/Form/SearchInput/SearchInput.tsx
```diff
@@ -228,43 +231,57 @@ class BaseSearchInput extends Component<SearchInputProps> {
...
+                  }}
+                  onKeyPress={onKeyPress}
+                  onKeyDown={onKeyDown}
+                />
+              </HtmlDiv>
+              <HtmlButton
+                {...clearButtonProps}
+                onClick={() => {
+                  onClear();
+                  cancelDebounce();
```
ketsappi, line 264: Now when user presses clear button the debounce is then cancelled and the `onChange` is called immediately even after the user continues to write. So possibly doing some unnecessary logic to search from API etc if it is not taken care by the user.

Maybe we should tell this in documentation to library user?

src/core/Form/SearchInput/SearchInput.tsx
LJKaski: Ah good points. I'll fix the waitFor value handling. At this point in the year everything that's not immediately written down will be forgotten 😅. What comes to the placement, to me it seems it's in the right spot. Wrapping the functionalityContainer, that is.

ketsappi:
As we are making changes to TextInput and tests, should we change these to be inline as they do not give any benefit actually _(well, only the first is used twice)_.


https://github.com/vrk-kpa/suomifi-ui-components/blob/4ee654b800c88f1d96cf043d986f613295bb4737/src/core/Form/TextInput/TextInput.test.tsx#L7-L30

src/core/utils/Debounce/Debounce.tsx
LJKaski: Changed it to cancelDebounce, which also allowed me to delete the extra method passed as a render prop.

src/core/Form/TextInput/TextInput.test.tsx
LJKaski: Thanks. Took some digging to find a way to test this 😄

src/core/Form/TextInput/TextInput.tsx
LJKaski: Probably just an oversight from earlier. I'll add it in this PR since it is such a small thing.

src/core/Form/TextInput/TextInput.tsx
LJKaski: Good point. Moved it inside.

src/core/Form/TextInput/TextInput.md
LJKaski: I moved it to the icon example above.

src/core/utils/Debounce/Debounce.tsx
LJKaski: I take that back somewhat. Losing the middle man method somehow seemed to break things, so I just renamed the internal method to cancelTimeout.

aappoalander: Good catch, should be changed in my opinion too!

src/core/Form/SearchInput/SearchInput.tsx
ketsappi: As discussed separately the consensus was that let's leave this as it is now.

src/core/Form/TextInput/TextInput.md
LJKaski: Moved the onBlur to another example.

src/core/Form/SearchInput/SearchInput.test.tsx
LJKaski: I've now changed all the tests that can plausibly use getByRole to use that.

src/core/utils/Debounce/Debounce.test.tsx
LJKaski: Replaced this with a test with no waitFor time.

src/core/Form/SearchInput/SearchInput.test.tsx
aappoalander: Great work!

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: True that, probably better not to touch it, but mention in the docs somehow.

src/core/Form/SearchInput/SearchInput.test.tsx
```diff
@@ -199,6 +199,45 @@ describe('props', () => {
...
+      const { getByRole } = render(searchInput);
+
+      const inputElement = getByRole('searchbox') as HTMLInputElement;
+      fireEvent.change(inputElement, { target: { value: 'new value' } });
+      expect(mockOnChange).not.toBeCalled();
+      jest.advanceTimersByTime(1000);
+      expect(mockOnChange).toBeCalledTimes(1);
+      expect(inputElement.value).toBe('new value');
+    });
+    it('resolves right when no onChange is given', () => {
```
aappoalander, line 224: I'd rather test that it works as expected when no debounce is given, would be more common scenario.

src/core/Form/SearchInput/SearchInput.tsx
aappoalander: Or just leave it out, it will be obvious when testing it. One would have to be ignorant of controlled pattern altogether if using with debounce.

src/core/Form/TextInput/TextInput.test.tsx
```diff
@@ -142,30 +135,22 @@ describe('props', () => {
...
-          labelText="Test input"
-          type="number"
-          data-testid="number-input"
-        />
-      );
-      const { getByTestId } = render(textInput);
-      const numberfield = getByTestId('number-input') as HTMLInputElement;
+      const textInput = <TextInput labelText="Number input" type="number" />;
+      const { getByRole } = render(textInput);
+      const numberfield = getByRole('spinbutton') as HTMLInputElement;
```
aappoalander, line 153: This was new to me as well. Not obvious that an input type number is a spinbutton...

src/core/Form/SearchInput/SearchInput.test.tsx
aappoalander: Ok, seems like there is already such a test written before this issue, so no need to duplicate it. Renaming this test would still be desirable in my opinion.

src/core/utils/Debounce/Debounce.test.tsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+import { Debounce } from '../Debounce/Debounce';
+import { HtmlInput } from '../../../reset/HtmlInput/HtmlInput';
+
+describe('debounce', () => {
+  it('runs given function immediately when no debounce time is specified', () => {
+    const mockFunction = jest.fn(() => 12);
```
ketsappi, line 8: It seems that it could be also like for the test to work:
`const mockFunction = jest.fn();`

Was there previously some more things to be done with that magic `12`?

LJKaski: Ah, forgot that. it was used in an earlier version of the test but forgot to remove that part. Removed the magic 12 :)

src/core/utils/Debounce/Debounce.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  render() {
+    // casted to any due to https://github.com/Microsoft/TypeScript/issues/10727
+    const { children } = this.props as any;
+    if (!!children) {
+      return children(this.debouncer, this.cancelDebounce);
+    }
+    return null;
+  }
```
ketsappi, line 48: One option is to cast the children to be the correct type; not as compact as currently but possible:

```jsx
render() {
    const children = this.props.children as (
      debouncer: Function,
      cancelDebounce: Function,
    ) => JSX.Element;

    if (!!children) {
      return children(this.debouncer, this.cancelDebounce);
    }
    return null;
  }
```

src/core/Form/SearchInput/SearchInput.test.tsx
```diff
@@ -199,6 +199,45 @@ describe('props', () => {
...
+          debounce={1000}
+          onChange={mockOnChange}
+        />
+      );
+      const { getByRole } = render(searchInput);
+
+      const inputElement = getByRole('searchbox') as HTMLInputElement;
+      fireEvent.change(inputElement, { target: { value: 'new value' } });
+      expect(mockOnChange).not.toBeCalled();
+      jest.advanceTimersByTime(1000);
```
ketsappi, line 220: _Just a small idea maybe not reason to change._
As the test name implies we want to make sure that `onChange` is delayed by a given time, we could check if that is really the case. As now we are advancing the time only once and straight to wanted.
But will it call `onChange` already with lower delay?

It could be replaced like this:
```js
      jest.advanceTimersByTime(500);
      expect(mockOnChange).not.toBeCalled();
      jest.advanceTimersByTime(500);
```

Then we know that by delaying only 500, it was not run yet.

aappoalander: Don't know the inner workings of the advanceTimersByTime, but doing the above might be problematic since we are doing an expect in between two timers and that might lead into situation where expect is run actually later than 500ms after the change event. This might depend on the environment as well. If advanceTimers is smart, this might work well too.

src/core/utils/Debounce/Debounce.tsx
aappoalander: Correct type is better if it's possible since it gives better support in IDE.

LJKaski: Added the suggested typings.

src/core/Form/SearchInput/SearchInput.test.tsx
LJKaski: It would be logical that the fake timers are only advanced when called to do so. That'd support testing with the extra step in between. Seems to work that way, so I'll add it.

src/core/Form/SearchInput/SearchInput.test.tsx
LJKaski: Removed it from here and added a similar one to Debounce's own tests.

ketsappi (dismissed):

src/core/Form/SearchInput/SearchInput.test.tsx
aappoalander: The issue I was expecting was about setTimeOut which we cannot fake and should resolve on its own regardless of fake timers. As further testing proved, in this case there is no such issue and the code is synchronous so this can be left as is.

aappoalander (approved):

ketsappi (approved):

PR #420, opened 2020-12-28
[Feature] Button soft disabled

aappoalander:
## Description
This PR adds support for aria-disabled attribute and allows focusing disabled buttons while navigating with tab-button. In addition, this PR removes unstyled variant and support for Aria live. Components level implementation is also removed as it became redundant.

## Motivation and Context
For example, Submit buttons should be completely focusable when disabled to better communicate reasons behind disabled state for screen reader users. Aria-assertive / live does not work well for these use cases since it announces changes in the disabled state immediately even if focus has never reached the button it self. A better strategy for this is to use aria-describedby with aria-disabled. This way all relevant information is always available for screen reader users as well.

## How Has This Been Tested?
Tested with Styleguidist, CRA-TS and SSR test app using Chrome, Firefox, Edge, Safari with VoiceOver on MacOS.

## Release notes
- Breaking changes ⚒️ : support for Aria-disabled, unstyled variant removed along with mouseNoFocus, ariaAssertiveDisabled and ariaAssertiveUnDisabled attributes


ketsappi:
Maybe add a note to PR description that Button is now removed from the components level.

src/core/Button/Button.md
```diff
@@ -16,6 +16,24 @@ import { Button } from 'suomifi-ui-components';
...
+<>
+  <Button
+    aria-describedby="additiona-info"
+    aria-disabled={true}
+    onClick={() => console.log('Test button click')}
+    type="submit"
+  >
+    Aria-disabled button
+  </Button>
+  <Paragraph id="additiona-info">
```
ketsappi, line 31: Not a problem, but as this is a example we might want to fix the typo in the id;
From _additiona-info_ to _additiona**l**-info_

aappoalander: I'd also like to apply a style here, but no comments from design yet.

src/core/Button/Button.test.tsx
```diff
@@ -1,44 +1,84 @@
...
-    <Button data-testid="button" disabled={true}>
-      Test
-    </Button>,
-  );
-  const { getByTestId, rerender } = button;
-  expect(getByTestId('button')).toBeDisabled();
+describe('Disabled Button', () => {
+  it('is disabled and changes to enabled', () => {
+    const button = render(<Button disabled={true}>Test</Button>);
+    const { getByRole, rerender } = button;
```
ketsappi, line 32: Could this and other tests be with "one line"?

```tsx
    const { getByRole, rerender } = render( <Button disabled={true}>Test</Button>);
```

src/core/Button/Button.tsx
```diff
@@ -1,34 +1,44 @@
...
-import { UnstyledButton } from './UnstyledButton';
 
 type ButtonVariant =
   | 'default'
   | 'inverted'
   | 'secondary'
   | 'secondary-noborder'
-  | 'tertiary'
-  | 'unstyled';
-
```
ketsappi, line 21: Just for readability, might be good to leave the empty row between `type` and `export interface` here.

src/core/Button/Button.tsx
```diff
@@ -41,102 +51,111 @@ export interface ButtonProps extends CompButtonProps, TokensProp {
...
+      children,
       ...passProps
     } = this.props;
-    const { variant } = passProps;
     const { className: iconPropsClassName, ...passIconProps } = iconProps;
-
-    if (variant === 'unstyled') {
-      return <UnstyledButton {...passProps} />;
-    }
+    const onClickProps = !!disabled || !!ariaDisabled ? {} : { onClick };
```
ketsappi, line 84: Maybe we could use singular, as there is only one property, --> `const onClickProp`

But not sure which is best, as this object is then spread later when used for button.

aappoalander: Prop should be sufficient for this case I guess

src/core/Button/Button.md
aappoalander: Typo fixed and text for additional-info improved.

src/core/Button/Button.test.tsx
aappoalander: Fixed. Also helped with quite a few naming issues, no name required for those any more.

src/core/Button/Button.tsx
aappoalander: Fixed.

src/core/Button/Button.tsx
aappoalander: Changed.

ketsappi (approved):

src/core/Button/Button.md
aappoalander: As agreed, styling this will be handled later on when PAMU team is again available.

PR #423, opened 2021-01-05
[Fix] Aria attribute issues

aappoalander:
## Description
This PR introduces fixes for aria attribute handling for multiple components. Empty aria attributes are now omitted completely and updating aria attributes are always present in DOM. This PR also includes a minor update to aria utils.

Some components were not updated due to bigger upcoming migration. These issues were documented to related tickets.

## Motivation and Context
Empty or appearing aria attributes may not work as intended and needed to be fixed.

## How Has This Been Tested?
Tested with Styleguidist and Create React App using Mac OS and Safari, Chrome and Firefox.

## Release notes
- Fixes and improvements for aria attribute handling.

### DIFF
--- a/src/utils/aria.ts
+++ b/src/utils/aria.ts
@@ -12,32 +12,30 @@ export const ariaLabelOrHidden = (ariaLabel?: string) =>
 export const ariaFocusableNoLabel = (ariaLabel?: string) =>
   ifAriaNoLabel(ariaLabel) ? {} : { focusable: false };
 
+type ariaPropName = 'aria-describedby' | 'aria-labelledby' | 'aria-label';
+type ariaProp = { [key in ariaPropName]: string } | {};
 /**
- * Returns object with 'aria-describedby' which can be spread to props.
+ * Returns object with 'aria-' property which can be spread to props.
  * E.g:
  * @example
  * <Component
- *   {...getAriaDescribedBy([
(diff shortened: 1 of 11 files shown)


src/utils/aria.ts
```diff
@@ -12,32 +12,30 @@ export const ariaLabelOrHidden = (ariaLabel?: string) =>
...
 /**
- * Returns object with 'aria-describedby' which can be spread to props.
+ * Returns object with 'aria-' property which can be spread to props.
  * E.g:
  * @example
  * <Component
- *   {...getAriaDescribedBy([
+ *   {...getConditionalAriaProp('aria-describedby', [
  *      describingElement ? "id-of-describing-element" : undefined,
  *      otherDescribingElement ? "id-of-other-describing-element" : undefined,
```
ketsappi, line 24: Maybe we could update the example to have the so called normal case also, as it is not required to tell the "undefined" version when using, but it is possible though.

e.g 
```tsx
{...getConditionalAriaProp('aria-label', [ariaLabel])}
```

aappoalander: Good point, changed the original example to have both, a static and a conditional id.

ketsappi (approved):

LJKaski (approved):
Seems to work as intended and looks good to me.

PR #425, opened 2021-01-08
[Feature] FilterInput (internal)

ketsappi:
## Description

- FilterInput to be used e.g Combobox
- Filters the given set of items based on text written to input

## Motivation and Context

This is going to be a part of the Combobox and in future maybe used by other components also. 
Good to have already a separate PR for this to tackle possible issues already etc.

## Example code

```tsx
interface Product {
  name: string;
  price: number;
  tax: boolean;
}

const tools: Product[] = [
  { name: 'Jackhammer', price: 230, tax: false },
  { name: 'Hammer', price: 15, tax: true },
  { name: 'Sledgehammer', price: 36, tax: false },
  { name: 'Spade', price: 50, tax: true },
  { name: 'Powersaw', price: 150, tax: false },
];

const filter = (tool: Product, query: string) => tool.name.includes(query);

<FilterInput
  labelText="Label"
  items={tools}
  onFiltering={(filtered) => console.log(filtered)}
  filterRule={filter}
/>

```

## Release notes

_No need to add to release notes as this is internal component to be used by other components._


src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+  // eslint-disable-next-line react/prop-types
+  tokens,
+  // eslint-disable-next-line react/prop-types
+  id: propId,
+  ...passProps
+}) => (
+  <AutoId id={propId}>
+    {(id) => <BaseFilterInput id={id} {...passProps} />}
+  </AutoId>
+);
```
ketsappi, line 171: - `FilterInputWithoutTokens` is added to get types working with styled-components. Might be possible to combine it  with `StyledFilterInput` (down below) but I couldn't figure it out 😅 
- Better naming ideas are welcome.
- eslint disables are added as it warns from missing prop-types
  - did left them both to have own disable line, so that future additions changes are not affected (vs. using prop-type disable for whole file)

src/core/Form/FilterInput/FilterInput.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+          items={tools}
+          onFiltering={(filtered) => console.log(filtered)}
+          filterRule={filter}
+        />,
+      );
+      const inputField = getByRole('textbox') as HTMLInputElement;
+      expect(inputField).toHaveAttribute('placeholder', 'Enter text here');
+    });
+  });
+});
```
ketsappi, line 199: Currently there are no tests for following props: `items`, `onFiltering` and` filterRule`.

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+        className={classnames(baseClassName, className, {
+          [filterInputClassNames.disabled]: !!passProps.disabled,
+          [filterInputClassNames.error]: status === 'error',
+        })}
+      >
+        <HtmlDiv
+          className={classnames(filterInputClassNames.wrapper, {
+            [filterInputClassNames.labelAlignLeft]: labelAlign === 'left',
+          })}
+        >
```
ketsappi, line 121: This is the additional wrapper what we have discussed. It tries to keep the outside container to not affect the component layout (e.g using `display`).

src/core/Form/FilterInput/FilterInput.baseStyles.tsx
```diff
@@ -0,0 +1,77 @@
+import { css } from 'styled-components';
+import { withSuomifiTheme, TokensAndTheme } from '../../theme';
+import { input, containerIEFocus, font } from '../../theme/reset';
+import { absoluteFocus } from '../../theme/utils';
+
+export const baseStyles = withSuomifiTheme(
+  ({ theme }: TokensAndTheme) => css`
+    & .fi-filter-input {
+      ${font({ theme })('bodyText')}
+      display: inline-block;
```
aappoalander, line 10: what happens if fi-filter-input is the only place with display set and someone gives the outermost div a style with display: flex? Should the display value be set deeper in the structure, let's say in the wrapper?

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+import {
+  HtmlInput,
+  HtmlInputProps,
+  HtmlDiv,
+  HtmlDivProps,
+} from '../../../reset';
+import { TokensProp, InternalTokensProp } from '../../theme';
+import { baseStyles } from './FilterInput.baseStyles';
+import classnames from 'classnames';
+import { Omit } from '../../../utils/typescript';
```
aappoalander, line 13: I think we should stop using the custom Omit for TS, since I guess we anyways require TS 3.8 or later.

aappoalander: And I'd prefer having the import ordered to some extent, e.g. React, external libs, utlis styles and resets, other comps, component styles and so on.

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+  error: `${baseClassName}--error`,
+  disabled: `${baseClassName}--disabled`,
+  labelAlignLeft: `${baseClassName}--label-align-left`,
+  wrapper: `${baseClassName}_wrapper`,
+  functionalityContainer: `${baseClassName}_functionalityContainer`,
+  inputElementContainer: `${baseClassName}_input-element-container`,
+  inputElement: `${baseClassName}_input`,
+};
+
+type FilterInputStatus = Exclude<InputStatus, 'success'>;
```
aappoalander, line 31: Is Exclude use intentionally or should Omit be used here as well?

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+   */
+  status?: FilterInputStatus;
+  /** Status text to be shown below the component and hint text. Use e.g. for validation error */
+  statusText?: string;
+  /** FilterInput name */
+  name?: string;
+  /** Align label on top or on the left side of the input field
+   * @default 'top'
+   */
+  labelAlign?: 'top' | 'left';
```
aappoalander, line 64: I'd prefer this the other way around alignLabel, what do you @LJKaski think?

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+  /** Items to be filtered */
+  items: Array<T>;
+  /** Returns the filtered items */
+  onFiltering: (filteredItems: Array<T>) => void;
```
aappoalander, line 68: What about onFilterChange, onChangeFilter, onFilter or even onChange simply? The -ing suffix sounds like this is called while filtering is happening e.g. to show a loading spinner.

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+  /** Filtering rule to be used */
+  filterRule: (item: T, query: string) => boolean;
```
aappoalander, line 70: And also for this, I'd prefer something that tells the user this is a function / callback rather than e.g. a reg.exp. rule. What about filterFunction or filterFunc? The functionality of these callbacks needs to be communicated through naming in my opinion.

aappoalander: maybe call these onFilter and filterFunc, that would be clear for me at least. And it's different from onChange which could also refer to changes in the search term / input value.

src/core/Form/FilterInput/FilterInput.tsx
aappoalander: Things that need to remain unchanged e.g. display, possibly max-width etc. should probably be specified for the wrapper rather than the input container?

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -0,0 +1,186 @@
...
+                {...getConditionalAriaProp('aria-describedby', [
+                  statusTextId,
+                  ariaDescribedBy,
+                ])}
+                autoComplete="off"
+                aria-autocomplete="list"
+                auto-capitalize="false"
+                spellCheck="false"
+                onChange={onChangeHandler}
+              />
```
aappoalander, line 147: This might benefit from aria role searchbox or listbox etc. Might require testing with screenreaders to determine which works the best. Had to add one for searchinput to make it work correctly at least.

src/core/Form/FilterInput/FilterInput.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+  { name: 'Powersaw', price: 150, tax: false },
+];
+
+const filter = (tool: Product, query: string) => tool.name.includes(query);
+
+const TestFilterInput = (
+  <FilterInput
+    labelText="Label"
+    items={tools}
+    onFiltering={(filtered) => console.log(filtered)}
```
aappoalander, line 27: Console logs might introduce interesting side-effects when run with CI, wouldn't use here if not absolutely necessary. Rather use a spy or mock etc?

src/core/Form/FilterInput/FilterInput.tsx
ketsappi: > And I'd prefer having the import ordered to some extent, e.g. React, external libs, utlis styles and resets, other comps, component styles and so on.

Tried quickly to look if there would be some nice automatic stuff to do this, but didn't find one.

src/core/Form/FilterInput/FilterInput.test.tsx
aappoalander: https://jestjs.io/docs/en/mock-function-api#mockfnmockresults

aappoalander: Maybe this could or something similar could be used to test the filter functionality as well?

src/core/Form/FilterInput/FilterInput.test.tsx
```diff
@@ -0,0 +1,199 @@
...
+      );
+      expect(container.firstChild).toHaveClass('fi-filter-input--disabled');
+
+      const inputField = getByRole('textbox') as HTMLInputElement;
+      expect(inputField).toHaveAttribute('disabled');
+    });
+  });
+
+  describe('labelText', () => {
+    it('should be found ', () => {
```
aappoalander, line 124: should there be a test for labelAlign as well? Maybe not necessary, but just thinking.

aappoalander:
Looks pretty solid overall! Left quite a few comments, but those were mostly about API naming and minor style things. Good work!

src/core/Form/FilterInput/FilterInput.tsx
ketsappi: These both are now changed and commits pushed.

aappoalander: There are tools to handle import order, but I think those use only external - internal and alphabetic order. Could be useful still.

src/core/Form/FilterInput/FilterInput.tsx
LJKaski: Could be either way in my opinion. This one is consistent with CSS rules like text-align, but alignLabel would be grammatically more correct. All the other label props start with 'label', so I'd probably keep it as it is. labelAlignment would be cleaner grammatically, but maybe unnecessarily long.

src/core/Form/FilterInput/FilterInput.tsx
LJKaski: onFilter would be better in my opinion. It would have to be 'uponFiltering' if using the '-ing' ending, so onFilter would be a better option here.

src/core/Form/FilterInput/FilterInput.tsx
ketsappi: I do not think it works correctly with `Omit`.

- [Omit takes an object type and removes the given properties from the type.](https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys)
- [Exclude takes a union type and removes given items of that union.](https://www.typescriptlang.org/docs/handbook/utility-types.html#excludetype-excludedunion)

src/core/Form/FilterInput/FilterInput.tsx
LJKaski: My thoughts as well. Is Func a commonly used abbreviation when it comes to these?  I'd maybe just use 'filterFunction' to be explicit.

src/core/Form/FilterInput/FilterInput.tsx
ketsappi: I think that current `labelAlign` is better as it in line with `labelMode` and `labelText`  as LJKaski mentioned.

aappoalander: I agree with comments above! Lets keep the original version!

src/core/Form/FilterInput/FilterInput.tsx
ketsappi: This is now changed to be `onFilter`

src/core/Form/FilterInput/FilterInput.tsx
ketsappi: Changed `filterRule` to `filterFunc` as discussed.

src/core/Form/FilterInput/FilterInput.test.tsx
ketsappi: Yes, there should be.

src/core/Form/FilterInput/FilterInput.tsx
ketsappi: `display` and `width` is now moved to be told for the wrapper in baseStyles.

src/core/Form/FilterInput/FilterInput.test.tsx
ketsappi: Test is now added for `labelAlign` also.

src/core/Form/FilterInput/FilterInput.baseStyles.tsx
ketsappi: This is now done in wrapper.

src/core/Form/FilterInput/FilterInput.test.tsx
ketsappi: `console.log`s are now removed and using mock functions. Also using empty array for `items`-prop and mock function for the `filterFunc`.

There is now test for the `onFilter` prop. It basically tests also the `items` and `filterFunc` props as it can not work without them really.

src/core/Form/FilterInput/FilterInput.tsx
ketsappi: Probably the `role="listbox"` would be the closest one. But currently it causes confusion as screenreader then tells that user could start interacting with the items.

I suggest that we add this or some other role to **FilterInput** when the **Combobox** is builded to a state where it uses FilterInput component and is actually able to show the items in popover.

FilterInput will highly likely to see more changes also when implementing Combobox.

src/core/Form/FilterInput/FilterInput.test.tsx
```diff
@@ -0,0 +1,244 @@
...
+          labelText="Label"
+          visualPlaceholder="Enter text here"
+          items={tools}
+          onFilter={mockedOnFilter}
+          filterFunc={filter}
+        />,
+      );
+      const inputField = getByRole('textbox') as HTMLInputElement;
+      fireEvent.change(inputField, { target: { value: 'hammer' } });
+      expect(mockedOnFilter.mock.results).toEqual([
```
aappoalander, line 233: So it works! 💯

src/core/Form/FilterInput/FilterInput.test.tsx
aappoalander: Looks like there is now a test for all of the above. Good to resolve this in my opinion. Or at least the mock.results test looks like it.

ketsappi: Yes, these do now have tests.

aappoalander (approved):

PR #426, opened 2021-01-11
[Bugfix] Box sizing inheritance

LJKaski:
## Description
This PR fixes a couple of reported issues where global styles caused inherited box-size CSS rules to cause some issues with radiobutton and external link. External link icon size was also wrong. That was fixed in the process.

## Motivation and Context
This was a reported bug, which, while it was caused by a global rule in a different project, was easier to fix on our end.

## How Has This Been Tested?
Tested in styleguidist in chrome.

## Release notes
-External link and RadioButton pseudo elements now have explicitly set box-size.

### DIFF
--- a/src/core/Block/__snapshots__/Block.test.tsx.snap
+++ b/src/core/Block/__snapshots__/Block.test.tsx.snap
@@ -25,6 +25,11 @@ exports[`calling render with the same component on the same container does not r
   white-space: normal;
 }
 
+.c0::before,
+.c0::after {
+  box-sizing: border-box;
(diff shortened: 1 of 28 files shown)


aappoalander:
What about the box-sizing for pseudo elements, especially focus? Should those be handled with this branch as well?

ketsappi (approved):

ketsappi (dismissed):

ketsappi (approved):

aappoalander (approved):

PR #428, opened 2021-01-18
[Feature] Move radio button to core

aappoalander:
## Description
This PR Removes component level implementation for RadioButton and switches completely to Context pattern. Support for using variants trough statics is removed. In addition, AutoId is now used instead of uuid.

Also some minor fixes for aria-prop handling were done, but this has no effect on functionality.

A new problem from existing implementation was also found, and it proved to be too difficult to fix within the scope of this issue. This problem concerns only Firefox + VoiceOver combination on Mac OS and does not prevent using the radio buttons. Navigating with only arrow keys is not working properly when VoiceOver is turned on. Root cause seems to be a Mozilla Firefox bug, which was also reported: https://bugzilla.mozilla.org/show_bug.cgi?id=1687257

## Motivation and Context
This change improves the internal implementation, aligns the component better with rest of the library and fixes some SSR related issues.

## How Has This Been Tested?
Tested with Styleguidist, CRA-TS app and SSR app using VoiceOver with Safari/Chrome/Firefox and using NVDA with Chrome/Firefox/Edge/IE

## Release notes
Breaking Changes: 
- variants are no longer supported using statics. Use `variant="large"` instead of `Radiobutton.large`.
- Label prop is renamed to LabelText
- Custom styles targetting component internal structure may no longer work


src/core/Form/RadioButton/RadioButton.tsx
```diff
@@ -1,74 +1,193 @@
...
+  input: `${baseClassName}_input`,
+  label: `${baseClassName}_label`,
+  hintText: `${baseClassName}_hintText`,
   disabled: `${baseClassName}--disabled`,
   large: `${baseClassName}--large`,
   checked: `${baseClassName}--checked`,
 };
 
-export interface RadioButtonProps extends CompRadioButtonProps, TokensProp {}
+type RadioButtonVariant = 'small' | 'large';
```
ketsappi, line 24: Is there need to export this or why this is a separate type? 
As earlier discussed with previous PR's we thought that it union type could just be in the interface if there is no certain need to have it out.

src/core/Form/RadioButton/RadioButton.tsx
```diff
@@ -1,74 +1,193 @@
...
+                onChange={(event: React.ChangeEvent<HTMLInputElement>) => {
+                  if (!!consumer.onRadioButtonChange) {
+                    consumer.onRadioButtonChange(event.target.value);
+                  }
+                  if (!!onChange) {
+                    onChange(event);
+                  }
+                }}
+              />
+            )}
```
ketsappi, line 187: Props from `consumer` could be also deconstucted here, but I think the current way is clearer what goes where.
I think this can be leaved as it is.

src/core/Form/RadioButton/RadioButtonGroup.tsx
```diff
@@ -1,23 +1,152 @@
...
+const radioButtonGroupClassNames = {
+  container: `${baseClassName}_container`,
+  label: `${baseClassName}_label`,
+  hintText: `${baseClassName}_hintText`,
+};
 
-export interface RadioButtonGroupProps
-  extends CompRadioButtonGroupProps,
-    TokensProp {}
+type Label = 'hidden' | 'visible';
```
ketsappi, line 19: Same as [other comment](https://github.com/vrk-kpa/suomifi-ui-components/pull/428/files#r560752173)

src/core/Form/RadioButton/RadioButtonGroup.tsx
```diff
@@ -1,23 +1,152 @@
...
+
+export interface RadioButtonGroupProps extends TokensProp {
+  /** Custom classname to extend or customize */ className?: string;
+  /** RadioButton or ReactNode */
+  children: Array<React.ReactElement<RadioButtonProps> | ReactNode>;
+  /** Hint text to be displayed under the label. */
+  hintText?: string;
+  /** Label for the group */
+  label: string;
```
ketsappi, line 28: Should the `label` prop to be `labelText` instead to be more consistent with other components?

src/core/Form/RadioButton/RadioButtonGroup.tsx
```diff
@@ -1,23 +1,152 @@
...
+  render() {
+    const {
+      children,
+      className,
+      label,
+      labelMode,
+      hintText,
+      id,
+      name,
+      defaultValue: dismissDefaultValue,
```
ketsappi, line 100: As discussed earlier in separate issue, should we get rid of this `dismissDefaultValue` and just leave it as `defaultValue` ?

ketsappi:
Noticed that the styleguidist will not show props of the RadiobuttonGroup (possibly haven't shown them before this PR either) .
But would be nice to see them to help with the use of the component.

src/core/Form/RadioButton/RadioButton.tsx
aappoalander: Might be better to just move it to the interface so there would be no need to export it.

src/core/Form/RadioButton/RadioButton.tsx
aappoalander: I don't like this part so much, but could not find a better way without polluting the underlying components with external state / consumer.

src/core/Form/RadioButton/RadioButtonGroup.tsx
aappoalander: Would be a breaking change, but I guess we could change it since there are already other breaking changes.

src/core/Form/RadioButton/RadioButtonGroup.tsx
aappoalander: 'discard' would be better wording, but I'm kind of torn between leaving it as is and not explicitly stating that this value is not used here.

src/core/Form/RadioButton/RadioButton.tsx
aappoalander: Changed

src/core/Form/RadioButton/RadioButton.tsx
aappoalander: It's now using destructuring and a bit cleaner in my opinion. Not clean by any means, but would probably require additional tooling to make it better.

src/core/Form/RadioButton/RadioButtonGroup.tsx
aappoalander: Changed as well

src/core/Form/RadioButton/RadioButtonGroup.tsx
aappoalander: Changed

src/core/Form/RadioButton/RadioButtonGroup.tsx
aappoalander: Removed the rename completely.

aappoalander:
> Noticed that the styleguidist will not show props of the RadiobuttonGroup (possibly haven't shown them before this PR either) .
> But would be nice to see them to help with the use of the component.

This should be now fixed and there are also some improved examples. The order of components is not alphabetical due to different import structure and I'm wondering if we should simplify the imports and change everything to manual order?

src/core/Form/RadioButton/RadioButton.tsx
ketsappi: Just tested that we might be able to shorten even more,
```tsx
{...(!!onRadioButtonChange
  ? {
      checked: selectedValue === passProps.value,
      name,
    }
  : {})}
```
to be like this:
```tsx
{...(!!onRadioButtonChange && {
  checked: selectedValue === passProps.value,
  name,
})}
```

Didn't test it very thoroughly yet, but what do you think about this?

aappoalander: It might work, but not sure what this would transpile into. We would be more or less spreading undefined but not inside of an object. Needs further investigation. https://stackoverflow.com/questions/46957194/javascript-es6-spread-operator-on-undefined

src/core/Form/RadioButton/RadioButtonGroup.tsx
aappoalander: Finally fixed. Mixed this up with similar issue in RadioButton.

src/core/Form/RadioButton/RadioButton.tsx
ketsappi: Maybe not worth the risk to change this.

src/core/Form/RadioButton/RadioButton.md
```diff
@@ -25,90 +64,102 @@ const handleChange = (event) => {
   setSelectedValue(event.target.value);
```
LJKaski, line 64: This isn't used in this example anymore it seems. So should either be removed or have onChange use this one.

aappoalander: Good catch, will fix!

aappoalander:
> > Noticed that the styleguidist will not show props of the RadiobuttonGroup (possibly haven't shown them before this PR either) .
> > But would be nice to see them to help with the use of the component.
> 
> This should be now fixed and there are also some improved examples. The order of components is not alphabetical due to different import structure and I'm wondering if we should simplify the imports and change everything to manual order?

Looks like this cannot be fixed easily since Styleguidist does not support mixing simple components and sections. Will leave as is for now.

src/core/Form/RadioButton/RadioButton.md
aappoalander: Fixed.

ketsappi (approved):

LJKaski (approved):

PR #429, opened 2021-01-20
[Breaking change] Move Dropdown to core and implement AutoId

LJKaski:
## Description
This PR removes Dropdown component from the components level and moves all the necessary logic to core level. In addition, uuid is replaced by AutoId, dropdown item static is replaced by a separate DropdownItem component and custom label is replaced by the internal LabelText component.

## Motivation and Context
The components level is no longer maintained so for the sake of development efficiency it needs to be removed from the components. AutoId helps the components perform better in SSR environments.

## How Has This Been Tested?
Tested with automated tests (only snapshots changed) and by running locally in styleguidist.

## Release notes
* [Breaking change] Dropdown to use cleaner composition


src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -108,7 +95,7 @@ export interface DropdownProps {
   onChange?(newValue: string): void;
 }
 
-export class Dropdown extends Component<DropdownProps> {
+export class BaseDropdown extends Component<DropdownProps> {
```
ketsappi, line 98: Maybe we do not want to export the `BaseDropdown`, only the one Dropdown at the very bottom?
Now styleguidist has the tree like this on left side menu (and also mentioning the BaseDropdown on the start go the page):

```
Dropdown
  BaseDropdown
  DropdownItem
```

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -129,37 +116,22 @@ export class Dropdown extends Component<DropdownProps> {
...
     const {
-      id: propId,
+      id,
       name,
       disabled,
       children,
-      labelProps,
       labelText,
-      labelTextProps,
-      labelMode = 'top',
```
ketsappi, line 154: It seems that **Dropdown** has also the version of having label before the input.
With filterInput we introduced the `labelAlign` prop for that:
https://github.com/vrk-kpa/suomifi-ui-components/blob/3a6e4a804d368f8f2c8f27ba38a480d179d00f8a/src/core/Form/FilterInput/FilterInput.tsx#L63

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -76,15 +69,10 @@ export interface DropdownProps {
...
-  labelTextProps?: ParagraphProps;
-  /** Label displaymode -
-   * top: show above, hidden: use only for screenreader
-   * @default top
-   */
-  labelMode?: DropdownLabel;
+  /** Label display mode */
+  labelMode?: LabelMode;
+  /** Text to mark a field optional. Will be wrapped in parentheses and shown after labelText. */
+  optionalText?: string;
```
ketsappi, line 75: In design, **Dropdown** also has the status and status text.

We may consider to add them in this PR or soon, in separate PR.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -241,37 +195,85 @@ export class Dropdown extends Component<DropdownProps> {
...
+    </AutoId>
+  ),
+)`
+  ${(props) => baseStyles(props)}
+`;
+
+interface NewListboxPopoverProps extends BaseListboxPopoverProps, TokensProp {}
+
+const StyledListboxPopover = styled(
+  ({ tokens, children, ...passProps }: NewListboxPopoverProps) => (
```
ketsappi, line 256: Maybe we do not need the `NewListboxPopoverProps` as it does not introduce anything.

Instead of:
```
const StyledListboxPopover = styled(
    ({ tokens, children, ...passProps }: NewListboxPopoverProps) => (
```

We could use it without it, like this:
```
const StyledListboxPopover = styled(
    ({ tokens, children, ...passProps }: BaseListboxPopoverProps & TokensProp) => (
```

LJKaski: Good point. I think that was from the old implementation. Changed it (not pushed yet).

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Dropdown previously had only top and hidden as options here so it made sense to switch to visible and hidden in order to make use of the LabelText component. If there's need for more options I think it might make sense to do that separately from this, especially if there are more things to change such as the status text.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Thanks for spotting that. Just a forgotten relic from moving stuff from components I think. Now removed.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -1,21 +1,23 @@
...
 } from '@reach/listbox';
+import { withSuomifiDefaultProps } from '../../theme/utils';
+import { TokensProp, InternalTokensProp } from '../../theme';
+import { baseStyles, listboxPopoverStyles } from './Dropdown.baseStyles';
+import { DropdownItemProps } from '../DropdownItem/DropdownItem';
+import { HtmlSpan } from '../../../reset';
+import { LabelText, LabelMode } from '../../Form/LabelText/LabelText';
+import { logger } from '../../../utils/logger';
+import { AutoId } from '../../../utils/AutoId';
 import { positionMatchWidth } from '@reach/popover';
```
aappoalander, line 20: Imports should be ordered so that external imports are introduced first, then e.g. utils / resets, other components and then finally imports related directly to this component.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -27,18 +29,12 @@ export const dropdownClassNames = {
...
-
-export interface DropdownItemProps {
-  /** Item value */
-  value: string;
-  /** Item content */
-  children: ReactNode;
-  /** Classname for item */
-  className?: string;
-}
+type BaseListboxPopoverProps = ListboxPopoverProps & {
```
aappoalander, line 35: Tested these quickly and looks like all Reach related custom types can be removed except OptionalListboxButtonProps. Instead of using [K in key... syntax, Partial<ListboxButtonProps> seems to do the job as well. The refs also seem to be unnecessary and at least styleguidist works without issues if removed.

I would also consider the API at this point, do we really want to support replacing the listbox component or all the related props? Maybe just allow id and className etc. or remove completely?

aappoalander: Specifically dropdownButtonProps, dropdownPopoverProps and listboxPopoverComponent. I would consider removing those since we are anyways introducing breaking changes.

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: These proposals make sense since we are introducing breaking changes. Maybe we should introduce a quick fix to master with the style fixes required for focus box-sizing? This way we could finalize the components level removal and bring all breaking changes to components at once. Or at least we should try to avoid introducing breaking changes to same components too often.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -241,37 +195,85 @@ export class Dropdown extends Component<DropdownProps> {
...
+ * <i class="semantics" />
+ * Use for selectable dropdown with items.
+ */
+export class Dropdown extends Component<DropdownProps> {
+  render() {
+    const props = withSuomifiDefaultProps(this.props);
+    return (
+      <StyledDropdown
+        {...props}
+        listboxPopoverComponent={StyledListboxPopover}
```
aappoalander, line 275: This abstraction is probably no longer necessary since we only support the version with suomifi styles? So we could just add the styles in the same place with StyledDropdown?

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: Left a related comment below as well. Could we just use the styled versions of these and drop the abstraction completely?

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: This could probably be dropped altogether? And should actually work with Partial<ListboxPopoverProps> & TokensProp or something similar.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -241,37 +195,85 @@ export class Dropdown extends Component<DropdownProps> {
         id={id}
         {...passProps}
       >
-        <HtmlLabel
+        <LabelText
```
aappoalander, line 198: I would consider adding a wrapper between the outermost span and label / other contents. So styles including display would be set on the new wrapper, but classnames, id and passprops left as is. This would allow setting margins, display: flex and so on for Dropdown without adding a wrapper or breaking styles.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Seems to work that way at a glance. Will change this 👍

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Good idea. Added a simple div as a wrapper.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Cleaned these up quite a bit. A good point about them being mostly obsolete.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -241,37 +177,60 @@ export class Dropdown extends Component<DropdownProps> {
...
   }
 }
+
+const StyledDropdown = styled(
+  ({
+    tokens,
+    id: propId,
+    ...passProps
+  }: DropdownProps & InternalTokensProp) => (
+    <AutoId id={propId}>
```
aappoalander, line 219: Could we move this to top level Dropdown? This way styled would only handle styles and the provider would not have any effect on the components below => a bit cleaner solution overall.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -241,37 +177,60 @@ export class Dropdown extends Component<DropdownProps> {
...
+            {labelText}
+          </LabelText>
+          <ListboxInput {...listboxInputProps}>
+            <ListboxButton
+              {...dropdownButtonProps}
+              id={buttonId}
+              className={classnames(dropdownClassNames.button, {
+                [dropdownClassNames.disabled]: !!disabled,
+              })}
+              {...buttonAriaLabelledByOverride}
```
aappoalander, line 196: This could be done using ariaConditional prop util with a single liner. I'd prefer to have it here since there are quite a few derived props above the template.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: It's now gone.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -241,37 +177,60 @@ export class Dropdown extends Component<DropdownProps> {
...
+        <HtmlDiv>
+          <LabelText
+            id={labelId}
+            labelMode={labelMode}
+            as="label"
+            optionalText={optionalText}
+          >
+            {labelText}
+          </LabelText>
+          <ListboxInput {...listboxInputProps}>
```
ketsappi, line 189: Could these props be told right here I the template for ListBoxInput, as this is the only place they are used?

https://github.com/vrk-kpa/suomifi-ui-components/blob/91b2a1fbc8770b5771e3a16cc211f1a412d90802/src/core/Dropdown/Dropdown/Dropdown.tsx#L158-L164

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: Vote for separate PR.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Should now be in a more logical order.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Good point. Moved it there.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Yes they could. And now they are :)

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Could it, since the parameter defining if the aria prop should be there is not the aria prop itself?

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: As agreed, this PR will not be included in 3.0 release, so handling these in separate PR's is completely fine. Following tickets should however be created before resolving these comments.

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander:
```
{...getConditionalAriaProp('aria-labelledby', [
            !!selectedValue ? ariaLabelledByIds : undefined,
          ])}
```
or even
{...getConditionalAriaProp('aria-labelledby', [
            !!selectedValue ? ariaLabelledBy : undefined,
            !!selectedValue ? labelId : undefined,
          ])}

LJKaski: Not quite a single liner, but now in use. At least it's now in a more logical place in the code.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Actually there were some repercussions from that, and as I took a closer look, it is in the styled variant in all the other implementations as well. Moved it back there in this one as well.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Ticket created for adding status and status text.

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: It is there for other implementations with the exception of RadioButton. It was moved there to have a cleaner implementations. There shoud be no obstacles for doing so here as well.

src/core/Dropdown/Dropdown/Dropdown.tsx
ketsappi: Seems weird that both; ListboxInput and ListboxButton have same aria-labelledby ids. But it does work with the screenreader when tested.

ketsappi (approved):

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Created a ticket about this.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Now on top level.

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: Probably would work even with one of those removed. Can both button and input be focused when navigation with modifier + arrow keys?

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -241,37 +164,66 @@ export class Dropdown extends Component<DropdownProps> {
...
+          >
+            <ListboxButton
+              {...dropdownButtonProps}
+              id={buttonId}
+              className={classnames(dropdownClassNames.button, {
+                [dropdownClassNames.disabled]: !!disabled,
+              })}
+              {...getConditionalAriaProp('aria-labelledby', [
+                selectedValue === undefined ? ariaLabelledByIds : undefined,
+              ])}
```
ketsappi, line 191: I think we could also have it like this and not using the const with string interpolation:

```tsx
{...getConditionalAriaProp(
  'aria-labelledby',
  selectedValue === undefined ? [labelId, ariaLabelledBy] : [],
)}
```

Would this be a little more readable?

aappoalander: The first could also probably be `selectedValue === undefined ? [labelId, ariaLabelledBy] : undefined,` and if selectedValue is always either truthy or not truthy (how do we indicate no selection yet?) then `!!selectedValue ? [labelId, ariaLabelledBy] : undefined`

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: Actually, tested the previous version with `{{display: 'inline', margin: '50px', position: 'absolute', top: '100px', left: '300px'}}` and nothing seems to be broken. It might work with new version as well even without the wrapping div. There seems to be no reset styles for the original wrapper (If there were, those styles should now be applied to the new wrapping div to prevent an unintended override), so the additional div might be redundant.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Yeah, maybe it's better to rely on the util in all cases. I'd stick with the `selectedValue === undefined` to be on the safe side since the values can vary.

src/core/Dropdown/Dropdown/Dropdown.tsx
LJKaski: Both are needed. If the button doesn't have a labelledBy, the visualPlaceholder gets read instead of the label if a visualPlaceholder is used.

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: Should be ok as discussed.

src/core/Dropdown/Dropdown/Dropdown.tsx
aappoalander: Should be ok without truthy checking as there can be empty strings in some cases.

ketsappi (approved):

aappoalander (approved):

PR #432, opened 2021-01-27
[Feature] SearchInput auto id

aappoalander:
## Description
This PR adds AutoId for SearchInput and replaces the old uuid implementation. By doing so SSR support is improved and the id mismatch between server and client no longer happens.

In addition, this PR introduces an intermediate style wrapper so that it is now possible to add margin/display etc styles directly to SearchInput without breaking the component.

## Motivation and Context
SSR Should be fully supported to allow faster page loads and hydration on client side. Positioning the components should be possible without the danger of breaking the internal layout and styles and this can be achieved by adding an additional style wrapper inside the component.

## How Has This Been Tested?
Tested with Mac OS and VoiceOver using Firefox, Safari and Chrome and with Win10 and NVDA using Chrome, Firefox and Edge.

## Release notes
**SearchInput**
- Add better support for SSR by replacing uuid with AutoId
- Add support for styling SearchInput directly without breaking the component's internal layout
- Breaking change: rename inputContainerProps to wrapperProps

### DIFF
--- a/src/core/Form/SearchInput/SearchInput.baseStyles.tsx
+++ b/src/core/Form/SearchInput/SearchInput.baseStyles.tsx
@@ -7,11 +7,16 @@ export const baseStyles = withSuomifiTheme(
   ({ theme }: TokensAndTheme) => css`
     &.fi-search-input {
       ${font({ theme })('bodyText')}
-      display: inline-block;
       width: 290px;
     }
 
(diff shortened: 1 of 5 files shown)


ketsappi (approved):

LJKaski (approved):
Good work :)

PR #434, opened 2021-02-02
[Feature] Move toggle to core

aappoalander:
## Description
This PR removes Toggle components level implementation and introduces various improvements and changes to the component:
- Toggle now uses AutoId
- Toggle input and button variants are now separate components with shared base
- Static variants are no longer supported
- Wrapper div added for better styling support.

## Motivation and Context
Toggle button had unstyled version which was not working properly and added unnecessary maintenance burden. In addition, use of uuid prevented SSR from working properly. Also, the use cases for these components, as well as the APIs are somewhat different, so having button and input as variants of the same component makes no sense.

## How Has This Been Tested?
Tested with Mac OS VoiceOver and NVDA + Windows using Chrome, Firefox, Safari and Edge. Tested using a CRA test project, styleguidist and SSR test project.

## Release notes
**Toggle**
- Breaking Changes:
  - Changed:
    - ToggleButton onClick parameter type is now boolean instead of { toggleState: boolean }
  - Removed:
    - variant support. <Toggle.withInput> or <Toggle variant="withInput"> no longer work. Use ToggleButton and ToggleInput components instead.
    - toggleInputProps
    - toggleInputComponent
    - onClick from ToggleInput, use onChange instead
    - Additional props are now passed for button and input elements.
 - Added support for wrapperProps to support styling better.
 - Added support for AutoId to support Server Side Rendering


src/core/Form/Toggle/ToggleBase/ToggleBase.tsx
```diff
@@ -0,0 +1,37 @@
...
+   */
+  'aria-label'?: string;
+  /**
+   * alternatively to aria-label you can define aria-labelledby with label-element id
+   */
+  'aria-labelledby'?: string;
+  /** Unique id
+   * @default uuidV4
+   */
+  id?: string;
```
ketsappi, line 32: Maybe this should be updated to be like:

```JS
  /**
   * Unique id
   * If no id is specified, one will be generated automatically
   */
  id?: string;
```

src/core/Form/Toggle/ToggleBase/ToggleBase.tsx
```diff
@@ -0,0 +1,37 @@
...
+  /** Disable usage */
+  disabled?: boolean;
+  /**
+   * Label element content
+   */
+  children?: ReactNode;
+  /**
+   * aria-label for the HTML input-element,
+   */
+  'aria-label'?: string;
```
ketsappi, line 24: We should update the JSDoc as now it says that `aria-label` is for input but it is used also for the **button** version.

src/core/Form/Toggle/ToggleButton/ToggleButton.tsx
```diff
@@ -0,0 +1,140 @@
...
+  label: `${baseClassName}_label`,
+  iconContainer: `${baseClassName}_icon-container`,
+};
+
+export interface ToggleButtonProps
+  extends ToggleBaseProps,
+    Omit<HtmlButtonProps, 'onClick'> {
+  /** Event handler to execute when clicked */
+  onClick?: ({ toggleState }: { toggleState: boolean }) => void;
+}
```
ketsappi, line 26: Should we also have **soft disable** for ToggleButton as we have for Button?

aappoalander: I don't see it as necessary for this as with the normal button. For normal button it was added to mainly support e.g. submit buttons when there is mandatory functionality that can be disabled due to errors. If toggle is disabled it should not be an issue for screen readers to not show it at all. Or maybe not, might add some value as well.

src/core/Form/Toggle/ToggleBase/ToggleBase.tsx
aappoalander: Changed it to "for toggle" instead. The other is passed on to button and the other on to input. Not sure if these should be there at all, as there is the "label" as ReactNode already.

src/core/Form/Toggle/ToggleBase/ToggleBase.tsx
aappoalander: Changed this to match other components using AutoId

src/core/Form/Toggle/ToggleButton/ToggleButton.tsx
aappoalander: Will leave this as future improvement in case needed.

src/core/Form/Toggle/ToggleBase/Toggle.baseStyles.tsx
```diff
@@ -0,0 +1,36 @@
+import { css } from 'styled-components';
+import { TokensAndTheme } from '../../../theme';
+import { disabledCursor } from '../../../../components/utils/css';
+import { element, font } from '../../../theme/reset';
+
+// Contains double underscore because it is written in the SVG-file
```
ketsappi, line 6: This seems to be some old comment that could be removed?

ketsappi (dismissed):

ketsappi (approved):

LJKaski (approved):

PR #435, opened 2021-02-05
[BreakingChange] Text input AutoId, aria util and styling

LJKaski:
## Description
This PR implements AutoId, aria util, style wrapper and fixes non-neutral status border width. 

## Motivation and Context
Style wrapper safeguards against alignment issues and ensures the integrity of the component. AutoId helps with server side rendering by ensuring the ID on the server and client side match. Internal aria util ensures the conditional aria props are put in place in a clean manner. Border was previously 1px wide for all states, but in designs it is 2px for the error and success statuses. 

## How Has This Been Tested?
Tested by running in styleguidist and by automated tests.

## Release notes
Text Input
* **Breaking change**: inputContainerProps renamed to wrapperProps
* Error and success status border width now 2px

### DIFF
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -77,15 +77,17 @@ export const baseStyles = withSuomifiTheme(
     &.fi-text-input--error {
       & .fi-text-input_input {
         border: 2px solid ${theme.colors.alertBase};
-        height: 38px;
         padding-left: ${math(`${theme.spacing.insetL} - 1`)};
+        padding-top: 7px;
+        padding-bottom: 7px;
       }
     }
     &.fi-text-input--success {
       & .fi-text-input_input {
         border: 2px solid ${theme.colors.successBase};
-        height: 38px;
         padding-left: ${math(`${theme.spacing.insetL} - 1`)};
+        padding-top: 7px;
+        padding-bottom: 7px;
       }
     }
     &.fi-text-input--disabled {
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -81,7 +81,7 @@ export const baseStyles = withSuomifiTheme(
     &.fi-text-input--error {
       & .fi-text-input_input {
         border: 2px solid ${theme.colors.alertBase};
-        padding-left: ${math(`${theme.spacing.insetL} - 1`)};
+        padding-left: 9px;
         padding-top: 7px;
         padding-bottom: 7px;
       }
@@ -89,7 +89,7 @@ export const baseStyles = withSuomifiTheme(
     &.fi-text-input--success {
       & .fi-text-input_input {
         border: 2px solid ${theme.colors.successBase};
-        padding-left: ${math(`${theme.spacing.insetL} - 1`)};
+        padding-left: 9px;
         padding-top: 7px;
         padding-bottom: 7px;
       }
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -76,12 +76,16 @@ export const baseStyles = withSuomifiTheme(
 
     &.fi-text-input--error {
       & .fi-text-input_input {
-        border-color: ${theme.colors.alertBase};
+        border: 2px solid ${theme.colors.alertBase};
+        height: 38px;
+        padding-left: ${math(`${theme.spacing.insetL} - 1`)};
       }
     }
     &.fi-text-input--success {
       & .fi-text-input_input {
-        border-color: ${theme.colors.successBase};
+        border: 2px solid ${theme.colors.successBase};
+        height: 38px;
+        padding-left: ${math(`${theme.spacing.insetL} - 1`)};
       }
     }
     &.fi-text-input--disabled {
--- a/src/core/Form/TextInput/TextInput.baseStyles.tsx
+++ b/src/core/Form/TextInput/TextInput.baseStyles.tsx
@@ -8,8 +8,12 @@ export const baseStyles = withSuomifiTheme(
   ({ theme }: TokensAndTheme) => css`
     &.fi-text-input {
       ${font({ theme })('bodyText')}
-      display: inline-block;
       width: 290px;
+    }
+
+    &_wrapper {
+      width: 100%;
+      display: inline-block;
     }
 
     & .fi-text-input_input-element-container {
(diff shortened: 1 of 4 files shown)


src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -76,12 +80,18 @@ export const baseStyles = withSuomifiTheme(
 
     &.fi-text-input--error {
       & .fi-text-input_input {
-        border-color: ${theme.colors.alertBase};
+        border: 2px solid ${theme.colors.alertBase};
+        padding-left: ${math(`${theme.spacing.insetL} - 1`)};
```
aappoalander, line 84: I assume this is going to be replaced with a fixed value?

src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -76,12 +80,18 @@ export const baseStyles = withSuomifiTheme(
...
+        padding-top: 7px;
+        padding-bottom: 7px;
       }
     }
     &.fi-text-input--success {
       & .fi-text-input_input {
-        border-color: ${theme.colors.successBase};
+        border: 2px solid ${theme.colors.successBase};
+        padding-left: ${math(`${theme.spacing.insetL} - 1`)};
```
aappoalander, line 92: Same as above.

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -86,17 +88,14 @@ export interface TextInputProps
...
 
   private statusTextId: string;
 
   constructor(props: TextInputProps) {
     super(props);
-    this.id = `${idGenerator(props.id)}`;
-    this.hintTextId = `${this.id}-hintText`;
-    this.statusTextId = `${this.id}-statusText`;
+    this.hintTextId = `${props.id}-hintText`;
+    this.statusTextId = `${props.id}-statusText`;
```
aappoalander, line 98: these can now be handled in render since id is handled outside the component. No need for constructor at all.

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -151,40 +137,46 @@ class BaseTextInput extends Component<TextInputProps> {
...
+                  {...{ 'aria-invalid': status === 'error' }}
+                  {...getConditionalAriaProp('aria-describedby', [
+                    statusText ? this.statusTextId : undefined,
+                    hintText ? this.hintTextId : undefined,
+                    ariaDescribedBy,
+                  ])}
+                  onChange={(event: ChangeEvent<HTMLInputElement>) => {
+                    if (propOnChange) {
+                      const eventValue = event.currentTarget.value;
+                      debouncer(propOnChange, eventValue);
```
aappoalander, line 168: Have we tested that this does not work without storing the const? If not, maybe we could try it out.

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -151,40 +137,46 @@ class BaseTextInput extends Component<TextInputProps> {
...
+                    }
+                  }}
+                />
+              )}
+            </Debounce>
+            {resolvedIcon && <Icon {...newIconProps} />}
```
ketsappi, line 174: If `const resolvedIcon = icon || iconProps?.icon;` is really needed it might make sense to give the typing manually as now it is `any`:
```tsx
const resolvedIcon: BaseIconKeys = icon || iconProps?.icon;
```


These `newIconProps` could be inlined if wanted:

```tsx
{resolvedIcon && <Icon {...{ iconProps, icon: resolvedIcon }} />}
```

src/core/Form/TextInput/TextInput.baseStyles.tsx
```diff
@@ -76,12 +80,18 @@ export const baseStyles = withSuomifiTheme(
...
+        padding-top: 7px;
+        padding-bottom: 7px;
```
ketsappi, line 94: Not sure if it is really better but might help in future to have the common styles separately from unique styles, like: 
```scss
&.fi-text-input--error,
&.fi-text-input--success {
  & .fi-text-input_input {
    padding-left: ${math(`${theme.spacing.insetL} - 1`)};
    padding-top: 7px;
    padding-bottom: 7px;
  }
}
&.fi-text-input--error {
  & .fi-text-input_input {
    border: 2px solid ${theme.colors.alertBase};
  }
}
&.fi-text-input--success {
  & .fi-text-input_input {
    border: 2px solid ${theme.colors.successBase};
  }
}
```

It is a few more lines, so what do you think about this approach?

LJKaski: I gave that some thought too, but since the styles are only needed in two instances that are right next to each other in the document, I thought it cleaner to keep the duplicate code. In the unlikely case there is a third state in the future it would make sense to do things differently, but until then I think there's no value in it in either readability or in saved lines.

src/core/Form/TextInput/TextInput.tsx
LJKaski: Seems to work without as well. It was put there to avoid problems with React event pooling. But it could be that simply by giving the value to the function separately counters this issue. But I'm only 90% sure even though it seems to work.

src/core/Form/TextInput/TextInput.tsx
LJKaski: Good point. I like the inline version better, as I'm not a big fan of newThisAndThatProps kind of names. Changed it.

src/core/Form/TextInput/TextInput.baseStyles.tsx
LJKaski: You assume correctly. Changed.

src/core/Form/TextInput/TextInput.baseStyles.tsx
aappoalander: In my opinion, both work well in this case. Since these states can never overlap, the styles are "kind of not shared". Styles that apply for all states should be shared and seems that they are.

src/core/Form/TextInput/TextInput.tsx
aappoalander: It's a primitive value, so it is passed to the function as a copy. If you were to pass the event, things would be different. One explanation here: https://www.javascripttutorial.net/javascript-pass-by-value/

src/core/Form/TextInput/TextInput.baseStyles.tsx
LJKaski: Changed as well.

src/core/Form/TextInput/TextInput.tsx
LJKaski: Makes sense. Let's leave it without the const.

ketsappi (approved):

aappoalander (approved):

PR #438, opened 2021-02-16
[Feature] MultiSelect (Combobox)

ketsappi:
## Description

Combobox component for selecting one or multiple items from the list. 
* Shows the selected items as chips underneath the input field
* Able to filter items on the list

## Screenshots

Screenshot of Combobox, list closed  Screenshot of Combobox, list opened Screenshot of Combobox with filtering text, list opened 

## Motivation and Context

There was a need for Combobox component for multi-select functionality.

## How Has This Been Tested?

**Screenreader**: VoiceOver (macOS & iOS)
**Browsers**: Brave, Safari, Firefox

Tested in different projects:
* In styleguidist
* cra-ts project
* Next.js ssr&ssg

## Release notes

* Combobox component (#438)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "suomifi-ui-components",
-  "version": "3.0.0",
+  "version": "4.0.0-beta.5",
```
aappoalander, line 3: Should this be reset to 3.0.0 before merging to dev or is it better to leave as an indicator that betas were released?

ketsappi: I think we can reset it back to `3.0.0` and then update it to `4.0.0` when making the actual stable release.

We have the beta-release commits and tags available, so we can utilise that information.

src/core/Form/Combobox/ChipList/ChipList.baseStyles.tsx
```diff
@@ -0,0 +1,17 @@
...
+  &.fi-chip-list {
+    ${font(theme)('bodyText')}
+    padding-top: 15px;
+  }
+
+  & .fi-chip-list_wrapper {
+    display: inline-flex;
+    flex-wrap: wrap;
+    width: 100%;
+    gap: 10px;
```
aappoalander, line 15: This might not work as expected if there is no polyfill in place: https://caniuse.com/flexbox-gap

src/core/Form/Combobox/ChipList/ChipList.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { baseStyles } from './ChipList.baseStyles';
+
+const baseClassName = 'fi-chip-list';
+
+const chipListClassNames = {
+  wrapper: `${baseClassName}_wrapper`,
+};
+
+export interface ChipListProps extends HtmlDivProps {
+  className?: string;
```
aappoalander, line 15: Is there a reason why className is not documented?

src/core/Form/Combobox/ChipList/ChipList.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { default as styled } from 'styled-components';
+import classnames from 'classnames';
+import { HtmlDiv, HtmlDivProps } from '../../../../reset';
+import { AutoId } from '../../../../utils/AutoId';
+import { baseStyles } from './ChipList.baseStyles';
+
+const baseClassName = 'fi-chip-list';
+
+const chipListClassNames = {
+  wrapper: `${baseClassName}_wrapper`,
```
aappoalander, line 11: This is really minor, but could this be `content_wrapper`? Now it's a bit misleading as there is first the actual chip list and then the wrapper inside it

src/core/Form/Combobox/ChipList/ChipList.tsx
```diff
@@ -0,0 +1,46 @@
...
+        <HtmlDiv className={chipListClassNames.wrapper} {...passProps}>
+          {children}
+        </HtmlDiv>
+      </HtmlDiv>
+    );
+  }
+}
+
+const StyledChipList = styled(({ id: propId, ...passProps }: ChipListProps) => (
+  <AutoId id={propId}>{(id) => <BaseChipList id={id} {...passProps} />}</AutoId>
```
aappoalander, line 37: If it is possible typings wise, I'd rather move the AutoId to top level and have the styled version only handle styles.

src/core/Form/Combobox/Combobox/Combobox.baseStyles.tsx
```diff
@@ -0,0 +1,51 @@
+import { css } from 'styled-components';
+import { defaultThemeTokens as theme } from '../../../theme';
+import { font } from '../../../theme/reset';
+
+export const baseStyles = css`
+  &.fi-combobox {
+    ${font(theme)('bodyText')}
```
aappoalander, line 7: Does this work if I change e.g. display property of the "top-level" element with style prop? In other words, are the styles wrapped so that display and spacing properties can be safely applied?

src/core/Form/Combobox/Combobox/Combobox.baseStyles.tsx
```diff
@@ -0,0 +1,51 @@
...
+    width: 290px;
+
+    & .fi-filter-input_input-element-container {
+      position: relative;
+
+      &:before {
```
aappoalander, line 13: What are these styles? I guess not possible to use existing styles?

src/core/Form/Combobox/Combobox/Combobox.md
```diff
@@ -0,0 +1,102 @@
...
+    name: 'Powersaw',
+    price: 150,
+    tax: false,
+    labelText: 'Powersaw',
+    disabled: true,
+    uniqueItemId: 'ps9081231'
+  }
+];
+
+<>
```
aappoalander, line 90: Should we add another controlled example?

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+        }
+        // eslint-disable-next-line react/no-array-index-key
+        return <span key={i}>{substring}</span>;
+      });
+    };
+
+    const filter = (data: ComboboxData, query: string) =>
+      data.labelText.toLowerCase().includes(query.toLowerCase());
+
+    const handleKeyDown = (event: React.KeyboardEvent) => {
```
aappoalander, line 259: I propose we create utils.ts for handling keyboard and maybe even state.ts for handling state. This file seems a bit cumbersome to read already.

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+import { FilterInput, FilterInputStatus } from '../../FilterInput/FilterInput';
+import { ComboboxItemList } from '../ComboboxItemList/ComboboxItemList';
+import { ComboboxItem } from '../ComboboxItem/ComboboxItem';
+import { ComboboxEmptyItem } from '../ComboboxEmptyItem/ComboboxEmptyItem';
+import { ChipList } from '../ChipList/ChipList';
+import { baseStyles } from './Combobox.baseStyles';
+
+const baseClassName = 'fi-combobox';
+const comboboxClassNames = {
+  wrapper: `${baseClassName}_wrapper`,
```
aappoalander, line 20: Same as previous, should this be `content_wrapper` etc. instead?

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+  open: `${baseClassName}--open`,
+  removeAllButton: `${baseClassName}_removeAllButton`,
+  error: `${baseClassName}--error`,
+};
+
+export interface ComboboxData {
+  /** Unique label that will be shown on combobox item and used on filter */
+  labelText: string;
```
aappoalander, line 28: Does this need to be text or could it be ReactNode?

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+  items: Array<T & ComboboxData>;
+  /**
+   * Unique id
+   * If no id is specified, one will be generated automatically
+   */
+  id?: string;
+  /** Label */
+  labelText: string;
+  /** Event that is sent when item selections change */
+  onItemSelectionsChange?: (selectedItems: Array<T>) => void;
```
aappoalander, line 50: would simply `onSelectionChange` be too ambiguous?

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+  /** Show chip list */
+  chipListVisible?: boolean;
```
aappoalander, line 52: have we used this form of naming with booleans? I'm thinking `showChipList` instead, but not sure what would be in line with other components.

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+  /** Chip action label */
```
aappoalander, line 53: I guess this is referring to clicking a chip, if this is not visible, I think name should contain aria somehow.

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+  chipActionLabel?: string;
+  /** Label for remove button. If it is given, button will be shown. */
+  removeAllButtonLabel?: string;
+  /** Placeholder text for input. Use only as visual aid, not for instructions. */
+  visualPlaceholder?: string;
+  /** Label to show when no items to show, e.g filtered all out */
+  emptyItemsLabel: string;
```
aappoalander, line 60: `noItemsLabel`, `noResultsLabel` or even -Text rather than label?

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: Is this redundant as there is onItemSelected as well?

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+  filteredItems: T[];
+  showPopover: boolean;
+  currentSelection: string | null;
+  selectedKeys: ItemKeys;
+  selectedItems: T[];
+  initialItems: T[];
+  disabledKeys: ItemKeys;
+}
+
+function getSelectedKeys<T>(items: (T & ComboboxData)[] = []): ItemKeys {
```
aappoalander, line 98: Would these be more efficient if there was just getItemKeys which would iterate all items and set selected and disabled values? Accessing items[key].disabled should be as efficient as accessing disabledItems[key] regardless of number of items in the objects.

package.json
aappoalander: I wonder what git-flow says about this. I kind of think that development should reflect what has been released as beta but not sure about it. It would be more clear what beta version should be released next with a new feature branch and having the beta version in package json would help with that. It should not conflict with hotfixes either since those should be merged to develop as well.

src/core/Form/Combobox/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+    await act(async () => {
+      const { container } = render(BasicCombobox);
+      expect(container.querySelectorAll('.fi-chip').length).toEqual(2);
+    });
+  });
+
+  it('first Chip should be removable and removed when clicked', async () => {
+    await act(async () => {
+      const { container } = render(BasicCombobox);
+      let chips = container.querySelectorAll('.fi-chip');
```
aappoalander, line 122: Would it be possible to follow testing library practices more closely and use getByText?

src/core/Form/Combobox/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+          emptyItemsLabel="No items"
+          defaultSelectedItems={defaultSelectedTools}
+          onItemSelect={onItemSelect}
+        />,
+      );
+      const hammerChip = container.querySelectorAll('.fi-chip')[0];
+      await act(async () => {
+        fireEvent.click(hammerChip, {});
+      });
+      expect(onItemSelectSpy).toBeCalledWith('h9823523');
```
aappoalander, line 157: This is neat, we haven't used this much as of yet 💪

src/core/Form/Combobox/Combobox/Combobox.test.tsx
```diff
@@ -0,0 +1,364 @@
...
+      await act(async () => {
+        fireEvent.click(removeAllButton, {});
+      });
+      expect(mockOnRemoveAll).toBeCalledTimes(1);
+    });
+  });
+});
+
+describe('Controlled', () => {
+  it('has the controlled items as selected + disabled', async () => {
```
aappoalander, line 212: Maybe add a test for checking that controlled state prevents selecting / deselecting if controlled state is not changed?

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+        if (onItemSelectionsChange) {
+          onItemSelectionsChange(disabledItems);
+        }
+        return { selectedKeys: newSelectedKeys, selectedItems: disabledItems };
+      },
+    );
+  };
+
+  render() {
+    const highlightQuery = (text: string, query: string) => {
```
aappoalander, line 243: These utils don't have to be inside render unless they access props directly

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+            }
+          }
+          break;
+        }
+
+        case 'Escape': {
+          event.preventDefault();
+          this.setState(
+            (
+              _: ComboboxState<T & ComboboxData>,
```
aappoalander, line 337: It seems there is no proper convention for dismissing unused parameters and destructuring cannot be used here I assume. Maybe name the parameter `_parameterName` so that it's more clear what parameter it is in addition to it not being used.

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+        aria-expanded={showPopover}
+        {...passProps}
+        className={classnames(baseClassName, className, {
+          [comboboxClassNames.open]: showPopover,
+          [comboboxClassNames.error]: status === 'error',
+        })}
+      >
+        <HtmlDiv className={classnames(comboboxClassNames.wrapper, {})}>
+          <Debounce waitFor={debounce}>
+            {(debouncer: Function) => (
```
aappoalander, line 476: Is this the convention we have been using before? Naming sounds funny, but it's understandable so I guess it's good as it is.

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+                onFilter={(filtered) =>
+                  this.setState({ filteredItems: filtered })
+                }
+                filterFunc={filter}
+                forwardedRef={this.filterInputRef}
+                onFocus={() => setPopoverVisibility(true)}
+                onKeyDown={handleKeyDown}
+                onBlur={handleBlur}
+                value={filterInputValue}
+                onChange={(value: string | undefined) => {
```
aappoalander, line 490: Is it really undefined if the input is empty, or can it be null or "" if no value is provided?

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+                status={status}
+                statusText={statusText}
+              />
+            )}
+          </Debounce>
+          <Popover
+            sourceRef={this.filterInputRef.current}
+            matchWidth={true}
+            id={`${id}-popover`}
+            portalStyleProps={{ backgroundColor: 'white' }}
```
aappoalander, line 506: Is there a better way for handling these? Popover is probably implemented this way currently, but using inline js styles is not the best way to go.

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,595 @@
...
+            {showPopover && (
+              <ComboboxItemList
+                forwardRef={this.popoverListRef}
+                onBlur={handleBlur}
+                aria-activedescendant={
+                  currentSelection ? `${id}-${currentSelection}` : undefined
+                }
+              >
+                {filteredItems.length > 0 ? (
+                  filteredItems.map((item) => {
```
aappoalander, line 518: Is the check necessary? mapping an empty array should return nothing?

src/core/Form/Combobox/ComboboxEmptyItem/ComboboxEmptyItem.baseStyles.tsx
```diff
@@ -0,0 +1,18 @@
+import { css } from 'styled-components';
+import { defaultThemeTokens as theme } from '../../../theme';
+import { font } from '../../../theme/reset';
+
+export const baseStyles = css`
+  &.fi-combobox-empty-item {
+    ${font(theme)('actionElementInnerText')}
+    padding: 10px;
```
aappoalander, line 8: Should we use a token here or is this "component internals"?

src/core/Form/Combobox/ComboboxEmptyItem/ComboboxEmptyItem.tsx
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from 'react';
+import { default as styled } from 'styled-components';
+import classnames from 'classnames';
+import { HtmlDiv, HtmlLi } from '../../../../reset';
+import { baseStyles } from './ComboboxEmptyItem.baseStyles';
+
+const baseClassName = 'fi-combobox-empty-item';
+
+const comboboxItemClassNames = {
+  wrapper: `${baseClassName}_wrapper`,
```
aappoalander, line 10: And same for this maybe conte_wrapper?

src/core/Form/Combobox/ComboboxItem/ComboboxItem.baseStyles.tsx
```diff
@@ -0,0 +1,70 @@
...
+    &:focus {
+      outline: none;
+    }
+
+    & .fi-checkbox {
+      padding: 10px;
+      user-select: none;
+      ${font(theme)('actionElementInnerText')}
+
+      & mark {
```
aappoalander, line 16: This might be a bit risky since we don't have any rests for mark-tag. Any global styles would be applied to these as well. Not sure if this is a problem, but maybe worth noting still.

src/core/Form/Combobox/ComboboxItem/ComboboxItem.baseStyles.tsx
```diff
@@ -0,0 +1,70 @@
...
+    &--currentSelection {
+      & .fi-checkbox {
+        background-color: ${theme.colors.highlightBase};
+        color: ${theme.colors.whiteBase};
+
+        & mark {
+          color: ${theme.colors.whiteBase};
+        }
+
+        &.fi-checkbox--checked {
```
aappoalander, line 44: Should this also be a descendant selector `& .fi-...`like other similar styles seem to be?

src/core/Form/Combobox/ComboboxItem/ComboboxItem.tsx
```diff
@@ -0,0 +1,66 @@
...
+const comboboxItemClassNames = {
+  wrapper: `${baseClassName}_wrapper`,
+  currentSelection: `${baseClassName}--currentSelection`,
+};
+
+export interface ComboboxItemProps
+  extends Omit<CheckboxProps, 'ref' | 'forwardedRef'> {
+  /** ComboboxItem container div class name for custom styling. */
+  className?: string;
+  currentSelection: boolean;
```
aappoalander, line 19: Does this refer to active, hover or something completely different?

src/core/Form/Combobox/ComboboxItemList/ComboboxItemList.baseStyles.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    border-width: 0 1px 1px 1px;
+    border-style: solid;
+    border-color: ${theme.colors.depthDark3};
+    border-bottom-left-radius: 2px;
+    border-bottom-right-radius: 2px;
+
+    overflow-y: auto;
+    overflow-x: hidden;
+    max-height: 260px;
```
aappoalander, line 18: Is this something that can be / should be possible to override?

src/core/Form/Combobox/ComboboxItemList/ComboboxItemList.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { default as styled } from 'styled-components';
+import classnames from 'classnames';
+import { HtmlDiv, HtmlUlWithRef } from '../../../../reset';
+import { ComboboxItemProps } from '../ComboboxItem/ComboboxItem';
+import { baseStyles } from './ComboboxItemList.baseStyles';
+
+const baseClassName = 'fi-combobox-item-list';
+
+const comboboxItemListClassNames = {
+  wrapper: `${baseClassName}_wrapper`,
```
aappoalander, line 11: ditto

aappoalander:
Huge effort with great success! 💪  Looks really good overall, most of comments I left concern naming conventions and need to be agreed together.

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: Typing this "the wrong way" creates an issue for projects using the component as type casts are required on the using end.

src/core/Form/Combobox/Combobox/Combobox.baseStyles.tsx
ketsappi: Can't remember if I have tested that. Need to check this.

What did you find when you tested this?

src/core/Form/Combobox/Combobox/Combobox.baseStyles.tsx
ketsappi: It is the caret icon of the input field; when list is open it will be then "upside" down.

Styles are taken from the Dropdown's caret icon with slight change (fix to pointer events). We could make it to some common, shared style?

aappoalander: Ah, makes sense. I would add this to "todo" as this same icon will be needed in other components later on. Doesn't have to be done now in my opinion.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: Then the filter functionality and highlighting feature needs changes to support that. And also the `chipText` etc.

At this point I don't see a reason to introduce more complex logic and increasing changes for user to break accessibility of the component.

But of course if this is needed / seen to prove enough value with low effort, then we could change it. Now or later in separate PR's.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: Dropping the `item` from prop names would be okay for me, as there is no other stuff to be selected than those items.

It is true that both are touching same area, so we can make changes here if needed. 
Here are the differences:

### onItemSelectionsChange
The first `onItemSelectionsChange` provides **all** the items when selection changes. 
e.g user could store current selections somewhere with these.

### onItemSelect
And the latter, `onItemSelect`,  just the **one** that is selected, with along it's `uniqueItemId`.
e.g user can use this for controlled stuff or some other detailed way for just that specific item.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: Will need to be checked against other components using booleans.

Somehow `showChipList`seems wrong to be for prop name as it starts with an action and therefore associated to be function (in my mind at least :D )

package.json
aappoalander: I propose we leave it as it is. This way it shows history better and it's clear what has been published.

src/core/Form/Combobox/Combobox/Combobox.baseStyles.tsx
aappoalander: I didn't test this particular case yet. I just know we have had issues with this in the past and didn't see a "style wrapper" anywhere, or at least didn't notice one.

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: Makes sense, I didn't take that into account. Lets leave it as it is!

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: Having both makes sense, but I'm just wondering if it's necessary? Maybe leave it as it is.

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: True, naming semantics are difficult. On the other hand, pretty often action start with something referring to an event, onChange, onBlur, onSelectItem and so on. Found a couple of references that proposed using is/has/can prefix for booleans. The most important thing in my opinion is to be consistent within this library - same conventions should be used everywhere.

src/core/Form/Combobox/Combobox/Combobox.test.tsx
ketsappi: It is possible, but thought was to reuse the `chips` as the length is needed in later part of the test. 
And also at the time of writing the test it was thought that the list items (popover) are visible also in the DOM and therefore causing troubles to find the correct one with minimum effort.

src/core/Form/Combobox/Combobox/Combobox.test.tsx
ketsappi: Good idea 👍

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: `TextInput`, `SearchInput` and `Debounce` uses this naming at this time so was following the naming.

Is there suggestion for more describing name?

src/core/Form/Combobox/ComboboxItem/ComboboxItem.tsx
ketsappi: It is the `true` when the keyboard focus is on that item. And that item is then styled based on that.

It could be renamed e.g `hasKeyboardFocus` to be more describing.

src/core/Form/Combobox/ComboboxItemList/ComboboxItemList.baseStyles.tsx
ketsappi: Good question. 
This height was chosen based on the wanted amount of items to be shown by the design and to give visual hint of the next items in the list.

Haven't tried to override it by the CSS to see if it is already possible. But hopefully users of the library won't need to customise things like these s that the looks between different services/sites are looking same.

src/core/Form/Combobox/ComboboxItem/ComboboxItem.baseStyles.tsx
ketsappi: Can not say straight away, need to check it.

What did you find when you tested to change it?

src/core/Form/Combobox/ComboboxItem/ComboboxItem.baseStyles.tsx
ketsappi: Not sure what change you are after, but do you mean that the `mark` "component" should also have some custom className?
e.g `fi-combobox-item_highlight-filter` or something?

src/core/Form/Combobox/ComboboxEmptyItem/ComboboxEmptyItem.baseStyles.tsx
ketsappi: Good find. I think we could use value from token here with the `padding`.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: As you can see it is used in ternary operator; if no items are in the array the `ComboboxEmptyItem` is shown. 

Of course it could be written otherwise. e.g having the condition for the empty item straight away.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: This `portalSyleProps` might be leftover from the popover example. 
Might not be even needed in this case to give it background color (assuming popover  has `white` colour by default)

src/core/Form/Combobox/ChipList/ChipList.tsx
ketsappi: No specific reason. 
Might have been just left out as the prop is not visible to outside/user.

Can copy&paste the same JSDoc for this to be in line with others.

src/core/Form/Combobox/ChipList/ChipList.tsx
ketsappi: This is now changed to be like suggested, `content_wrapper`.

src/core/Form/Combobox/ChipList/ChipList.tsx
ketsappi: Added JSDoc for the `className` prop.

src/core/Form/Combobox/ChipList/ChipList.tsx
ketsappi: `AutoId` have now been moved to top level.

src/core/Form/Combobox/Combobox/Combobox.baseStyles.tsx
ketsappi: JIRA issue is now created to our backlog from this.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: This is now changed to be `content_wrapper` also.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: Would `ariaChipActionLabel` be good candidate?

It seems that Chip component itself is missing the aria from prop name too: 
https://github.com/vrk-kpa/suomifi-ui-components/blob/a5d6f9bc378805114682b0adca1d15fbb2cb20f0/src/core/Chip/Chip/Chip.tsx#L36-L37

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: Maybe `noItemsText` would be good one?

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: As discussed elsewhere, this is probably mostly in line with what we have in other components. So I propose we leave it as is.

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: Sounds a bit funny this way, but it would be better in line with what we have elsewhere and would make it distinguishable from other props without reading the API desc.

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: Text is better than label as it really is text in this case and will be rendered in "different" context.

src/core/Form/Combobox/Combobox/Combobox.test.tsx
aappoalander: Sometimes it is cumbersome to use "usage based" selectors. Especially when there is multiple layers on top of the actual element containing the text.

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: I'd prefer something like `debounceFn` etc. But if this is used elsewhere I propose we keep using the same convention.

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: If it doesn't we might have to use inline styles, as the portal is not a descendant of the actual component => styles are not necessarily applied correctly when nested.

aappoalander: If it works and is required, I'd prefer something like `.fi-popover { ...`

src/core/Form/Combobox/Combobox/Combobox.tsx
aappoalander: I missed the conditional empty item there at the end. Not sure if it would be cleaner with the map without any condition and then the condition right next to the empty item description. There should be no performance impact, so both are fine for me.

src/core/Form/Combobox/ComboboxItem/ComboboxItem.baseStyles.tsx
aappoalander: Sorry for not being descriptive with my comment. As this is wrapped inside .fi-checkbox it is probably fine the way it is now, but I'd still prefer the approach with classname rather than tag name.

src/core/Form/Combobox/ComboboxItem/ComboboxItem.baseStyles.tsx
aappoalander: I didn't test changing, it just caught my eye as it was different from other selectors.

src/core/Form/Combobox/ComboboxItemList/ComboboxItemList.baseStyles.tsx
aappoalander: I hope so too. Just crossed my mind as I was just working on another fix to allow resizing the Modal.

src/core/Form/Combobox/ComboboxItemList/ComboboxItemList.tsx
ketsappi: This is now also changed to be `content_wrapper`.

src/core/Form/Combobox/ComboboxItem/ComboboxItem.tsx
ketsappi: It is now renamed as `hasKeyboardFocus`.

src/core/Form/Combobox/Combobox/Combobox.tsx
```diff
@@ -0,0 +1,601 @@
...
+  /** Items for the combobox */
+  items: Array<T & ComboboxData>;
+  /**
+   * Unique id
+   * If no id is specified, one will be generated automatically
+   */
+  id?: string;
+  /** Label */
+  labelText: string;
+  /** Event that is sent when item selections change */
```
LJKaski, line 49: Nitpicking here, but events are usually fired, triggered or raised, but not sent.

src/core/Form/Combobox/ComboboxItem/ComboboxItem.baseStyles.tsx
ketsappi: `mark` tag now has the `fi-combobox-item--query_highlight` class applied to it and it is used for the styling.

src/core/Form/Combobox/ComboboxEmptyItem/ComboboxEmptyItem.baseStyles.tsx
ketsappi: Changed `padding` to now use the token: 
``` css
padding: ${theme.spacing.insetL};
```

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: It seems that without giving the `color` for  `portalStyleProps` the colour will be transparent by default.

The easiest solution seems to be that I give white background color for the `.fi-combobox-item-list`.

src/core/Form/Combobox/Combobox/Combobox.test.tsx
aappoalander: Actually, I would prefer to check with getByText rather than relying on length only. But not a big deal and will leave it up to you to either leave as is or change.

src/core/Form/Combobox/ComboboxItem/ComboboxItem.baseStyles.tsx
aappoalander: This is as it should be, --checked class is applied to the same element with fi-checkbox.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: This is now renamed to `_prevState`

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: Moved the majority of the utils outside of render where applicable and with slight changes.

src/core/Form/Combobox/ChipList/ChipList.baseStyles.tsx
ketsappi: This is now using margin + padding combination. As discussed it can in some cases wrap to next line because of that vs to the gap that works smart.
But at this time I think is a valid compromise to not be stopped by small style issue. And if having good fix for this later, it can the be added in separate PR.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: This is now changed to `noItemsText`.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: This is now changed to `ariaChipActionLabel`.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: This is now changed to be `fired` instead of previous `sent`.

src/core/Form/Combobox/Combobox/Combobox.baseStyles.tsx
ketsappi: I now tested with different values for `display`, having some flex related stuff and using `padding` and `margin` like the following:
```tsx
  <Combobox style={{display: 'flex', flexDirection: 'row', justifyContent: 'space-around', margin: 50, padding: 20}}
...
```
But couldn't notice any breaking of the component or unwanted behaviour.

src/core/Form/Combobox/Combobox/Combobox.md
ketsappi: Controlled example is now added.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: Not sure about the efficiency here. By quickly glancing that causes many changes to the current logic and therefore possibly increasing complexity in many places. So the efficiency gained in here vs the efficient lost in those other places seems unwanted.
I suggest that this can be improved in later PR if it is a real problem and we need to optimise it.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: I would like to see able to make it more simple and easier to read, but the handling of keydown and state are so tightly entwined.
I suggest that we could improve this in separate PR to increase readability and maintainability and to not delay this PR with these changes.

src/core/Form/Combobox/Combobox/Combobox.test.tsx
ketsappi: Test is now changed to use `getByText` and `queryByText` because getByText throws an error if trying to find an element that does not exist.
Not relying on the **length** anymore

src/core/Form/Combobox/Combobox/Combobox.test.tsx
ketsappi: Added a test that checks that item couldn't be removed when it was controlled, using `selectedItems` prop.

src/core/Form/Combobox/Combobox/Combobox.tsx
ketsappi: This is currently done in `FilterInput` as it was done with `TextInput`, but only not having the `number` as one option.

https://github.com/vrk-kpa/suomifi-ui-components/blob/0b50e33f48b8175bc2495293318ba590fe772e98/src/core/Form/TextInput/TextInput.tsx#L34-L47

ketsappi: When investigated this a little more and IIRC, it seems that only the **string** value is possible.

This is now changed to (and some other changed that was needed after this)
```tsx
onChange?: (value: string) => void;
```

aappoalander (approved):

LJKaski (approved):
Great work!

PR #439, opened 2021-02-22
[Feature] Remove tokens prop support

aappoalander:
## Description
This PR removes support for tokens prop for every component. In addition, it removes internal theme utils that handle merging custom and default tokens. This is done to have better performance and not recalculate styles on every render, but rather during build.

Some styles that depend on props were changed to use a class approach instead. Colors is now the only component that uses props for calculating styles and it is only used internally in styleguidist.

Some minor fixes for previously broken or missing styles and class names were also done.

## Motivation and Context
Merging custom and default tokens and calculating styles on every render caused poor performance on situations where there were a large number of components in one view.

## How Has This Been Tested?
Tested using Styleguidist, CRA TS and SSR test projects. In addition a performance test was done using a large number of components and a clean CRA TS setup. Tested using Mac OS and Win 10 with Chrome, Firefox, Safari and Edge + VoiceOver/NVDA.

There are only a couple of minor changes in the snapshots and none in the actual resulting styles.

## Release notes
- Breaking Change: Tokens prop is now removed from all components


src/core/Breadcrumb/Breadcrumb.baseStyles.ts
```diff
@@ -1,40 +1,38 @@
...
 import { nav, list, listItem, font } from '../theme/reset';
 
-export const baseStyles = withSuomifiTheme(
-  ({ theme }: TokensAndTheme) => css`
-    ${nav({ theme })}
-    ${font({ theme })('bodyTextSmall')}
+export const baseStyles = css`
+  ${nav(theme)}
+  ${font(theme)('bodyTextSmall')}
     background-color: ${theme.colors.whiteBase};
```
ketsappi, line 9: Indentation doesn't match with others for the background-color.

Tried locally to change it to as others and ran linters and prettier; all seemed happy also then. 

_I remember we had similar things earlier also, but then there was something that linter/prettier did change it back to wrong indentation(?)_

src/core/Button/Button.tsx
```diff
@@ -127,12 +123,8 @@ class BaseButton extends Component<ButtonProps> {
...
 }
 
-const StyledButton = styled(
-  ({ tokens, ...passProps }: ButtonProps & InternalTokensProp) => (
-    <BaseButton {...passProps} />
-  ),
-)`
-  ${(props) => baseStyles(props)}
+const StyledButton = styled((props: ButtonProps) => <BaseButton {...props} />)`
+  ${baseStyles}
```
ketsappi, line 127: In this and in similar places it seems to be possible to use simplified version as the props are not used like those were before.

Like this:
```
const StyledButton = styled(BaseButton)`
  ${baseStyles}
`;
```

src/core/Dropdown/Dropdown/Dropdown.test.tsx
```diff
@@ -209,6 +208,5 @@ describe('Dropdown', () => {
 });
 
 test('CSS export', () => {
-  const css = cssFromBaseStyles(baseStyles);
-  expect(css).toEqual(expect.stringContaining('color'));
+  expect(baseStyles.join('')).toEqual(expect.stringContaining('color'));
 });
```
ketsappi, line 212: I vote to remove this whole test as it's value seems to be very low and vague.

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -22,15 +20,16 @@ const textareaClassNames = {
   fullWidth: `${baseClassName}--full-width`,
   textareaContainer: `${baseClassName}_textarea-element-container`,
   textarea: `${baseClassName}_textarea`,
+  resizeHorizontal: `${baseClassName}_textarea-resize--horizontal`,
+  resizeBoth: `${baseClassName}_textarea-resize--both`,
+  resizeNone: `${baseClassName}_textarea-resize--none`,
```
ketsappi, line 25: This was good change (with all related changes) 👍

src/core/Form/Textarea/Textarea.baseStyles.tsx
```diff
@@ -1,86 +1,94 @@
 import { css } from 'styled-components';
-import { TextareaProps } from './Textarea';
-import { withSuomifiTheme, TokensAndTheme } from '../../theme';
+import { defaultThemeTokens as theme } from '../../theme';
 import { disabledCursor } from '../../../components/utils/css';
 import { element, font } from '../../theme/reset';
 import { absoluteFocus } from '../../theme/utils/focus';
-import { Omit } from '../../../utils/typescript';
```
ketsappi, line 7: This remove is good as TypeScript has the **Omit** nowadays. 

Maybe creating ticket to backlog to remove the use the custom Omit also in other places so we can get rid of it altogether.

src/core/Form/Toggle/ToggleButton/ToggeButton.baseStyles.tsx
```diff
@@ -1,24 +1,25 @@
...
-        outline: 0;
-        & .fi-toggle_icon-container {
-          &:after {
-            ${absoluteFocus}
-            ${focusOverrides}
-          }
+export const baseStyles = css`
+  ${toggleBaseStyles}
+  &.fi-toggle--disabled > button {
+    ${disabledCursor}
```
ketsappi, line 12: Just remembered that in some other place we removed the use of this util and just told the style right away:

```css
cursor: not-allowed;
```

Maybe not the matter of this task, but creating ticket for backlog to get rid of this util?

src/core/StaticIcon/__snapshots__/StaticIcon.test.tsx.snap
```diff
@@ -6,18 +6,31 @@ exports[`calling render with the same component on the same container does not r
   vertical-align: baseline;
 }
 
+<<<<<<< HEAD
```
ketsappi, line 9: This snapshot seem to have some Git stuff going 👀

src/core/StaticIcon/StaticIcon.tsx
```diff
@@ -49,18 +57,18 @@ export class StaticIcon extends Component<StaticIconProps> {
...
     if (icon !== undefined) {
       return (
         <StyledSuomifiStaticIcon
           {...passProps}
-          className={classnames(baseClassName, className)}
           {...(!!highlightColor
             ? { highlightColor }
             : { highlightColor: tokens.colors.accentBase })}
           icon={icon}
+          className={getClassNames(this.props)}
```
ketsappi, line 71: Was there a certain benefit for having the classnames done in separate function?

src/core/LanguageMenu/LanguageMenu.test.tsx
```diff
@@ -36,6 +35,5 @@ test(
 );
 
 test('CSS export', () => {
-  const css = cssFromBaseStyles(baseStyles);
-  expect(css).toEqual(expect.stringContaining('color'));
+  expect(baseStyles.join('')).toEqual(expect.stringContaining('color'));
```
ketsappi, line 38: Voting to get rid of this test also.

src/core/Breadcrumb/Breadcrumb.baseStyles.ts
aappoalander: Will fix, this are left untouched by linters/prettier.

src/core/Button/Button.tsx
aappoalander: Good point here as well. Not sure if there are any perf implications, but it's better to have it cleaner.

src/core/Form/Textarea/Textarea.tsx
aappoalander: Might be needed elsewhere as well, but gets complicated quickly when there are a lot of options.

src/core/StaticIcon/__snapshots__/StaticIcon.test.tsx.snap
aappoalander: Fixed already. Bad rebase.

src/core/Dropdown/Dropdown/Dropdown.test.tsx
aappoalander: Removed from both Dropdown and Language menu.

src/core/Form/Textarea/Textarea.baseStyles.tsx
aappoalander: There were only three cases left. Those are now removed with the definition.

src/core/Form/Toggle/ToggleButton/ToggeButton.baseStyles.tsx
aappoalander: Was such a small effort that replaced all with `cursor: not-allowed;` and removed the whole tooling. These kind of tools with only one function and no updates in the foreseeable future should be removed. Unnecessary abstractions.

src/core/StaticIcon/StaticIcon.tsx
aappoalander: Removed and logic now in JSX.

src/core/LanguageMenu/LanguageMenu.test.tsx
aappoalander: Removed.

src/core/Button/Button.tsx
aappoalander: These should be now better. Did some refactoring in few places and moved AutoId to main level. The ones with className logic in styled were left untouched.

src/core/Breadcrumb/Breadcrumb.baseStyles.ts
aappoalander: Fixed all I could find.

LJKaski (approved):
Tested in Chrome, Edge and FF with styleguidist, DS-site and and CRA project + using NVDA just in case. Styles seem to work as they should.

PR #441, opened 2021-03-04
[Feature] Ref prop for form components

LJKaski:
## Description
This PR adds ref prop to form components to allow  passing custom refs to the underlying html components.
### Changes in short:
* Ref prop added to form components
* Static variants removed from many components
  * Variant prop should be used instead
* Chip component split into Chip and StaticChip components
  * StaticChip replaces the static variant
* Many components now export a const instead of a class due to the ref addition
 
## Motivation and Context
Projects using the components asked for custom ref options in order to use the components with form libraries and control focus in certain cases.

## How Has This Been Tested?
Tested in styleguidist and a CRA project on chrome so far as well as with automated tests.

### Button, Checkbox, Chip, Dropdown, RadioButton, TextInput, Textarea and Toggle
* Added ref prop
### Button
* **Breaking change**: static variants no longer available. Use variant prop instead. 
### Checkbox
* **Breaking change**: static variants no longer available. Use variant prop instead. 
### Chip
* **Breaking change**: component split into two different components, Chip and StaticChip. Old static variants are no longer in use.


src/core/Button/Button.md
```diff
@@ -1,11 +1,18 @@
...
+import React from 'react';
+
+const testi = React.createRef();
 
 <>
   <Button
     className="my-button--test"
     onClick={() => console.log('Test button click')}
     type="submit"
+    ref={testi}
```
aappoalander, line 12: should this be in english `test` rather than `testi`?

LJKaski: Ah, this is something I meant to do before pressing 'ready for review' but forgot 😅 Many components still have these makeshift test examples. Do you think it would be necessary to present the ref prop in the examples in the first place? I originally thought about simply getting rid of them when I'm done.

src/core/Button/Button.tsx
```diff
@@ -82,11 +92,12 @@ class BaseButton extends Component<ButtonProps> {
 
     return (
       <>
-        <HtmlButton
+        <HtmlButtonWithRef
```
aappoalander, line 95: Do we have a use case for HtmlButton without the ref still? And if we do, could we just use Omit<HtmlButtonWithRefProps, 'ref'>? This way there would be no need for separate variants.

src/core/Chip/Chip/Chip.md
```diff
@@ -1,15 +1,23 @@
...
+
 <>
   <div>
     <span style={{ marginRight: '10px' }}>
-      <Chip removable actionLabel="Deselect" onClick={removeAction}>
-        Removable chip 1
+      <Chip
+        removable
+        actionLabel="Deselect"
+        ref={testi}
```
aappoalander, line 17: Same for this and following, I'd rather have this in english.

src/core/Chip/Chip/Chip.tsx
```diff
@@ -0,0 +1,92 @@
...
+import classnames from 'classnames';
+import { default as styled } from 'styled-components';
+import { baseStyles } from './Chip.baseStyles';
+import { HtmlButtonWithRef, HtmlSpan } from '../../../reset';
+import { Icon } from '../../Icon/Icon';
+import { logger } from '../../../utils/logger';
+import { VisuallyHidden } from '../../../components/Visually-hidden/Visually-hidden';
+import { StaticChipProps, chipClassNames, baseClassName } from '../StaticChip';
+
+interface InternalChipProps extends StaticChipProps {
```
aappoalander, line 11: Would it make sense to extend also button props?

src/core/Chip/StaticChip.test.tsx
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { StaticChip as OrigStaticChip, StaticChipProps } from './StaticChip';
+
+const StaticChip = (props: StaticChipProps) => {
```
aappoalander, line 5: Having this at the top level does not feel necessary. Would be easier to read if all describes would just have their own definitions as this abstraction does not provide much additional value.

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -175,7 +185,8 @@ class BaseDropdown extends Component<DropdownProps> {
             ])}
             disabled={disabled}
             onChange={onChange}
-            name={name}
+            ref={forwardedRef}
+            name={name} // TODO: Find the correct typing for forwardRef
```
aappoalander, line 189: Is this comment still correct? At least it seems to be pointing to wrong line?

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -1,12 +1,12 @@
...
+import React, { Component, ChangeEvent, forwardRef } from 'react';
 import { default as styled } from 'styled-components';
 import classnames from 'classnames';
 import { InputStatus } from '../types';
 import {
-  HtmlInput,
   HtmlInputProps,
   HtmlDiv,
   HtmlDivProps,
+  HtmlInputWithRef,
```
aappoalander, line 9: Same comment for this as for the HtmlButton. Should we just drop the HtmlInput without ref and rename the withRef to match the original? Btw. there was a typing issue with the original input with ref, has this been fixed?

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -153,16 +165,11 @@ class BaseFilterInput<T> extends Component<FilterInputProps<T>> {
 }
 
 const FilterInputWithoutTokens: <T>(
```
aappoalander, line 167: This should be renamed as there are no more tokens, probably leftovers from earlier changes.

src/core/Form/Textarea/Textarea.md
```diff
@@ -1,11 +1,17 @@
 ```js
 import { Textarea } from 'suomifi-ui-components';
+import React from 'react';
 
+const testiReffi = React.createRef();
```
aappoalander, line 5: I'd also change this testi riffi to just be const ref, const test, etc. Maybe use ref for all of these?

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -1,9 +1,9 @@
-import React, { Component, ChangeEvent, FocusEvent } from 'react';
+import React, { Component, ChangeEvent, FocusEvent, forwardRef } from 'react';
 import { default as styled } from 'styled-components';
 import classnames from 'classnames';
 import { baseStyles } from './Textarea.baseStyles';
 import {
-  HtmlTextarea,
+  HtmlTextareaWithRef,
```
aappoalander, line 6: And the same for this, just merge the variants?

src/core/Form/Textarea/Textarea.tsx
```diff
@@ -79,7 +79,16 @@ export interface TextareaProps extends Omit<HtmlTextareaProps, 'placeholder'> {
   containerProps?: Omit<HtmlDivProps, 'className'>;
 }
 
-class BaseTextarea extends Component<TextareaProps> {
+interface InnerRef {
+  forwardedRef: React.RefObject<HTMLTextAreaElement>;
+}
+
+export interface TextareaProps extends InternalTextareaProps {
+  /** Ref object to be passed */
```
aappoalander, line 87: + to the textarea element

src/core/Button/Button.md
aappoalander: Not necessary, but might be beneficial also as I think this is pretty commonly used pattern. Could drop for buttons, but leave in place for inputs?

aappoalander:
Looks solid! 💪  A couple of internal things would benefit from minor changes (see comments), but otherwise looks good to go.

src/core/Button/Button.md
aappoalander: Looks good now!

src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -95,7 +95,16 @@ export interface CheckboxProps {
   value?: string;
 }
 
-class BaseCheckbox extends Component<CheckboxProps> {
+interface InnerRef {
+  forwardedRef: React.RefObject<HTMLInputElement>;
+}
+
+export interface CheckboxProps extends InternalCheckboxProps {
+  /** Ref object to be passed */
```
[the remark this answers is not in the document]
aappoalander: Seems to be fixed now.

src/core/Form/FilterInput/FilterInput.tsx
aappoalander: Might have been better to indicate a type cast in the name as it is now the only function for the renamed FilterInputWithoutTokens (UnstyledFilterInput). But it's ok for me as it is now.

src/core/Chip/Chip/Chip.tsx
aappoalander: Looks good now!

src/core/Form/FilterInput/FilterInput.tsx
aappoalander: Looks good now!

src/core/Chip/Chip/Chip.tsx
```diff
@@ -1,11 +1,11 @@
...
 import { default as styled } from 'styled-components';
 import { baseStyles } from './Chip.baseStyles';
-import { HtmlButton, HtmlSpan } from '../../reset';
-import { Icon } from '../Icon/Icon';
-import { logger } from '../../utils/logger';
-import { VisuallyHidden } from '../../components/Visually-hidden/Visually-hidden';
+import { HtmlButton, HtmlButtonProps, HtmlSpan } from '../../../reset';
+import { Icon } from '../../Icon/Icon';
+import { logger } from '../../../utils/logger';
+import { VisuallyHidden } from '../../../components/Visually-hidden/Visually-hidden';
```
ketsappi, line 8: Just a small thing, but maybe ordering imports based on their folder structure like this:

```tsx
import { HtmlButton, HtmlButtonProps, HtmlSpan } from '../../../reset';
import { VisuallyHidden } from '../../../components/Visually-hidden/Visually-hidden';
import { logger } from '../../../utils/logger';
import { baseStyles } from './Chip.baseStyles';
```

src/core/Chip/StaticChip.test.tsx
LJKaski: You're absolutely right. Since all of the test cases have been given content I could simply remove this and did just that.

src/core/Chip/Chip/Chip.tsx
ketsappi: These seems to be better now, not like as commented but close enough to be clearer.

src/core/Chip/BaseChip/BaseChip.tsx
```diff
@@ -0,0 +1,19 @@
+import { ReactNode } from 'react';
+
+export const baseClassName = 'fi-chip';
+export const chipClassNames = {
+  disabled: `${baseClassName}--disabled`,
+  icon: `${baseClassName}--icon`,
+  content: `${baseClassName}--content`,
+  removable: `${baseClassName}--removable`,
+  button: `${baseClassName}--button`,
```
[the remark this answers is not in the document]
LJKaski: Moved the button variant specific ones to Chip.

ketsappi (approved):

aappoalander (approved):

PR #446, opened 2021-03-22
[Feature] Button secondary noborder styles and button link variant

LJKaski:
## Description
This PR replaces the button tertiary variant with link variant. It also adjusts the button secondary noborder styles to work on different backgrounds. To make these changes, suomifi-design-tokens dependency was updated to get the recently updated color token values.

## Motivation and Context
The implementation differed from the designs, and there were requests from the projects using the library to make button secondary noborder compatible with other background colors than white.

## How Has This Been Tested?
Automated tests and running locally in Windows using Chrome, Firefox and Edge.

secondary noborder hover styles with colored background
image

### Breaking changes:
* Button tertiary variant replaced by link button variant
* Button secondary-noborder variant renamed to secondaryNoBorder
* depthSecondaryDark1 color token value changed
### Other changes:
* suomifi-design-tokens dependency updated to latest available
* Button secondaryNoBorder variant hover styles changed

### DIFF
--- a/src/core/Button/Button.baseStyles.tsx
+++ b/src/core/Button/Button.baseStyles.tsx
@@ -75,15 +75,23 @@ const linkStyles = css`
   &.fi-button--link {
     color: ${theme.colors.highlightBase};
     ${secondary}
-    background: ${theme.gradients.highlightLight3ToHighlightLight2};
+    background: ${theme.gradients.depthSecondaryToDepthSecondaryDark1};
     border: none;
 
     &:hover {
-      background: ${theme.colors.highlightLight3};
+      background: ${theme.gradients.highlightLight4ToDepthSecondary};
     }
 
     &:active {
       background: ${theme.gradients.depthLight3ToDepthLight2};
+    }
+
+    &.fi-button--disabled,
+    &[disabled],
+    &:disabled {
+      color: ${theme.colors.depthBase};
+      background: none;
+      background-color: ${theme.colors.depthLight3};
     }
   }
 `;
--- a/src/core/Button/Button.baseStyles.tsx
+++ b/src/core/Button/Button.baseStyles.tsx
@@ -64,7 +64,7 @@ const secondaryStyles = css`
 `;
 
 const secondaryNoBorderStyles = css`
-  &.fi-button--secondaryNoBorder {
+  &.fi-button--secondary-noborder {
     ${secondary}
     border: none;
     background-color: transparent;
--- a/src/core/Button/Button.baseStyles.tsx
+++ b/src/core/Button/Button.baseStyles.tsx
@@ -64,7 +64,7 @@ const secondaryStyles = css`
 `;
 
 const secondaryNoBorderStyles = css`
-  &.fi-button--secondary-noborder {
+  &.fi-button--secondaryNoBorder {
     ${secondary}
     border: none;
     background-color: transparent;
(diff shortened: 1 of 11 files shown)


src/core/Button/Button.tsx
```diff
@@ -10,12 +10,12 @@ type ButtonVariant =
   | 'inverted'
   | 'secondary'
   | 'secondary-noborder'
```
aappoalander, line 12: I think this might actually be bad practice / not according to our conventions to have the variant name this way. Should it be in lowerCamelCase?

src/core/theme/colors.ts
```diff
@@ -106,9 +106,13 @@ export const gradients = {
   whiteBaseNegative: `linear-gradient(-180deg, ${alphaHex(0.1)(
     colors.whiteBase,
   )} 0%, ${alphaHex(0)(colors.whiteBase)} 100%)`,
-  whiteBaseToDepthLight2: `linear-gradient(0deg, ${colors.depthLight2} 0%, ${colors.whiteBase} 100%)`,
+  whiteBaseToDepthLight2: `linear-gradient(-180deg, ${alphaHex(1)(
```
aappoalander, line 109: Is the alphaHex necessary for the first color as it has no transparency?

src/core/Modal/Modal/__snapshots__/Modal.test.tsx.snap
```diff
@@ -641,7 +641,7 @@ exports[`Basic modal should match snapshot 1`] = `
 }
 
 .c8.fi-button--secondary:hover {
-  background: linear-gradient(0deg,hsl(202,7%,93%) 0%,hsl(0,0%,100%) 100%);
+  background: linear-gradient(-180deg,#ecedee 0%,rgba(255,255,255,0) 100%);
```
aappoalander, line 644: I assume this hsl/hex conversion comes from using alphaHex?

src/core/Button/Button.tsx
aappoalander: Or do we have similar variant names somewhere else?

LJKaski: That's the original name from 2 years ago. I don't think we have kebab-case names elsewhere. Should probably update this one if we are breaking things anyway.

LJKaski: As discussed elsewhere, I renamed the variant to secondaryNoBorder to put it in line with the general naming policy.

src/core/theme/colors.ts
LJKaski: Doesn't seem to be. Removed it 👍

src/core/Modal/Modal/__snapshots__/Modal.test.tsx.snap
LJKaski: Seems to be. Only tokens are used in the code.

src/core/Button/Button.baseStyles.tsx
```diff
@@ -64,25 +64,35 @@ const secondaryStyles = css`
 `;
 
 const secondaryNoBorderStyles = css`
-  &.fi-button--secondary-noborder {
+  &.fi-button--secondaryNoBorder {
```
aappoalander, line 67: this should probably be `fi-button--secondary-no-border` or `--secondary-noborder`

aappoalander: Seems to be fixed!

aappoalander (approved):

PR #447, opened 2021-03-23
Fix/chip props and styles

aappoalander:
## Description
This PR fixes Chip remove icon alignment issue with Safari and removes support for as-prop. Examples were updated and improved.

Also minor refactoring of styles was done to have better isolation between variants.

## Related Issue
The remove icon for Chip was not aligned according designs in Safari. As-prop made it possible to replace the chip button/div component with other elements which is not desired for this component.

## Motivation and Context
Components should look and behave the same way with all supported browsers. This fix is needed to align styles in Safari with rest of the supported browsers. Also changing the chip semantics with as-prop is not desired for this component.

## How Has This Been Tested?
Tested with CRA-TS project using MacOS with Safari, Firefox and Chrome.

Screenshot 2021-03-22 at 14 31 31
Screenshot 2021-03-23 at 9 27 37

## Release notes
- Chip
  - Breaking changes: as prop is now removed
  - Remove icon is now properly aligned in Safari

### DIFF
--- a/src/core/Chip/Chip/Chip.md
+++ b/src/core/Chip/Chip/Chip.md
@@ -24,11 +24,15 @@ const chipStyle = {
     </Chip>
     <Chip
       style={chipStyle}
+      removable
       actionLabel="Log referenced element"
       ref={exampleRef}
-      onClick={() => console.log(exampleRef.current)}
+      onClick={() => {
+        console.log(exampleRef.current);
+        removeAction();
+      }}
     >
-      Chip with ref
+      Removable chip with ref
     </Chip>
   </div>
   <div>
--- a/src/core/Chip/Chip/Chip.md
+++ b/src/core/Chip/Chip/Chip.md
@@ -29,7 +29,7 @@ const chipStyle = {
       ref={exampleRef}
       onClick={() => console.log(exampleRef.current)}
     >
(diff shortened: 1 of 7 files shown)


src/core/Chip/Chip/Chip.md
```diff
@@ -7,24 +7,30 @@ const removeAction = () => {
 };
 
 const exampleRef = React.createRef();
-
+const chipStyle = {
+  display: 'inline-block',
+  marginRight: '10px',
+  marginBottom: '24px'
```
ketsappi, line 13: Should we remove the bottom margin to keep the look more consistent. Or is there some specific test purpose for this?

ketsappi (approved):
Worked with Safari and Brave when tested.

No more weird behaviour on the Safari with the close icon as it was before the fix.

src/core/Chip/Chip/Chip.md
ketsappi: As discussed, this was to demonstrate the need to use margins by the library user. Margins are not provided per component.

LJKaski (dismissed):
Looks to be working well on windows + chrome, FF and Edge.

LJKaski (approved):

PR #450, opened 2021-03-25
[Fix] Modal mouse event issue

aappoalander:
## Description
This PR removes the fix for Modal NVDA cursor escaping issue as it proved to be not working as expected. In addition, small improvements for styles and component structure.

## Related Issue
NVDA fix prevented normal focusing on items inside the modal with mouse.

## Motivation and Context
The fix for NVDA cursor escaping the modal with events bubbling through the modal background was made as a last minute improvement and was not properly tested. Preventing mouse interactions with modal background overlay resulted in undesired side effect and blocked most mouse interactions inside the modal also.

Removing the fix will re-introduce the cursor escaping issue where a user is clicking non interactive elements on the modal dialog with mouse and NVDA on. This allows the cursor to escape to the underlying DOM elements and there seems to be no easy fix for this behaviour (other than complex solutions discouraged by W3C.).

## How Has This Been Tested?
Tested with Styleguidist on MacOS and Windows using NVDA/VoiceOver with Chrome, Firefox, Safari and Edge.

## Release notes
- Modal
  - Remove non-working NVDA cursor escape fix and improve styles

### DIFF
--- a/src/core/Modal/Modal/Modal.baseStyles.tsx
+++ b/src/core/Modal/Modal/Modal.baseStyles.tsx
@@ -51,6 +51,7 @@ export const baseStyles = css`
 
     &--small-screen {
       & .fi-modal_content-container {
+        border-radius: 0;
         width: 100%;
         height: 100%;
         max-height: 100%;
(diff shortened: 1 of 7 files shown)


LJKaski (approved):
Looks good to me :)

ketsappi (approved):

PR #452, opened 2021-03-29
[Fix] ExpanderGroup performance improvements

aappoalander:
## Description
This PR increases ExpanderGroup performance especially for use cases with large amounts of expanders in one expander group.

In addition, this PR fixes small font style issues.

## Motivation and Context
ExpanderGroup was setting state for each Expander close or open event and sharing that state via context. This was causing redundant renders and change detection for all nested Expanders and also unnecessary expander open state comparisons. This caused ExpanderGroups with hundreds or more Expanders to become unusable.

In addition, there was a bug with font-family style which resulted into invalid css being built. 

## How Has This Been Tested?
Tested on MacOS with Safari, Edge, Chrome and Firefox with VoiceOver. Test environment was React Styleguidist build, Create React App TypeScript build and SSR test build. Performance was tested with 500 expanders.

## Release notes
- ExpanderGroup
  - Performance improvements
 - Expander
   - Breaking change, undocumented consumer prop removed from public API
   - Style fixes for fonts

### DIFF
--- a/src/core/Expander/Expander/Expander.tsx
+++ b/src/core/Expander/Expander/Expander.tsx
@@ -56,7 +56,6 @@ export interface ExpanderProps {
   open?: boolean;
   /** Event handler to execute when clicked */
   onOpenChange?: (open: boolean) => void;
-  consumer?: ExpanderGroupProviderState;
 }
 
 export interface ExpanderTitleBaseProps {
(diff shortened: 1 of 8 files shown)


ketsappi (approved):

LJKaski (dismissed):
Looks good and seems to work well as far as I can tell.

ketsappi (approved):

PR #454, opened 2021-03-30
[Fix] Move Heading to core

LJKaski:
## Description
This PR removes component level implementation of the Heading component and moves all logic to core. It also addresses a documentation issue as well as an issue where as and asProp could both be given to the component and could potentially clash.

In addition, static variants were removed and replaced in tests and examples with using the variant prop.

## Motivation and Context
Components level is no longer useful and is being removed from the library as a whole. Static variants are also being removed in general. The smaller issues were bugs that needed to be fixed in any case.

## How Has This Been Tested?
Tested with automated tests as well as running locally in styleguidist in Chrome, Edge and Firefox.

### Heading
* **Breaking changes:**
  * Static variants removed. Use variant prop instead.
  * 'asProp' prop removed. Use 'as' instead.
* Other changes:
  * Fixed a bug where API descriptions were in wrong order


src/components/Heading/Heading.tsx
```diff
@@ -1,25 +0,0 @@
-import React, { Component } from 'react';
-import classnames from 'classnames';
-import { HtmlH, HtmlHProps, hLevels } from '../../reset/HtmlH/HtmlH';
-export { hLevels };
-
-export interface HeadingProps extends HtmlHProps {
-  /** Change HTML-element */
-  variant: hLevels;
-  className?: string;
```
aappoalander, line 9: Should we have className explicitly mentioned in the core level API as it was also here even if already provided by HtmlHProps?

LJKaski: Might be a good idea. Should we also document the 'as' prop since it's used in the examples (Sami's comment from elsewhere)?

aappoalander: Looks good!

aappoalander (approved):

PR #455, opened 2021-04-09
[Feature] Update documentation

LJKaski:
## Description
This PR adds mention of supported browsers and setups we test on. It also updates the content of README.md, DEVELOPMENT.md, CONTRIBUTING.md and FAQ.md to match the current state of the library.

## Motivation and Context
The documentation was partly out of date and needed some cleaning up and updating.

## How Has This Been Tested?
Checked that the previews looked ok.

## Release notes
- Documentation updated


DEVELOPMENT.md
```diff
@@ -20,34 +26,30 @@ After cloning suomifi-ui-components, run `yarn` to fetch its dependencies. Then,
...
 
 ## Source
 
-[![suomifi-ui-components architecture diagram](suomifi-ui-components.png 'suomifi-ui-components architecture')](suomifi-ui-components.png)
-
 Source contains 3 stages of components:
 
 1. `reset` is for resetting html tags (don't do too opinionated stuff here).
-2. `components` are accessible/a11y version of components and HTML-semantics without Suomi.fi related.
+2. `components` are accessible/a11y version of components and HTML-semantics without Suomi.fi related. The components level is being removed, and is not supported any longer.
```
aappoalander, line 36: As the components level is not even provided by the build I would document this a bit differently. e.g. `components` are bare bones accessibility implementations without Suomi.fi styles. This abstraction is being removed and will not be used in the future.

DEVELOPMENT.md
```diff
@@ -20,34 +26,30 @@ After cloning suomifi-ui-components, run `yarn` to fetch its dependencies. Then,
...
+**Do not create duplication of source or styles for component.**
 
 Export interfaces for exported functions/components. Typescript will generate declaration files from exported interfaces (.d.ts). Write comments/documentatiion to all properties that need to be shown at styleguide.
 
 ### Props that are passed on to another component
 
 ```jsx
-// put rest of the props to passProps, but takeAwayProp and takeAanother
-const { takeAwayProp, takeAanother: renamedAsProp, ...passprops } = withSuomifiDefaultProps(this.props);
+// put rest of the props to passProps, but takeAwayProp and takeAnother
```
aappoalander, line 51: => destructure takeAway and takeAnother props and put rest of the props to passProps.

DEVELOPMENT.md
LJKaski: That sounds a lot better. I'll replace that. Thanks :)

DEVELOPMENT.md
LJKaski: That is clearer, I'll replace it. 👍

README.md
```diff
@@ -1,19 +1,23 @@
-# [<img src="https://avatars0.githubusercontent.com/u/11345641?s=88&v=4" alt="VRK" width="18"/> suomifi-ui-components](https://vrk-kpa.github.io/suomifi-ui-components/)
+# [<img src="https://avatars0.githubusercontent.com/u/11345641?s=88&v=4" alt="DVV" width="18"/> suomifi-ui-components](https://vrk-kpa.github.io/suomifi-ui-components/)
+
+![npm](https://img.shields.io/npm/v/suomifi-ui-components)
+
+![NPM](https://img.shields.io/npm/l/suomifi-ui-components)
+
+![WCAG](https://img.shields.io/badge/WCAG%202.1-AA-green)
```
ketsappi, line 7: Badges would be great to have on the same line. Like this, by separating them only with space-character.

npm NPM WCAG

DEVELOPMENT.md
```diff
@@ -89,32 +82,23 @@ styled(Component)`
...
-```jsx
-styled(props => <Component {...props} />)`
-  ${props => baseStyles(props)}
-`;
-```
-
 ## Styling
 
-`src/components` will provide 'fi-' -prefixed classNames to all components - and then at `src/core` styling is based on those base classNames.
+components will have ’fi-’ -prefixed class names and styling is based on those class names.
```
ketsappi, line 87: Maybe change to different backticks or something as now it is not looking correct:

Now it looks like this:
```
’fi-’
```

README.md
LJKaski: Good idea. I'll do that 👍

CONTRIBUTING.md
```diff
@@ -1,45 +1,43 @@
...
+
+1. Fork the repository and create your branch from proper branch.
+
+2. Run `yarn` in the repository root.
+
+3. If you’ve fixed a bug or added code that should be tested, add tests!
+
+4. Ensure the test suite passes (`yarn test`)
+
+5. Run `yarn test -- -u` to update jest snapshots. Commit any changes.
```
ketsappi, line 23: Just tested this to be sure and it seems that it is enough to just write:
```bash
yarn test -u
```

With longer version it works, but you also get a warning like this:

```
warning From Yarn 1.0 onwards, scripts don't require "--" for options to be forwarded. In a future version, any explicit "--" will be forwarded as-is to the scripts.
```

README.md
```diff
@@ -109,7 +104,9 @@ See [FAQ.md](/FAQ.md).
 
 See [DEVELOPMENT.md](/DEVELOPMENT.md).
 
-## 🤝 Contributing [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat-square)](http://makeapullrequest.com)
+## 🤝 Contributing
+
+[![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg?style=flat-square)](http://makeapullrequest.com)
```
ketsappi, line 105: As we now have the badges in the start, it might be nice to move this "PRs welcome" also there.

DEVELOPMENT.md
LJKaski: Fixed 👍

README.md
LJKaski: Good idea. Should we keep it here as well or just have it at the top?

ketsappi: I think that only at the top is enough.

ketsappi: And.. could you remove the following from the definition of this badge as it is having different style 😅 
```
?style=flat-square
```

LJKaski: you have a keen eye. I didn't even notice that in the preview :D

ketsappi: Thanks! Now it looks purrrfect 😸

ketsappi (approved):

aappoalander (approved):

aappoalander (approved):

ketsappi (approved):

PR #456, opened 2021-04-12
[Fix] Text component color prop

LJKaski:
## Description
This PR fixes the broken color prop in the Text component. The color prop was not passed on or utilized in any way after the recent tokens revamp.

## Motivation and Context
This was an obvious bug that needed to be fixed asap.

## How Has This Been Tested?
Tested by running locally in styleguidist in Chrome.

image

### Text:
* Fix broken color prop


ketsappi (approved):
Colour prop seems seems to work now. 👍

aappoalander (approved):

PR #457, opened 2021-04-12
Fix/modal mobile focus lock

aappoalander:
## Description
This PR introduces fixes for Modal for browsers not properly supporting aria-dialog property. Sibling DOM nodes are now aria-hidden when the modal is open and parent nodes have a role of presentation.

## Related Issue
Previously some browser and screenreader combinations allowed navigating to elements hidden beneath the Modal. This should no longer be possible.

There are some caveats though, an existing issue with iOS and Safari allows screenreader to read number of pages from the main document below the modal. This issue is reproducible with many other library implementation as well and there seems to be no feasible workaround. The bugs have been reported to Apple and Material-UI.

## Motivation and Context
Android and iOS users should be supported as those with Safari and Chrome form the majority of mobile users using a screen reader. Most of the issues (excluding the above mentioned) have now been fixed.

## How Has This Been Tested?
Tested using styleguidist build with Mac OS, Windows, iOS and Android with Chrome, Safari, Edge and Firefox.

## Release notes
- Fixes for Modal Accessibility with mobile devices

### DIFF
--- a/src/core/Modal/Modal/Modal.tsx
+++ b/src/core/Modal/Modal/Modal.tsx
@@ -137,6 +137,31 @@ class BaseModal extends Component<ModalProps> {
     }
   };
 
+  private getNodePrevState = (node: HTMLElement): null | PrevState => {
+    const ariaDisabledPrevState = node.getAttribute('aria-hidden');
+    const rolePrevState = node.getAttribute('role');
+    const prevState: PrevState = {};
+    if (
+      typeof ariaDisabledPrevState === 'string' &&
+      ariaDisabledPrevState.length > 0
+    ) {
+      prevState['aria-hidden'] = ariaDisabledPrevState;
+    }
+    if (typeof rolePrevState === 'string' && rolePrevState.length > 0) {
+      prevState.role = rolePrevState;
+    }
+    return Object.keys(prevState).length > 0 ? prevState : null;
+  };
+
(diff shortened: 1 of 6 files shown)


LJKaski (dismissed):
Seems to work as intended 👍

ketsappi (dismissed):
Tested with iOS VoiceOver and seems to work nicely :)

src/core/Modal/Modal/Modal.tsx
```diff
@@ -137,6 +137,31 @@ class BaseModal extends Component<ModalProps> {
     }
   };
 
+  private getNodePrevState = (node: HTMLElement): null | PrevState => {
+    const ariaDisabledPrevState = node.getAttribute('aria-hidden');
```
LJKaski, line 141: Why is this named ariaDisabledPrevState yet it containts aria-hidden attribute value? Why not make it ariaHiddenPrevState to be explicit?

LJKaski (dismissed):
Works fine for me :)

src/core/Modal/Modal/Modal.tsx
aappoalander: Should not be disabled. Due to original implementation where it covered both aria-hidden and role presentation. Will fix.

aappoalander: Fixed

LJKaski (approved):

ketsappi (approved):

PR #459, opened 2021-04-19
Fix/expander styles and aria support

aappoalander:
## Description
This PR improves Expander accessibility features and fixes minor style issues.

ExpanderGroup now has aria-expanded prop for open/close all button. In addition, ExpanderGroup props now follow the naming conventions.

ExpanderTitleButton no longer supports custom aria attributes explicitly due to those being redundant when used with aria-expanded property. Standard aria props can still be used via toggleButtonProps. 

ExpanderTitle now requires providing a title id for the toggle button to make this variant more accessible for screen readers. With this title variant the toggle button does not contain any title information and could be difficult to understand e.g. when using button navigation mode with a screen reader. By adding the title id as aria-describedby for the toggle button solves this issue.

ExpanderContent height and overflow styles are now fixed and should work better in some contexts.

## Related Issue
Some improvements are based on results from accessibility testing by Avaava while others are based on reports from Suomi.fi services.

## Motivation and Context
Expander and related components had some accessibility and style issues. These changes were necessary to fix the issues.

## How Has This Been Tested?
Tested with MacOS using VoiceOver with Safari, Chrome, Firefox and Edge, with Android using Talkback with Chrome and with iOS using Safari.

## Release notes
- ExpanderGroup
  - **Breaking change:** Rename props OpenAllText, AriaOpenAllText, CloseAllText, AriaCloseAllText => openAllText, ariaOpenAllText, closeAllText, ariaCloseAllText,
  - Add aria-expanded prop for Open/Close all button
- ExpanderTitleButton
  - **Breaking change:** remove ariaOpenText and ariaCloseText props
- ExpanderTitle:
  - **Breaking change:** Add required toggleButtonAriaDescribedBy prop
- ExpanderContent:
  - Fix expander content height and overflow styles

### DIFF
--- a/src/core/Expander/Expander/Expander.md
+++ b/src/core/Expander/Expander/Expander.md
@@ -44,10 +44,10 @@ import {
 } from 'suomifi-ui-components';
 
 <ExpanderGroup
-  OpenAllText="Open all"
-  AriaOpenAllText="Open all expanders"
-  CloseAllText="Close all"
-  AriaCloseAllText="Close all expanders"
+  openAllText="Open all"
+  ariaOpenAllText="Open all expanders"
+  closeAllText="Close all"
+  ariaCloseAllText="Close all expanders"
 >
   <Expander>
     <ExpanderTitleButton>Test expander 1</ExpanderTitleButton>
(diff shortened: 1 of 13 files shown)


src/core/Expander/Expander/Expander.md
```diff
@@ -25,8 +25,11 @@ import {
...
     ariaCloseText="close expander"
+    toggleButtonAriaDescribedBy="checkbox-id"
   >
-    <Checkbox hintText="Checkbox hint text">Checkbox label</Checkbox>
+    <Checkbox hintText="Checkbox hint text" id="checkbox-id">
+      Checkbox label
+    </Checkbox>
   </ExpanderTitle>
   <ExpanderContent>Test expander</ExpanderContent>
 </Expander>;
```
ketsappi, line 35: It seems that both examples using `ExpanderGroup` still uses the old, capitalised, prop names.

aappoalander: Good catch! Damn I hate these md files without proper linter support.

aappoalander: These should be now fixed!

ketsappi (approved):

LJKaski (approved):
Aside from a small aria hickup with NVDA + FF outside of our control looks to be working well 👍

PR #460, opened 2021-04-22
Fix/modal initial focus and width

aappoalander:
## Description
This PR fixes Modal component initial focus to be in the heading instead of the first tabbable element unless a specific element ref is provided for focusing after opening the Modal. In addition, this PR introduces changes to styles so, that it's easier to modify the size of the Modal without relying on implementation details.

Now it's possible to use `<Modal className="custom-modal"` with `.fi-modal.custom-modal.custom-modal { width: '1000px'` and `.fi-modal--smallScreen.custom-modal.custom-modal { width: '100%'`.

## Related Issue
Changes are based on feedback from stakeholders and accessibility testing.

## Motivation and Context
Modal contents other than confirm buttons were not communicated for screen reader users in a proper manner. Having initial focus on the title should resolve the issue.

It was also difficult to customize the modal size due to style structure and css specificity issues. With restructured styles things should now be more simple. 

## How Has This Been Tested?
Tested with Styleguidist build, CRA TS project, SSR test app
- MacOS and VoiceOver using Safari, Chrome, Firefox and Edge
- Android and TalkBack using Chrome

## Release notes
- Modal
 - **Breaking change:** custom className is now passed onto outermost content container instead of component root
 - Initial focus is now in title

### DIFF
--- a/src/core/Modal/Modal/Modal.tsx
+++ b/src/core/Modal/Modal/Modal.tsx
@@ -17,7 +17,7 @@ export interface ModalProps
   extends Omit<HtmlDivProps, 'children' | 'className' | 'title' | 'ref'> {
   /** Show or hide the modal */
   visible: boolean;
-  /** Modal container div class name for custom styling. */
+  /** Modal container div classname for custom styling. */
   className?: string;
   /** Children */
   children: ModalContent | ModalFooter | ReactNode;
@@ -74,8 +74,9 @@ const {
 
 export const baseClassName = 'fi-modal';
 const modalClassNames = {
-  noPortal: `${baseClassName}--no-portal`,
   smallScreen: `${baseClassName}--small-screen`,
+  base: `${baseClassName}_base`,
+  noPortal: `${baseClassName}_base--no-portal`,
   overlay: `${baseClassName}_overlay`,
   contentContainer: `${baseClassName}_content-container`,
   content: `${baseClassName}_content`,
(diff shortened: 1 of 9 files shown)


src/core/Modal/Modal/Modal.tsx
```diff
@@ -255,6 +256,7 @@ class BaseModal extends Component<ModalProps> {
     const {
       visible,
       id,
+      propClassName,
```
ketsappi, line 259: Could we think some more describing name for this `propClassName` to explain it's purpose?
As it does not have JSDoc where to check what it does.

aappoalander: What would be more descriptive name? This the className prop provided by the user, but as it is not passed onto the outermost element in the Base component, it needs to be renamed. The Modal component cannot pass className prop to StyledModal, because it would result styled components to add it's own 'hidden' classNames to it. Then the 'hidden' classNames would not end up to the outermost div and styles would break.

The other way would be to use InternalModalProps and have the prop there, but in my opinion it was less elegant solution - this is debatable of course.

ketsappi: Based on that explanation, it might hard to give it a describing name. Then the less elegant internal props idea might be good; to have that long explanation for it.
The `Internal*Props` seems to be used in quite many other components so it might be a viable solution.

aappoalander: Sure, I can add internal props. Just seemed too much additional boilerplate in my opinion, but if the purpose of the prop is not clear from the naming, it's probably needed.

aappoalander: Added internal props with explanation.

ketsappi (approved):

LJKaski (dismissed):
Looks good! Good work :)

ketsappi (approved):

PR #461, opened 2021-04-23
[Fix] Form element status accessibility

LJKaski:
## Description
This PR makes status text an aria-live region, so screen reader users get information of status changes immediately. It also implements this change into form components using StatusText.

## Motivation and Context
The lack of information about component status changes was something that has come up in the recent accessibility clinics and needed to be addressed.

## How Has This Been Tested?
Tested with NVDA + chrome + FF and Talkback + Chrome.

### TextInput, TextArea, Checkbox, SearchInput
* Status text now has `aria-live="assertive"` by default. This can be changed via `statusTextAriaLiveMode` prop.

### DIFF
--- a/src/core/Form/StatusText/StatusText.test.tsx
+++ b/src/core/Form/StatusText/StatusText.test.tsx
@@ -10,6 +10,10 @@ describe('props', () => {
       const { container } = render(<StatusText>Test text</StatusText>);
       expect(container.firstChild).toHaveTextContent('Test text');
     });
+    it('renders element even without content', () => {
+      const { container } = render(<StatusText />);
+      expect(container.children.length).toEqual(1);
+    });
   });
 
   describe('id', () => {
--- a/src/core/Form/StatusText/StatusText.test.tsx
+++ b/src/core/Form/StatusText/StatusText.test.tsx
@@ -38,6 +38,24 @@ describe('props', () => {
       expect(container.firstChild).toHaveClass('fi-status-text--error');
     });
   });
+
+  describe('aria-live', () => {
+    it('should have the specified aria-live attribute', () => {
+      const { container } = render(
+        <StatusText ariaLiveMode="assertive">Test text</StatusText>,
+      );
+      expect(container.firstChild).toHaveAttribute('aria-live', 'assertive');
+    });
+
+    it('should not have aria-live attribute when disabled', () => {
+      const { container } = render(
+        <StatusText disabled ariaLiveMode="assertive">
+          Test text
+        </StatusText>,
+      );
+      expect(container.firstChild).not.toHaveAttribute('aria-live');
+    });
+  });
 });
 
 test(
--- a/src/core/Form/StatusText/StatusText.test.tsx
+++ b/src/core/Form/StatusText/StatusText.test.tsx
@@ -44,7 +44,7 @@ describe('props', () => {
   });
 
   describe('aria-live', () => {
-    it('should have aria-live=assertive as default even when empty', () => {
+    it('should have given aria-live attribute even when empty', () => {
       const { container } = render(<StatusText ariaLiveMode="assertive" />);
       expect(container.firstChild).toHaveAttribute('aria-live', 'assertive');
     });
--- a/src/core/Form/StatusText/StatusText.test.tsx
+++ b/src/core/Form/StatusText/StatusText.test.tsx
@@ -10,11 +10,6 @@ describe('props', () => {
       const { container } = render(<StatusText>Test text</StatusText>);
       expect(container.firstChild).toHaveTextContent('Test text');
     });
-
-    it('is null when no children given', () => {
-      const { container } = render(<StatusText />);
-      expect(container.firstChild).toEqual(null);
-    });
   });
 
   describe('id', () => {
@@ -41,13 +36,6 @@ describe('props', () => {
         <StatusText status="error">Test text</StatusText>,
       );
       expect(container.firstChild).toHaveClass('fi-status-text--error');
-    });
-  });
-
-  describe('disabled', () => {
-    it('is null when disabled', () => {
-      const { container } = render(<StatusText disabled>Test text</StatusText>);
-      expect(container.firstChild).toEqual(null);
     });
   });
 });
--- a/src/core/Form/StatusText/StatusText.test.tsx
+++ b/src/core/Form/StatusText/StatusText.test.tsx
@@ -44,20 +44,24 @@ describe('props', () => {
   });
 
   describe('aria-live', () => {
-    it('should have the specified aria-live attribute', () => {
-      const { container } = render(
-        <StatusText ariaLiveMode="assertive">Test text</StatusText>,
-      );
+    it('should have aria-live=assertive as default even when empty', () => {
+      const { container } = render(<StatusText ariaLiveMode="assertive" />);
       expect(container.firstChild).toHaveAttribute('aria-live', 'assertive');
     });
+    it('should have the specified aria-live attribute', () => {
+      const { container } = render(
+        <StatusText ariaLiveMode="polite">Test text</StatusText>,
+      );
+      expect(container.firstChild).toHaveAttribute('aria-live', 'polite');
+    });
 
-    it('should not have aria-live attribute when disabled', () => {
+    it('aria-live should be off when disabled', () => {
       const { container } = render(
         <StatusText disabled ariaLiveMode="assertive">
           Test text
         </StatusText>,
       );
-      expect(container.firstChild).not.toHaveAttribute('aria-live');
+      expect(container.firstChild).toHaveAttribute('aria-live', 'off');
     });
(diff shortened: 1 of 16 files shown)


src/utils/aria.ts
```diff
@@ -5,6 +5,8 @@ export const ariaLabelOrHidden = (ariaLabel?: string) =>
     ? { 'aria-label': ariaLabel, role: 'img' }
     : { 'aria-hidden': true };
 
+export type ariaLiveModes = 'assertive' | 'polite' | 'off';
```
aappoalander, line 8: We should probably name types using these conventions: https://basarat.gitbook.io/typescript/styleguide#type

And should use PascalCase for these?

src/core/Form/StatusText/StatusText.test.tsx
```diff
@@ -10,11 +10,6 @@ describe('props', () => {
       const { container } = render(<StatusText>Test text</StatusText>);
       expect(container.firstChild).toHaveTextContent('Test text');
     });
-
-    it('is null when no children given', () => {
```
aappoalander, line 14: Should there still be a test to check that it renders correctly even without children?

src/core/Form/StatusText/StatusText.test.tsx
```diff
@@ -43,13 +38,6 @@ describe('props', () => {
       expect(container.firstChild).toHaveClass('fi-status-text--error');
     });
   });
-
-  describe('disabled', () => {
-    it('is null when disabled', () => {
```
aappoalander, line 48: Same here, maybe we should check that the aria live is not enabled when a component is disabled?

src/core/Form/StatusText/StatusText.tsx
```diff
@@ -24,10 +24,17 @@ export interface StatusTextProps extends HtmlSpanProps {
...
 const StyledStatusText = styled(
-  ({ className, children, status, ...passProps }: StatusTextProps) => (
+  ({
+    className,
+    children,
+    disabled,
+    status,
+    ...passProps
+  }: StatusTextProps) => (
     <HtmlSpan
```
aappoalander, line 34: I would check here if the disabled prop is true and remove aria-live if not needed.

src/utils/aria.ts
ketsappi: So the suggested type would be after changes like this, if I'm on the same line:
`export type AriaLiveMode = 'assertive' | 'polite' | 'off';`

- Using PascalCase
- No plural in name

src/core/Form/StatusText/StatusText.tsx
```diff
@@ -24,10 +24,17 @@ export interface StatusTextProps extends HtmlSpanProps {
 }
```
ketsappi, line 25: ### This might be over abstraction and not needed

As the same `statusTextAriaLiveMode` and `statusText` prop is repeated over and over what about if we change it to separate type?

```tsx
// src/core/Form/types.ts
export interface StatusText {
  /** Status text to be shown below the component and hint text. Use e.g. for validation error */
  statusText?: string;

  /** Aria-live mode for the status text element
   * @default assertive
   */
  statusTextAriaLiveMode?: ariaLiveModes;
}
```

Then just extending that type instead, `SearchInput` as an example:
```tsx
// src/core/Form/SearchInput/SearchInput.tsx
export interface SearchInputProps
  extends Omit<
      HtmlInputProps,
      | 'type'
      | 'disabled'
      | 'onChange'
      | 'children'
      | 'onClick'
      | 'value'
      | 'defaultValue'
    >,
    StatusText {
	//...
}
```

aappoalander: I think that's a good idea. Assuming it fits all use cases, which in this case I think it does. Should not be too big of a task to change. But I think we still need to export the granular types like AriaLiveMode as well.

src/utils/aria.ts
LJKaski: Already changed it to PascalCase, but will remove the plural too.

src/core/Form/StatusText/StatusText.tsx
LJKaski: I like the idea as well. Maybe a thing for a separate ticket just to keep things clean on that front?

aappoalander: Fits here as well in my opinion. Would save a round of testing without affecting issues scope much. But up to author to decide.

LJKaski: That's fine as well :) I'll do it while I'm at it.

src/utils/aria.ts
```diff
@@ -5,6 +5,8 @@ export const ariaLabelOrHidden = (ariaLabel?: string) =>
     ? { 'aria-label': ariaLabel, role: 'img' }
     : { 'aria-hidden': true };
 
+export type AriaLiveMode = 'assertive' | 'polite' | 'off';
+
```
ketsappi, line 9: Nitpicking here, but as this is not an util it could be located to `src/core/Form/types.ts`

_On the other end, filename `types.ts` is also somehow misleading when the `interface StatusTextCommonProps` was added. But interface and type declarations seems to close in TypeScript so maybe we can live with this._

src/utils/aria.ts
LJKaski: Should now be in order.

src/core/Form/StatusText/StatusText.test.tsx
LJKaski: Added a test for this.

src/core/Form/StatusText/StatusText.test.tsx
LJKaski: This is now tested.

src/core/Form/StatusText/StatusText.tsx
LJKaski: This should now work as described.

src/core/Form/StatusText/StatusText.tsx
LJKaski: Implemented as `StatusTextCommonProps` interface.

src/utils/aria.ts
LJKaski: Moved the type to types.ts

src/core/Form/StatusText/StatusText.test.tsx
aappoalander: I'm wondering how we should handle this. It might be safer to keep aria-live attribute in DOM always when status text component is used. When there is no content, live mode should be off and same goes with disabled. Possible values are off, polite or assertive. This test should also check that the attribute exists rather than just checking there is children present.

src/core/Form/StatusText/StatusText.tsx
aappoalander: Reconsidering this to use `aria-live="off"` only when set with props or when component is disabled. Otherwise it should be always on, even when there is no content.

src/core/Form/StatusText/StatusText.test.tsx
aappoalander: See other comments for better explanation, but I think we should always keep aria-live attribute in DOM.

src/core/Form/StatusText/StatusText.tsx
LJKaski: That might be a better option. Not sure how differently the approaches would work in practice, but this one should be a cleaner solution at least 👍

src/core/Form/StatusText/StatusText.test.tsx
LJKaski: When there's no content, I don't see why it should be off. The way I think about it is that having an empty region with assertive aria-live mode means that it is ready to read out loud whatever content it gets. Turning it off while waiting for content doesn't really accomplish anything in my opinion. And for the test, do you mean we should add a test to check if an empty element has the aria-live attribute as well?

src/core/Form/StatusText/StatusText.tsx
LJKaski: It should now be 'off' when the element is disabled and on with the given mode in other cases - even when empty.

src/core/Form/StatusText/StatusText.test.tsx
LJKaski: Added tests to see that the element has an aria-live attribute even when empty and that the mode is 'off' when the element is disabled.

src/core/Form/StatusText/StatusText.test.tsx
LJKaski: Commented elsewhere. Should now be fixed.

ketsappi (approved):

ketsappi (approved):
Tested again after the rebase. Seems to work as intended.
Small hiccups with the Brave browser with aria-live as discussed elsewhere but we are not able to effect that.

aappoalander (approved):

PR #463, opened 2021-04-30
Feature/update deps

aappoalander:
## Description
Update dependencies to latest or latest supported version without conflicts. Including major updates to Reach UI (and related styles) and Husky. Minor updates and patches for tooling, types and other dependencies. 

## Related Issue
This upgrade should fix an existing iOS bug and some security issues with transitive dependencies.

Closes #
(potentially, needs further testing)
SFIDS-357
SFIDS-364

## Motivation and Context
Security needs to be up-to-date and keeping dependencies otherwise in the latest available version distributes updating tasks more evenly.

## How Has This Been Tested?
Tested with styleguidist, CRA TS and SSR projects using Chrome, Safari and Edge on MacOS and Safari on iOS.

## Release notes
- **Breaking changes:** Reach-ui peer dependency versions upgraded.


ketsappi (approved):
It now seems possible to use Dropdown with iOS VoiceOver as it should be.

ketsappi (approved):

PR #467, opened 2021-05-12
[Fix] Modal focus lock portals and React 17 support

aappoalander:
## Description
This PR replaces current Modal implementation with react-modal under the hood. This change was required to support React 17 and components using portals inside modal.

In addition, this PR fixes focus styles for modal title.

Doing this change has some side-effects :
- React-modal focus lock does not consider "non visible" components to be focusable which rules out the use of certain styles used elsewhere in the library, e.g. containing a focusable element in display: inline element leaves the component outside tab order. These should be fixed separately from this issue.
- Modal is now always rendered into portal
- passProps are no longer supported and ModalPros no longer extends div props.

## Motivation and Context
Focus lock didn't work well with React 17 and there is no timeline for a fix. Other low-level solutions did not support rendering into portals in a suitable way for library implementation. The only viable solution was to switch to higher level solution like react-modal. Implementing all from scratch would have been too much work and too much maintenance burden.

## How Has This Been Tested?
Tested using Styleguidist, CRA-TS project and SSR-application. Testing with Win, MacOS, Android and iOS using Chrome, Edge, Safari, Firefox, VoiceOver, NVDA and TalkBack

## Release notes
- Refactor Modal to use react-modal under the hood
- Fix ModalTitle focus styles
- **Breaking changes**: These changes can be considered braking for beta releases
  - props no longer extend div props
  - usePortal prop is removed
  - style and appElementId props added


src/core/Modal/Modal/Modal.tsx
```diff
@@ -1,24 +1,23 @@
...
 
 export type ModalVariant = 'smallScreen' | 'default';
 
-export interface ModalProps
-  extends Omit<HtmlDivProps, 'children' | 'className' | 'title' | 'ref'> {
+export interface ModalProps {
   /** Show or hide the modal */
   visible: boolean;
+  /** Application root element id for setting aria-hidden when necessary */
+  appElementId?: string;
```
aappoalander, line 16: Should this prop be mandatory? Implementation breaks if it's empty, but there is a test before passing it to the underlying react-modal. Might make testing more difficult if this is changed to mandatory.

ketsappi: I would vote for it to be mandatory, as we should guide the usage of the component with the props.
It's unfortunate if testing will then come more difficult, but hopefully there is another way to have it tested.

aappoalander: Our tests should mostly already support this, I was referring to testing when using this component in projects.

aappoalander: This is now mandatory prop. However, there is still a truthy check before applying it, because things break silently and more badly if appElement id has a bad value.

aappoalander: And now there is also a runtime check for the prop and a logger.error if the prop is not truthy.

ketsappi (dismissed):

LJKaski (dismissed):
Seems to work as it should :)

LJKaski (approved):

PR #471, opened 2021-05-24
[Fix] Component focusability in modal

aappoalander:
## Description
This PR fixes some component styles to enable focusing inside modal using tab-key. In addition, this PR introduces scroll padding for the ModalContent to ensure focused content fits into view. 

**NOTE: React-modal does not consider some inline elements to be focusable. To ensure focusability inside Modal, components must have non zero size, be visible, have tabindex of 1 or 0, must be of node type /input|select|textarea|button|object or a with either tabIndex or href attribute and cannot be disabled!**
(https://github.com/reactjs/react-modal/blob/master/src/helpers/tabbable.js)

## Motivation and Context
After refactoring the Modal to use react-modal under the hood, some components became impossible to focus using tab-key when placed inside modal. This was caused by react-modals white-listing strategy for focus lock. Changing the styles to comply with the white list rules (element type/role, tab index, styles visible and size possible to calculate), makes these components focusable when necessary.

Automatic scrolling when a component focused is focused is handled by browser and the default behaviour varies between different browsers. In some cases parts of the focused components remained outside the current view due to these scrolling strategies. Adding scroll-padding-bottom resolves this issue without manual focus or scroll position management.

## How Has This Been Tested?
Tested using Create React App TypeScript project, Styleguidist build and SSR-test application. Test setup with MacOS and Windows using Chrome, Safari and Firefox.

## Release notes
- ModalContent: Add scroll-padding-bottom and -top default styles
- Button: Remove tabIndex for disabled button
- Dropdown: Change default display css value to inline-block.
- ExpanderTitle: Change title content display css value to inline-block.
- TextInput: Change input element display to inline-block and width to 100% by fixing internal class name issue

### DIFF
--- a/src/core/Modal/Modal/Modal.md
+++ b/src/core/Modal/Modal/Modal.md
@@ -6,7 +6,7 @@ Small screen variant takes all available screen space.
 
 Aria props override the primary and secondary button labels for screenreaders and allow more detailed description.
 
-**NOTE:** React-modal does not consider some inline elements to be focusable. To ensure focusability inside Modal, components must have non-zero size, be visible, have tabindex of 1 or 0, must be of node type input, select, textarea, button, object or a with either tabIndex or href attribute and cannot be disabled! ([source](https://github.com/reactjs/react-modal/blob/827796d48e7d4c74b4362cf90955e162082ee46d/src/helpers/tabbable.js))
+**NOTE:** React-modal does not consider some inline elements to be focusable. To ensure focusability inside Modal, components must have non-zero size, be visible, have tabindex of 0 or greater, must be of node type input, select, textarea, button, object or a with either tabIndex or href attribute and cannot be disabled! ([source](https://github.com/reactjs/react-modal/blob/827796d48e7d4c74b4362cf90955e162082ee46d/src/helpers/tabbable.js))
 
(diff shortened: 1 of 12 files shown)


src/core/Modal/Modal/Modal.md
```diff
@@ -6,6 +6,8 @@ Small screen variant takes all available screen space.
 
 Aria props override the primary and secondary button labels for screenreaders and allow more detailed description.
 
+**NOTE:** React-modal does not consider some inline elements to be focusable. To ensure focusability inside Modal, components must have non-zero size, be visible, have tabindex of 1 or 0, must be of node type input, select, textarea, button, object or a with either tabIndex or href attribute and cannot be disabled! ([source](https://github.com/reactjs/react-modal/blob/827796d48e7d4c74b4362cf90955e162082ee46d/src/helpers/tabbable.js))
```
LJKaski, line 9: The tabIndex should probably be "0 or larger" instead of explicitly "1 or 0".

aappoalander: This should now be fixed.

LJKaski (approved):

ketsappi (approved):

PR #476, opened 2021-06-09
[Feature] Soft disable support for Chip

ketsappi:
## Description

Adds soft disable support for a Chip. Implementation is mimicked from the Button, where the same functionality exists.

## Motivation and Context

Soft disabled Chip was needed for currently in progress component, Combobox (MultiSelect). It enables screenreader users also to "see" selected but disabled Chips in the listing under the input.

## How Has This Been Tested?
Locally in styleguidist.

## Release notes

* Soft disable support for Chip (#476)

### DIFF
--- a/src/core/Chip/Chip/Chip.md
+++ b/src/core/Chip/Chip/Chip.md
@@ -47,5 +47,7 @@ const chipStyle = {
   </Chip>
 
   <Chip disabled>Disabled chip</Chip>
+
+  <Chip aria-disabled>Aria-disabled chip</Chip>
 </>;
 ```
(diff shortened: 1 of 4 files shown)


LJKaski (approved):
Looks good to me and seems to be working as it should. Good work :)

aappoalander:
Looks good, but I would also add a styleguidist example. Use cases where a mandatory selection is indicated with a disabled chip should probably use this feature to work better with screen readers.

aappoalander (approved):

aappoalander:
Looks good and seems to be working with supported setup on MacOS and Android.

PR #478, opened 2021-06-16
[Fix] VisuallyHidden to export props + move it to core

ketsappi:
## Description

VisuallyHidden was not exporting props and therefore those were not available to library's users. This is now fixed.
Also moved VisuallyHidden from `components` to `core` level.

## How Has This Been Tested?
Built locally and tested in in CRA-TS test project.

## Release notes

* VisuallyHiddenProps are now exported from the library.


aappoalander:
Tested with Styleguidist build and Brave with VoiceOver and with TalkBack and Chrome. Seems to be working as expected / without changes to behaviour. Props export tested with DS-site and seems to be also working.

LJKaski (approved):
Tested with NVDA + Chrome, FF and Edge and it seems to be working just fine :)

PR #483, opened 2021-06-29
[Fix] MultiSelect accessibility improvements 

ketsappi:
## Description

Following changes improves the accessibility of the MultiSelect component. These issues were noticed in the accessibility clinic.

- `hintText` prop for MultiSelect as optional prop. Useful for user to tell that the input field can be used to filter available options.
-  `ariaOptionsAvailableText` prop as required prop for MultiSelect. It will be used to tell for screen reader user how many options are available after filtering.
-  `ariaOptionChipRemovedText` prop as required prop for MultiSelect. It will be read when selection chip is removed.
- MultiSelect is now made visible again in the styleguidist and exported from the library

## Motivation and Context

This improves accessibility of the MultiSelect, especially for screen reader users.

## How Has This Been Tested?

Locally in styleguidist on macOS Safari and Brave browsers.

## Release notes

- Add `hintText` prop for MultiSelect
- Add `ariaOptionsAvailableText` prop for MultiSelect
- Add `ariaOptionChipRemovedText` prop for MultiSelect

### DIFF
--- a/src/core/Form/FilterInput/FilterInput.tsx
+++ b/src/core/Form/FilterInput/FilterInput.tsx
@@ -11,6 +11,7 @@ import {
 import { AutoId } from '../../../utils/AutoId';
 import { getConditionalAriaProp } from '../../../utils/aria';
 import { LabelText, LabelMode } from '../LabelText/LabelText';
+import { HintText } from '../HintText/HintText';
 import { StatusText } from '../StatusText/StatusText';
 import { baseStyles } from './FilterInput.baseStyles';
 
@@ -46,6 +47,8 @@ interface InternalFilterInputProps<T>
   labelMode?: LabelMode;
   /** Text to mark a field optional. Will be wrapped in parentheses and shown after labelText. */
   optionalText?: string;
+  /** Hint text to be shown below the label */
+  hintText?: string;
   /**
    * 'default' | 'error'
    * @default default
@@ -84,6 +87,7 @@ class BaseFilterInput<T> extends Component<FilterInputProps & InnerRef> {
       visualPlaceholder,
       labelText,
       labelMode,
+      hintText,
       optionalText,
       status,
       statusText,
@@ -120,6 +124,7 @@ class BaseFilterInput<T> extends Component<FilterInputProps & InnerRef> {
       }
     };
 
+    const hintTextId = hintText ? `${id}-hintText` : undefined;
     const statusTextId = statusText ? `${id}-statusText` : undefined;
 
     return (
@@ -140,6 +145,7 @@ class BaseFilterInput<T> extends Component<FilterInputProps & InnerRef> {
           >
             {labelText}
           </LabelText>
+          <HintText id={hintTextId}>{hintText}</HintText>
           <HtmlDiv className={filterInputClassNames.functionalityContainer}>
             <HtmlDiv className={filterInputClassNames.inputElementContainer}>
               <HtmlInput
--- a/src/core/Form/FilterInput/FilterInput.tsx
+++ b/src/core/Form/FilterInput/FilterInput.tsx
@@ -157,6 +157,7 @@ class BaseFilterInput<T> extends Component<FilterInputProps & InnerRef> {
                 placeholder={visualPlaceholder}
                 {...getConditionalAriaProp('aria-describedby', [
                   statusTextId,
+                  hintTextId,
                   ariaDescribedBy,
                 ])}
                 autoComplete="off"
--- a/src/core/Form/MultiSelect/MultiSelect/AriaAnnounceText.tsx
+++ b/src/core/Form/MultiSelect/MultiSelect/AriaAnnounceText.tsx
@@ -21,9 +21,13 @@ const debounce = (func: Function, waitFor: number) => {
 };
 
 export interface AriaAnnounceTextProps {
+  /** Unique id */
   id: string;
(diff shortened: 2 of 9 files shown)


src/core/Form/MultiSelect/MultiSelect/MultiSelect.tsx
```diff
@@ -680,6 +719,25 @@ class BaseMultiSelect<T> extends Component<
...
+          aria-live="polite"
+          aria-atomic="true"
+          id={`${id}-filteredItems-length`}
+        >
+          {this.focusInInput(this.getOwnerDocument()) ? (
+            <>
+              {filteredItems.length}
+              {ariaOptionsAvailableText}
+            </>
+          ) : null}
```
ketsappi, line 732: Safari doesn't seem to work nicely with aria region when I first tried like this (might had some error):

```tsx
{this.focusInInput()
  ? `${filteredItems.length} ${ariaFilteredAmountText}`
  : null}
```

src/core/Form/FilterInput/FilterInput.tsx
```diff
@@ -140,6 +145,7 @@ class BaseFilterInput<T> extends Component<FilterInputProps & InnerRef> {
           >
             {labelText}
           </LabelText>
+          <HintText id={hintTextId}>{hintText}</HintText>
```
LJKaski, line 148: This needs to be added to the getConditionalAriaProp call below.

ketsappi: This change should now be committed.

src/core/Form/MultiSelect/MultiSelect/MultiSelect.tsx
```diff
@@ -84,11 +84,11 @@ export interface MultiSelectProps<T extends MultiSelectData> {
   /** Text for screen reader to indicate how many items are selected */
   ariaSelectedAmountText: string;
   /** Text for screen reader indicating the amount of available options after filtering by typing. Will be read after the amount.
-   * E.g 'options available' as prop value would result in '<amount> options available' being read to screen reader upon removal.
+   * E.g 'options available' as prop value would result in '{amount} options available' being read to screen reader upon removal.
```
LJKaski, line 87: Another nitpicking comment, but the ending should probably be "-- being read **by** screen reader upon removal" instead of "to" since it's the screen reader that is doing the reading. Same goes for the other instances of this doc format.

ketsappi: This fix is now committed.

LJKaski (approved):

src/core/Form/MultiSelect/MultiSelect/MultiSelect.md
```diff
@@ -89,15 +89,18 @@ const defaultSelectedTools = [
...
   <MultiSelect
-    labelText="MultiSelect"
+    labelText="Tools"
+    hintText="You can filter results by typing in the field"
     items={tools}
     chipListVisible={true}
     ariaChipActionLabel="Remove"
     removeAllButtonLabel="Remove all selections"
-    visualPlaceholder="Choose your tool(s)"
+    visualPlaceholder="Choose your tools"
```
ketsappi, line 98: Removed the parens from this and another example as those sounded weird when read by screenreader.

src/core/Form/MultiSelect/MultiSelect/AriaAnnounceText.tsx
```diff
@@ -0,0 +1,90 @@
+import React from 'react';
+import { AriaLiveMode } from '../../types';
+import { VisuallyHidden } from '../../../VisuallyHidden/VisuallyHidden';
+
+const debounce = (func: Function, waitFor: number) => {
```
aappoalander, line 5: This whole thing seems to be overly complicated by a quickly looking. Is there some reasoning for having this kind of somewhat complex structure instead of:
- state with timeoutId
- getDerivedStateFromProps to reset timeout and re-apply it when necessary
- Function for the timeout to set announce text state
- Using announce text from state for render

src/core/Form/MultiSelect/MultiSelect/MultiSelect.tsx
```diff
@@ -303,25 +315,36 @@ class BaseMultiSelect<T> extends Component<
...
+  };
+
+  private focusInInput = (ownerDocument: Document | null) =>
+    ownerDocument
+      ? ownerDocument.activeElement === this.filterInputRef.current
+      : false;
 
+  private handleBlur = (event: React.FocusEvent<HTMLInputElement>) => {
+    event.preventDefault();
+    if (this.popoverListRef !== null && this.popoverListRef.current !== null) {
```
aappoalander, line 338: This seems to be unnecessary as there is a guard clause just below doing the exact same check and returning if either of the checked values is null.

aappoalander (approved):

aappoalander:
Tested with multiple browsers and VoiceOver on Mac OS. The improvements seem to be working as expected, but some old issues remain not fixed. Left a couple of comments for future improvements, but will merge the PR to keep things rolling during vacation.

src/core/Form/MultiSelect/MultiSelect/MultiSelect.tsx
ketsappi: Good find. Removing this in next fix branch for MultiSelect.

PR #487, opened 2021-06-30
[Fix] ToggleInput example

aappoalander:
## Description
Fix for ToggleInput example. This PR combines two separate onClick handlers for a single ToggleInput component example.

## Motivation and Context
Having more than one of the same event handlers in a React component is considered an error. This error had gone unnoticed due to md file formats not being checked by linters or build.

## Release notes
ToggleInput
- Fix example onClick handler.


ketsappi (approved):
Tested locally in styleguidist to now work nicely. No more onClick prop told twice.

PR #489, opened 2021-07-12
Remove normalize.cssinjs

jolle-reaktor:
## Description

This PR removes the `normalize.cssinjs` dependency and replaces it with an explicit object with all of the reset styles. This does not change how the utility reset functions work (just the way the data is stored); however, updates (if any) from the normalize.css stylesheet will no longer be applied automatically.

## Motivation and Context

- `normalize.cssinjs`'s dependencies haven't been updated updated in a while, and as such there are some vulnerable packages like lodash and postcss included (this causes `npm audit` to fail unless the packages are overridden manually via `resolutions`)
- `normalize.cssinjs`'s dependency `@j-kallunki/css-in-to-js` requires postcss to be included even in runtime bundles, which increases bundle sizes
  - this is unnecessary as the underlying normalize.css will not change between build-time and runtime, so it's just unnecessary work for the client bundle to parse CSS via postcss and download postcss itself when the outcome is basically the same always
- explicitly stating the used reset styles allows editing/versioning them down the line and also acts as a reference for what is being reset and how
- by removing `normalize.cssinjs` runtime bundle sizes can be decreased by as much as ~(53 + 8 = 61) kB (tree shaken, non-tree shaken ~(500 + 10 = 510) kB) because `normalize.cssinjs` is not tree-shakeable

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,7 +138,6 @@
     "@reach/popover": "0.15.0",
     "@types/react-modal": "3.12.0",
     "classnames": "2.3.1",
-    "normalize.cssinjs": "1.1.1",
     "polished": "4.1.2",
     "react-modal": "3.13.1",
     "react-popper": "2.2.5",
--- a/src/reset/utils/index.ts
+++ b/src/reset/utils/index.ts
@@ -1,7 +1,68 @@
 import { css } from 'styled-components';
-import { normalizeCssInJs } from 'normalize.cssinjs';
 
-const normalize = normalizeCssInJs({ cssToString: true });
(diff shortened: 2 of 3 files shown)


ketsappi:
Thanks for the contribution and your hard work!
As with the other [PR ](/vrk-kpa/suomifi-icons/pull/57) it might take a little longer than usual to review & merge PRs during summertime.

aappoalander:
This proposition was implemented as part of major styles refactoring #491, mostly as proposed with this PR. Thanks for contributing!

PR #490, opened 2021-07-13
[Fix] Move LanguageMenu to core

ketsappi:
## Description

- **LanguageMenu** is now moved to core level
- Use object style for CSS class names to prevent duplicated content that are more prone to error
- Directory structure added for **LanguageMenu** 
  - Some components and props were renamed as it was possible to drop unwanted abstraction with folder structure rethinking.
- Whole **components** level is removed from the library as nothing were left after the removal of this last component.

This PR is continuing the job done in #488.

## Motivation and Context

Removing the components levels abstraction will remove unwanted complexity. This will possibly increase the speed of making changes or new features.

## How Has This Been Tested?
Locally in styleguidist and in CRA-TS project.

## Release notes

**Breaking changes**
- `LanguageMenuItemLanguage,
  LanguageMenuItemLanguageProps,
  LanguageMenuLinkLanguage,
  LanguageMenuLinkLanguageProps` removed. Use `LanguageMenuItem, LanguageMenuItemProps, LanguageMenuLink, LanguageMenuLinkProps` instead.
- Statics `languageItem, LinkLanguage` removed from `LanguageMenu`. Use `LanguageMenuItem` or `LanguageMenuLink` instead.


src/core/LanguageMenu/LanguageMenuItem/LanguageMenuItem.tsx
```diff
@@ -1,22 +1,20 @@
...
 import classnames from 'classnames';
-import {
-  MenuLink as CompMenuLink,
-  LanguageMenuLinkProps,
-} from '../../components/LanguageMenu/LanguageMenu';
+import { MenuItem } from '@reach/menu-button';
+import { LanguageMenuItemBaseProps } from '../LanguageMenu';
 
-export interface LanguageMenuLinkLanguageProps extends LanguageMenuLinkProps {
+export interface LanguageMenuItemProps extends LanguageMenuItemBaseProps {
```
ketsappi, line 6: It seems that GitHub shows wrong files in the compare, because of the folder structure change and rename. Also same case with the `LanguageMenuLink` file below.

src/index.tsx
```diff
@@ -48,17 +48,11 @@ export {
...
-} from './core/LanguageMenu/LanguageMenu';
-export {
   LanguageMenuItem,
   LanguageMenuItemProps,
   LanguageMenuLink,
   LanguageMenuLinkProps,
-  LanguageMenuItemLanguage,
-  LanguageMenuItemLanguageProps,
-  LanguageMenuLinkLanguage,
-  LanguageMenuLinkLanguageProps,
```
ketsappi, line 60: These are no longer exported as those can be used directly from `LanguageMenuItem, LanguageMenuItemProps` and `LanguageMenuLink, LanguageMenuLinkProps`.

src/core/LanguageMenu/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,236 @@
...
+  as?: SupportedMenuLinkComponent;
+}
+
+export interface LanguageMenuLinkBaseProps
+  extends Omit<LanguageMenuLinkPropsWithType, 'type'> {}
+
+export type LanguageMenuPopoverItemsProps =
+  | LanguageMenuItemBaseProps
+  | LanguageMenuLinkPropsWithType;
+type OptionalLanguageMenuPopoverProps = {
```
aappoalander, line 59: This could probably be written as Partial<MenuPopoverProps>, but should not make any difference if left as is.

aappoalander:
Looks good! This component could be simplified further, because it supports unnecessary use cases like custom props and components as pros instead of using composition. These improvements should be implemented later on as a separate PR, but I decided to comment here anyways.

aappoalander (approved):

aappoalander:
Tested with all Mac OS setups (SSR test project, CRA-TS, Styleguidist and DS-site) using Brave, Safari, Firefox and Edge with VO on. Seems to be working as expected.

Issue #225, opened 2019-10-28
TokensProvider (was ThemeProvider)


J-Kallunki:
## The Problem

Custom tokens need to be given to every component separately.

## Proposed Solution

Component API has prop `tokens` as `PartialTokens`. Implement wrapper `SuomifiTokensProvider` with React context API - like Styled Components ThemeProvider.

## Alternative Solutions

`SuomifiThemeProvider` containing also internal theme customizations? Might be too complicated to get working as theme is based on tokens and what tokens are passed to define what theme etc.

## Additional Info

Implement themeProvider #79
https://www.styled-components.com/docs/advanced

aappoalander:
Resolved by #491

PR #491, opened 2021-08-02
Fix/styles and types cleanup

aappoalander:
## Description
This maintenance PR removes redundant abstractions and typings and simplifies styles handling. Theme handling is refactored to use a provider pattern instead of constant theme. In additions, some minor improvements to folder structure are introduced.

All snapshots remain unchanged.

## Motivation and Context
Current styles and types implementations date back over two years to situation where the goal was to have an unstyled abstraction of all components. This is no longer the goal and most of these abstractions have become redundant.

To simplify the codebase, the theme and tokens handling had to be refactored. Theme is now provided via context and can be customized when needed. Components only use the theme via the context and do not import anything else from the theme directly. Theme utils are still used directly, but style related common tools are now located under utils.

In addition, css-in-to-js had multiple vulnerabilities due to no updates and was removed as inspired by PR:#489

Removing of custom Reach-UI typings resulted in simplified implementation of LanguageMenu, which no longer supports custom components or menu popover class names. 

## How Has This Been Tested?
Tested on MacOS using Chrome, Edge and Safari with VoiceOver. Tested using styleguidist build, CRA-TS and SSR test applications via verdaccio. In addition, tested with styleguidist using Android and Chrome.

## Release notes
Theme and styles:
  - Add export for suomifiTheme and related interfaces
  - Add and export SuomifiThemeProvider and related documentation
  - Remove css.in-to-js dependency and replace it with static object
  - Refactor theme generation and remove redundant focus utils. Include derived tokens to theme.
Directory structure:
  - Remove unused login icon and directory
  - Move styleguidist components to docs directory
  - Cleanup utils directory structure
Types:
  - Remove all custom Reach-UI types and replace with the ones exported by Reach-UI
LanguageMenu:
  - Breaking changes: Remove languageMenuPopoverProps and languageMenuPopoverComponent  from LanguageMenuProps, remove related interfaces, remove popover and popoverLang classNames from internal styles.


src/core/theme/index.ts
```diff
@@ -1,73 +1,67 @@
...
  * @param tokens SuomifiTokens with libraryTokenOverrides, defaults to suomifi-design-tokens
  * @param tokens.colors color tokens, defaults to suomifi-design-tokens colors
  * @param tokens.spacing spacing tokens, defaults to suomifi-design-tokens spacing
  * @param tokens.typography typography tokens, defaults to suomifi-design-tokens typography
  * @param tokens.values token values in object format, defaults to suomifi-design-tokens values, not connected to other params
  */
-export const suomifiTheme = (
-  props: Partial<DefaultInternalTokens & PartialTokens>,
-) => {
+export const getSuomifiTheme = (
```
aappoalander, line 62: This is currently unused and probably will not work as expected. The tooling needs further improvements once a ThemeProvider will be introduced. The basic ides should be like this:
- Default tokens are static
- User can provide overriding tokens
- Derived tokens like focus are generated from default and provided tokens, but only when the tokens change on toplevel (e.g. Provider props change).
Most of the things above should already work.

src/core/theme/index.ts
```diff
@@ -1,73 +1,112 @@
...
-  typography?: Partial<TypographyDesignTokens>;
+/**
+ * SuomifiTheme
+ * General styles and design tokens for suomifi-ui-components
+ */
+export interface SuomifiTheme {
+  /** Color design tokens as [key:string]: CSS string format with color name as key. */
+  colors: ColorDesignTokens;
+  /** Gradient design tokens as [key:string]: CSS string format with key derived from gradient name colors */
+  gradients: GradientDesignTokens;
```
aappoalander, line 49: It is questionable if all of these should be called design tokens. Some are based on colors etc. but cannot / should not be directly derived by overridden colours due to various reasons. In a way gradients, shadows and transitions are similar to colors, spacing and typography => not derived, but smallest granular building blocks in the ds.

ketsappi:
_Remembered the old issue #225 that can be closed when this PR is ready._

src/core/theme/SuomifiTheme/SuomifiTheme.ts
```diff
@@ -0,0 +1,205 @@
...
+  focus: {
+    absoluteFocus: absoluteFocus(desingTokens),
+    boxShadowFocus: boxShadowFocus(desingTokens),
+    noMouseFocus,
+  },
+  shadows: shadows(desingTokens.colors),
+  gradients: gradients(desingTokens.colors),
+};
+
+/** Suomi.fi theme */
```
ketsappi, line 95: Maybe changing this
```
Suomi.fi theme
```
Also to be like it seems to be in other places

```
SuomifiTheme
```

src/core/Modal/ModalContent/ModalContent.tsx
```diff
@@ -17,6 +18,7 @@ interface InternalModalContentProps extends ModalContentProps {
   id?: string;
   modalVariant: ModalVariant;
   scrollable: boolean;
+  theme: SuomifiTheme;
```
ketsappi, line 21: Could this be done like in some other components, by extending `SuomifiThemeProp` instead of having it told in these props? Or is there certain requirement for it to work correctly to do it like it is now?

E.g
```tsx
class BaseModalContent extends Component<
  InternalModalContentProps & SuomifiThemeProp
> {
```

This same applies to other components that have now `theme` added to their props.

src/core/utils/AutoId/AutoId.tsx
```diff
@@ -5,7 +5,7 @@
  * Idea of the AutoId is to patch incremented id after the first render.
```
ketsappi, line 5: Would it be nice to have index in the `utils` level that when using utils, no need to tell the directories so much?

From current 
`import { AutoId } from '../../../utils/AutoId/AutoId';`
 to 
`import { AutoId } from '../../../utils';`

ketsappi (commented):
Left some comments here and there but none shouldn't be any real problem. Suggestions here and there to open discussion.

Will next test library with test benches.

src/core/utils/Debounce/Debounce.tsx
```diff
@@ -24,7 +24,7 @@ export class Debounce<T extends Object> extends Component<DebounceProps<T>> {
     const { waitFor } = this.props;
     this.timeout = setTimeout(() => {
       callback(value);
-    }, waitFor);
+    }, waitFor) as any;
```
LJKaski, line 27: Why is this change needed?

src/core/theme/index.ts
aappoalander: This has been implemented after the initial comment. Advanced merge is still missing, so it's not possible e.g. to provide custom colors for gradients and in addition one custom gradient.

src/core/Modal/ModalContent/ModalContent.tsx
aappoalander: No need to have it this way. It was implemented before SuomifiThemeProp was implemented. Will fix.

src/core/utils/AutoId/AutoId.tsx
aappoalander: Might make sense

src/core/utils/Debounce/Debounce.tsx
aappoalander: As discussed, something has changed in typings / typescript / linting. Not sure what exactly, since there were no custom typings for this before the changes of this PR, but nevertheless, typings are broken without the change. For some reason, IDE/TS resolves the ReturnType for setTimeout to be Node.js Timeout rather than a number (browser environment setTimeout). For the setTimeout function however, it is resolved as number => number is not compatible with Node.js Timeout. No idea why this is happening as both declarations are inside the same React class and should resolve to the same environment (browser / Node). There is really no way around it without getting the environment to resolve the ReturnType and setTimeout types the same way.

LJKaski: Thanks for clarifying. So be it then, at least for now 👍

src/index.tsx
```diff
@@ -82,4 +82,30 @@ export {
   VisuallyHidden,
   VisuallyHiddenProps,
 } from './core/VisuallyHidden/VisuallyHidden';
+export {
```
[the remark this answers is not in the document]
aappoalander: Seems to be badly ordered, there was some kind of order somewhere, but might a different file. Not sure what would be a good way to arrange these still. Maybe top down => actual theme object, theme type, types theme consists of, provider and types? Or alphabetically. Not sure.

src/core/utils/Debounce/Debounce.tsx
aappoalander: Yeah, not ideal, but could not find a workaround other than type casting of which the implemented one was the simplest / safest and had the least amount of side effects.

src/core/utils/AutoId/AutoId.tsx
aappoalander: Tried this, but it makes imports difficult to read due to relative paths with multiple utils folders. Decided not to implement this with this PR.

.styleguidist/themevalues.md
```diff
@@ -0,0 +1,157 @@
...
+<SuomifiThemeConsumer>
+  {({ suomifiTheme }) => (
+    <ThemeProps title="spacing:" values={suomifiTheme.spacing} />
+  )}
+</SuomifiThemeConsumer>;
+```
+
+### Typography
+
+Theme typography is of type _`TypographyDesignTokens`_ and defines most typograhpy styles used in the components.
```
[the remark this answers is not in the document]
aappoalander: should be now fixed

src/core/theme/SuomifiTheme/SuomifiTheme.ts
```diff
@@ -0,0 +1,205 @@
...
+  values: RawDesignTokens;
+  /** Z-idnex design tokens as [key:string]: CSS string format with z-index type name as key */
+  zindexes: ZIndexDesignTokens;
+}
+
+export interface SuomifiThemeProp {
+  theme: SuomifiTheme;
+}
+
+const desingTokens = {
```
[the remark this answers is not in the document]
aappoalander: should be now fixed

src/index.tsx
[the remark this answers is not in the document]
aappoalander: Used similar order with theme/index as agreed.

src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
```diff
@@ -71,6 +70,8 @@ export const baseStyles = css`
     border-color: ${theme.colors.depthDark3};
     border-style: solid;
     border-width: 0 1px 1px 1px;
+    /* stylelint-disable */
+    /* prettier-ignore */
```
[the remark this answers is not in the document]
aappoalander: These should be now removed

src/core/theme/SuomifiThemeProvider/SuomifiThemeProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+export interface SuomifiThemeProviderProps {
+  /**
+   * Partial Suomifi Theme with needed tokens and styles defined.
+   * Tokens and styles not provided will be merged from default theme.
+   * SuomifiTheme values property is not supported.
+   * Gradients, Shadows and Focus are derived from provided colors, except if explicitly defined.
+   */
+  theme: PartialSuomifiTheme;
+  /** Children, returns null if no childred are provided */
```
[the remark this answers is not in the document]
aappoalander: fixed

src/core/theme/SuomifiTheme/SuomifiTheme.ts
aappoalander: Changed and added short explanation referring to default styles

src/core/Modal/ModalContent/ModalContent.tsx
aappoalander: Fixed as proposed

src/core/theme/SuomifiThemeProvider/SuomifiThemeProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+  children?: JSX.Element;
```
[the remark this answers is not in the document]
aappoalander: Fixed as proposed

ketsappi (approved):
Seems to work nicely when smoke tested in CRA-TS and next.js projects.

Good work!

LJKaski (approved):
A lot of changes for sure, but seems to work as advertised. Good work :)

PR #495, opened 2021-08-17
[Feature] Remove breadcrumb variant prop

aappoalander:
## Description
Refactor breadcrumbs and remove variant prop. Separate style concerns and improve structure to better support custom styles. Remove background color.

## Motivation and Context
As part of removing components level abstractions, static variants were also removed. Some components still use variants even if the components have separate API's and behaviours. This was the case with breadcrumbs and breadcrumb items as well, which made using the component, custom styling and maintaining the component difficult.

## How Has This Been Tested?
Tested with CRA-TS and SSR test application using MacOS, Chrome and Safari with VoiceOver on.

## Release notes
Breadcrumb
- Breaking changes: Remove variant prop. Use BreadcrumbLink instead of variant="link". Internal classnames have changed which may affect custom styles.
- Remove background color


ketsappi (approved):
Seems to work nicely when tested in styleguidist and cra-TS project.

PR #496, opened 2021-08-17
[Fix] MultiSelect adjustments

ketsappi:
## Description

- Removes unneeded check as found in [PR](https://github.com/vrk-kpa/suomifi-ui-components/pull/483#discussion_r678857334)
- Removes the problematic AriaAnnounceText component as it made it worse for the Windows.
- Add check to not move focus to menu if it's already there.

## Motivation and Context

Small adjustments that were missing from the previous PR.

## How Has This Been Tested?
Locally in styleguidist.

## Release notes

- MultiSelect adjustments for aria-live when using filter, selection from popover and removing chip.


aappoalander:
Seems to be working a bit better when compared to current dev. No new findings, but these old issues are still present:
- MacOS + Chromium: if only one option is selected, active descendant is not read (probably VO issue)
- MacOS + Safari: unusable with VO, active descendant is not read mostly at all
- Android + Chromium: difficult to select options with TalkBack on

aappoalander (dismissed):

aappoalander (approved):

PR #502, opened 2021-09-08
[Feature] Sync develop with master

LJKaski:
### Description
Master branch received a couple of dependency updates and is ahead of develop due to those. This PR synchronizes the develop branch with the master branch. Recent commits to master were cherry-picked onto develop.

A couple of adjustments were needed after the rebase to get everything working as it should. Some snapshots were also changed due to new and changed structures in develop that required attention.

### Motivation and Context
The develop branch should be based on the latest master branch.

### How Has This Been Tested?
Tested manually in styleguidist in chrome (so far).


src/core/Form/MultiSelect/MultiSelectItem/MultiSelectItem.baseStyles.tsx
```diff
@@ -20,16 +20,11 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       background-color: ${theme.colors.depthSecondaryDark1};
       color: ${theme.colors.blackBase};
 
-      & > .fi-multiselect-item_checkbox_icon {
-        position: absolute;
```
aappoalander, line 24: Is this correct? Looks like selected state no longer applies position absolute the checkbox icon, or was this redundant?

src/core/Icon/Icon.baseStyles.tsx
```diff
@@ -1,13 +1,31 @@
...
+  return css`
+    display: inline-block;
+    vertical-align: baseline;
+
+    .fi-icon-base-fill {
+      fill: ${resolvedColor};
+    }
+
+    .fi-icon-base-stroke {
+      stroke: ${resolvedColor};
```
aappoalander, line 21: I started to wonder what happens if there is no e.g. stroke defined for a given icon but color is given in props? Could this introduce problems by applying a stroke to something that was supposed to be "invisible"? One alternative could be to only apply stroke color when explicitly provided in props.

aappoalander: My bad, this should work as expected since fi-icon-base-stroke is only present for paths that should be rendered. So all good here.

src/core/Icon/Icon.md
```diff
@@ -7,8 +8,31 @@ import { Icon } from 'suomifi-ui-components';
   <Icon
     icon="login"
     ariaLabel="Login here"
-    className="my-icon--test"
+    className="my-icon-class"
```
aappoalander, line 11: Does class name have any effect? Should we rather leave it out or mention it in the docs?

src/core/Icon/Icon.md
```diff
@@ -1,4 +1,5 @@
-- Uses currentColor by default if `fill` prop is not given.
+- Uses currentColor by default if no `fill` or `color` prop is given
```
aappoalander, line 1: could we change currentColor to italic so it would stand out more?

src/core/Form/MultiSelect/MultiSelectItem/MultiSelectItem.baseStyles.tsx
LJKaski: Seemed redundant to me since it's defined on line 57 in the "base styles" for the icon class.

src/core/Icon/Icon.baseStyles.tsx
LJKaski: Yeah, the class should only be there if the stroke has the same predefined color as the fill. I remember stopping to think about this as well for a bit when writing the logic 😄

src/core/Icon/Icon.md
LJKaski: I don't think so. Some of the examples seem to have classNames defined "just because". I'll remove it to make the example clearer.

src/core/Form/MultiSelect/MultiSelectItem/MultiSelectItem.baseStyles.tsx
aappoalander: So it seems, looks like these selected styles for the actual icon might be redundant as the icon is never there when an item is not checked, but it should not be a concern for this PR.

aappoalander:
Tested with Styleguidist, CRA TS project and SSR test project using MacOS Brave and Safari. Seems to work as expected! Good work!

aappoalander (approved):

PR #504, opened 2021-09-15
[Feature] Update dependencies to latest possible versions.

LJKaski:
## Description
This PR updates dependencies to their latest suitable versions.

Some dependencies couldn't be upgraded because of some breaking changes that couldn't be accounted for. Prettier upgrade also apparently brought some changes to the way it handles '=' signs and line changes even though the config file stayed the same.

## Motivation and Context
Dependencies should be kept up to date.

## How Has This Been Tested?
Tested by building and running styleguidist in chrome as well as by using in CRA project.

### Dependencies
* Update vulnerable dependencies


src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -214,7 +214,7 @@ class BaseDropdown extends Component<DropdownProps & InnerRef> {
             <ListboxPopover
               position={positionMatchWidth}
               {...passDropdownPopoverProps}
-              onKeyDownCapture={(event: KeyboardEvent) => {
+              onKeyDownCapture={(event: React.KeyboardEvent) => {
```
aappoalander, line 217: This could / should probably be explicitly imported from React, but there is probably no agreed way to use these types for the time being. We should consider aligning all of these.

src/core/LanguageMenu/LanguageMenu/__snapshots__/LanguageMenu.test.tsx.snap
```diff
@@ -131,7 +131,7 @@ exports[`calling render with the same component on the same container does not r
       aria-haspopup="true"
       class="fi-language-menu_button fi-language-menu-language_button menu-language-test"
       data-reach-menu-button=""
-      id="menu-button--menu--1"
+      id="menu-button--menu"
```
aappoalander, line 134: I wonder if AutoId is still working as the increment has been removed?

aappoalander: Seems to be working :)

aappoalander:
Tested using Styleguidist, CRA-TS project and SSR test project using MacOS and Chrome. Also tested external builds using Node v.12 instead of internally required v.14. Seems to be working as expected! Good work!

aappoalander (approved):

PR #515, opened 2021-10-06
[Fix] documentation related to statics and component level.

aappoalander:
## Description
This PR aligns documentation with previous static and components level removals.

## Motivation and Context
Some of the documentation was still referring to statics and components level and needed to be updated to reflect the latest changes.

## How Has This Been Tested?
Tested using Styleguidst on MacOS using Chrome.

## Release notes
General changes
- Fix documentation related to use of statics and components level.

### DIFF
--- a/.styleguidist/typography.md
+++ b/.styleguidist/typography.md
@@ -1,4 +1,6 @@
-Suomifi-styleguide typography
+Suomifi-styleguide typography.
+
+Click the examples to copy the typography token name to clipboard.
 
 ```js noeditor
 import { default as styled } from 'styled-components';
--- a/.styleguidist/typography.md
+++ b/.styleguidist/typography.md
@@ -1,6 +1,6 @@
 Suomifi-styleguide typography.
 
-Click the examples to copy the typography token name to clipboard.
+Click the examples to copy related JSX tags to clipboard.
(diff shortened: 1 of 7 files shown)


.styleguidist/typography.md
```diff
@@ -1,4 +1,6 @@
-Suomifi-styleguide typography
+Suomifi-styleguide typography.
+
+Click the examples to copy the typography token name to clipboard.
```
ketsappi, line 3: It seems to copy the tag and not token name:

`Heading 1 hero` --> `<Heading variant='h1hero' smallScreen='true'></Heading>`

README.md
```diff
@@ -73,19 +73,17 @@ Components have variant-property for different versions of the current component
```
ketsappi, line 73: ^^ Do we still have some static variants as stated above?

aappoalander: And I did. From two almost identical descriptions, but not from here. Should be no fixed.

.styleguidist/typography.md
aappoalander: Fixed.

ketsappi (approved):

LJKaski (approved):

PR #524, opened 2021-10-21
[Feature] Use rollup for build

aappoalander:
THIS PR SHOULD NOT BE MERGED BEFORE 6.1. RELEASE / release containing alerts at least.

## Description
This PR replaces TSDX with Rollup and updates all possible dependencies to latest. Styleguide related dependencies are not updated. In addition, Builds have been changed to preserve modules instead of 

## Motivation and Context
TSDX has not been updated for over a year now and has a lot of vulnerabilities and quirks that cause additional maintenance burden. In addition, treeshaking is not supported for cjs builds with TSDX which might be an issue for some projects.

## How Has This Been Tested?
Should be tested thoroughly - under works.

## Release notes
General changes
* Replace TSDX with Rollup
* Introduce module based library instead of bundle


ketsappi:
Tested to build locally, and it seems that `.d.ts` files does no longer exist in the builds at the moment.

package.json
```diff
@@ -2,9 +2,10 @@
   "name": "suomifi-ui-components",
   "version": "6.1.0-beta.0",
   "description": "Suomi.fi UI component library",
-  "main": "dist/index.js",
-  "module": "dist/suomifi-ui-components.esm.js",
-  "typings": "dist/suomifi-ui-components.esm.d.ts",
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "typings": "dist/types/index.d.ts",
```
ketsappi, line 7: Tested to build locally, and it seems that .d.ts file/files does no longer exist in the builds at the moment.

ketsappi: This should be now fixed with the f602fd566095a6babc499d6aaba0bd8d83f21073 commit.
Was missing a script to copy the types to the **dist** directory

ketsappi (approved):

valger11 (approved):

PR #521, opened 2021-10-12
[Feature] Alert component

LJKaski:
## Description
This PR introduces the alert component to the library. It is used mainly for maintenance and disturbance alerts in the Suomi.fi services.

🚨 For testing purposes you might want to change the `maxWidth` value from `styleguidist.styles.js`

## Motivation and Context
This has been in the style guide as a design for a while, but hasn't been implemented in the library yet.

## How Has This Been Tested?
So far tested by running styleguidist locally on Windows with Chrome and Firefox + NVDA.

## Screenshots (if appropriate):
image

### Alert
* Introduces `Alert` and `InlineAlert` components
* Add `whiteBaseToDepthLight1` gradient

PR #528, opened 2021-11-02
[Feature] Alert component

ketsappi:
## Description
This PR introduces the alert component to the library. It is used mainly for maintenance and disturbance alerts in the Suomi.fi services.

🚨 For testing purposes you might want to change the `maxWidth` value from `styleguidist.styles.js`

_Original PR was #521 and it's commit history were somewhat "broken" so to have better looking commit history with least effort, we created new branch with the same changes. And the changes of PR is created by @LJKaski  and @valger11_

## Motivation and Context
This has been in the style guide as a design for a while, but hasn't been implemented in the library yet.

## How Has This Been Tested?
So far tested by running styleguidist locally on Windows with Chrome and Firefox + NVDA.

image

### Alert
* Introduces `Alert` and `InlineAlert` components
* Add `whiteBaseToDepthLight1` gradient


src/core/Alert/InlineAlert/InlineAlert.tsx
```diff
@@ -0,0 +1,100 @@
...
+  alertClassNames,
+  baseClassName,
+} from '../BaseAlert/BaseAlert';
+import { AutoId } from '../../utils/AutoId/AutoId';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
+import { baseStyles } from './InlineAlert.baseStyles';
+
+export interface InlineAlertProps extends BaseAlertProps {
+  /** Label for the alert */
+  labelText?: string;
```
ketsappi, line 17: This might be unwanted repetition, as `labelText` prop seems already in the `BaseAlertProps`.

ketsappi: I looked this closer and it seems that **InlineAlert** is the only one using the `labelText` prop. So it should be removed from the **BaseAlertProps** instead.

src/core/Alert/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  it('contains base classname', () => {
+    const { container } = render(customClassAlert);
+    expect(container.firstChild).toHaveClass('fi-alert');
+  });
+
+  it('contains custom classname', () => {
+    const { container } = render(customClassAlert);
+    expect(container.firstChild).toHaveClass('custom-class');
+  });
+});
```
ketsappi, line 40: This whole `describe` block could be moved inside the `props` describe block. 
And to be changed to singular as the prop name is `className` instead of `classnames`.

src/core/Alert/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  );
+  it('should have status and smallScreen prop specific classNames', () => {
+    const { container } = render(ErrorAlert);
+    expect(container.firstChild).toHaveClass('fi-alert--error');
+    expect(container.firstChild).toHaveClass('fi-alert--small-screen');
+  });
+  it('should have given className as props', () => {
+    const { container } = render(ErrorAlert);
+    expect(container.firstChild).toHaveClass('alert-test-class');
+  });
```
ketsappi, line 61: This test is redundant as it basically tests the same `className` prop that does not depend which status is used.
So this can be removed.

src/core/Alert/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+      smallScreen
+    >
+      Testcontent
+    </Alert>
+  );
+  it('should have status and smallScreen prop specific classNames', () => {
+    const { container } = render(ErrorAlert);
+    expect(container.firstChild).toHaveClass('fi-alert--error');
+    expect(container.firstChild).toHaveClass('fi-alert--small-screen');
+  });
```
ketsappi, line 57: After the removal of the duplicate test of `className` prop, we are only having one test left In this **describe** block.

We could simplify like this and make it more readable when actually testing the status prop that is not yet tested and what is basically left after the removal of duplicate test.

```tsx
  describe('status', () => {
    const ErrorAlert = (
      <Alert status="error" closeText="Close">
        Testcontent
      </Alert>
    );

    test('error status should have correct classname applied', () => {
      const { container } = render(ErrorAlert);
      expect(container.firstChild).toHaveClass('fi-alert--error');
    });

    const WarningAlert = (
      <Alert status="warning" closeText="Close">
        General warning
      </Alert>
    );

    test('warning status should have correct classname applied', () => {
      const { container } = render(WarningAlert);
      expect(container.firstChild).toHaveClass('fi-alert--warning');
    });
  });
```

With the above we have more describing test and it will test `error` and `warning` statuses.

src/core/Alert/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  it('Button should have given props', () => {
+    const { getByRole } = render(AlertWithButtonProps);
+    expect(getByRole('button')).toHaveAttribute(
+      'aria-labelledby',
+      'test-element',
+    );
+    expect(getByRole('button')).toHaveAttribute('disabled');
+    expect(getByRole('button')).toHaveTextContent('CLOSE');
+  });
```
ketsappi, line 88: We should better describe that we are testing. We want to test that close button will have all the given custom props, so we could modify this to be something like this instead. And we already have test for the content, so no need to test it anymore.

```tsx
describe('props', () => {
  test('closeButton should have the given closeButtonProps', () => {
    const AlertWithButtonProps = (
      <Alert
        closeText="Close"
        closeButtonProps={{
          'aria-labelledby': 'test-element',
          disabled: true,
          className: 'testClass',
        }}
      >
        Testcontent
      </Alert>
    );
    const { getByRole } = render(AlertWithButtonProps);
    const closeButton = getByRole('button');

    expect(closeButton).toHaveAttribute('aria-labelledby', 'test-element');
    expect(closeButton).toHaveAttribute('disabled');
    expect(closeButton).toHaveClass('testClass');
  });
//... other tests
)};
```

src/core/Alert/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  const AlertWithDefaultAriaLiveMode = (
+    <Alert id="testId" closeText="Close" ariaLiveMode="off">
+      Testcontent
+    </Alert>
+  );
```
ketsappi, line 94: This naming is misleading as the default `ariaLiveMode` seems to be `assertive` and not `off` as the variable naming suggests.

After the naming fix also check better describing name for the actual test to reflect this situation.

Maybe we should have separate test to see that there is correct `aria-live` by default and separate test that it is changed when giving `ariaLiveMode` prop.

src/core/Alert/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  it('has the given content', () => {
+    const { container } = render(alertWithElementChild);
+    expect(container.firstChild).toHaveTextContent('Test alert');
+  });
+
+  it('should match snapshot', () => {
+    const { container } = render(alertWithElementChild);
+    expect(container).toMatchSnapshot();
+  });
+});
```
ketsappi, line 22: This could also be moved inside the `props` describe block as the `children` is one of the props we are having for the Alert.

src/core/Alert/Alert/Alert.test.tsx
ketsappi: And like other props, this describe block of the `status` prop should be inside the existing `props` block.

```tsx
describe('props', () => {
  describe('status', () => {
  // .. stuff here
  });
 // .. other tests for other props here like before
  });
```

src/core/Alert/BaseAlert/BaseAlert.baseStyles.ts
```diff
@@ -0,0 +1,62 @@
...
+      background-color: ${theme.colors.alertLight1};
+      & .fi-alert_icon-wrapper .fi-icon .fi-icon-base-fill {
+        fill: ${theme.colors.alertBase};
+      }
+    }
+
+    &--warning {
+      background-color: #fff6e0; /** needs to be warningLight1 but the token is yet to be added */
+      & .fi-alert_icon-wrapper .fi-icon .fi-icon-base-fill {
+        fill: #b54439; /* Later add theme.colors.example styling */
```
ketsappi, line 58: Maybe change little more describing comment because 
`/* Later add theme.colors.example styling */` is confusing as we might not want to add stuff named like that.

If existing token in our code bases does not match with the design (design have e.g updated the tokens in design), we should tell that instead.


**Check and fix  also other baseStyles of Alert/InlineAlert for this kind of comments.**

src/core/Alert/Alert/Alert.md
```diff
@@ -0,0 +1,46 @@
+```js
+import { Alert, InlineAlert } from 'suomifi-ui-components';
+
+<>
+  <Alert closeText="Close" smallScreen>
+    The service will be temporarily unavailable on 5.6.2021 at 21.00 –
+    23.59 due to maintenance.
+  </Alert>
+
+  <Alert closeText="Close" labelText="Maintenance notification">
```
ketsappi, line 10: **Alert** interface does not have `labelText` as it was only for the **InlineAlert**.
So we need to remove it from this example.

valger11 (approved):

ketsappi:
CircleCI has problems with old dependencies and there is a working branch (#524) where this issue is fixed already.
As that PR needs more testing, we could not merge it first before this one.

Builds ran locally and built packages worked as expected.

PR #530, opened 2021-11-08
[Feature] Notification component

valger11:
## Description
This PR introduces the Notification component to the library. It is used to show notifications for suomifi users.

## Motivation and Context
There was need to implement this component as it was designed. And there was need inform users with component.

## How Has This Been Tested?
On Mac with Chrome and Safari

Screenshot 2021-11-04 at 17 41 34

## Release notes
* Introduces `Notification` and `InlineNotification` components


src/core/Notification/BaseNotification/BaseNotification.tsx
```diff
@@ -0,0 +1,32 @@
...
+   * @default 'neutral'
+   */
+  status?: 'neutral' | 'success' | 'error';
+  /** Set aria-live mode for the alert text content and label.
+   * @default 'assertive'
+   */
+  ariaLiveMode?: 'polite' | 'assertive' | 'off';
+  /** Label for the  alert */
+  labelText?: string;
+  /** Main content of the alert */
```
ketsappi, line 30: JSDocs seems to mention **Alert** instead of **Notification**.

src/core/Notification/Notification/Notification.tsx
```diff
@@ -0,0 +1,139 @@
...
+      className: customCloseButtonClassName,
+      'aria-describedby': closeButtonPropsAriaDescribedBy,
+      ...closeButtonPassProps
+    } = closeButtonProps;
+    const variantIcon =
+      status === 'neutral'
+        ? 'info'
+        : status === 'success'
+        ? 'checkCircle'
+        : status;
```
ketsappi, line 62: Not sure what is the best way, but `status` prop seems to have 3 different possibilities: `'neutral' | 'success' | 'error'`.
So, it might be easier for reader if the last option of this condition says it straight instead of defaulting to status, it could be `'error'` as there are not other options left according the interface. 

Or is there some other reason for the current implementation?

src/core/Notification/BaseNotification/BaseNotification.tsx
ketsappi: `/** Label for the  alert */`
Two spaces between **the** and **alert**

src/core/Notification/InlineNotification/InlineNotification.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  it('has the given content', () => {
+    const { container } = render(notificationWithElementChild);
+    expect(container.firstChild).toHaveTextContent('Test notification');
+  });
+
+  it('should match snapshot', () => {
+    const { container } = render(notificationWithElementChild);
+    expect(container).toMatchSnapshot();
+  });
+});
```
ketsappi, line 21: describe block of the `children` prop to be inside the existing `props` block.

src/core/Notification/InlineNotification/InlineNotification.test.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const InlineInlineNotificationComponent = (
+    <InlineNotification labelText="Inline Notification label">
+      Testcontent
+    </InlineNotification>
+  );
+  it('Inline component should contain given labelText', () => {
+    const { getByText } = render(InlineInlineNotificationComponent);
+    const label = getByText('Inline Notification label');
+    expect(label).toHaveClass('fi-notification_label');
+  });
```
ketsappi, line 33: These seems to test `labelText` prop, so it would be nice to be wrapped in it's own describe block located inside this props block to be consistent.

src/core/Notification/InlineNotification/InlineNotification.test.tsx
ketsappi: Check also for the Alert.test.tsx for same kind.

src/core/Notification/InlineNotification/InlineNotification.test.tsx
ketsappi: Check also for the Alert.test.tsx for same kind.

src/core/Notification/Notification/Notification.md
```diff
@@ -0,0 +1,97 @@
...
+      The service will be temporarily unavailable on 5.6.2021 at 21.00
+      – 23.59 due to maintenance.
+    </p>
+  </Notification>
+  <Notification
+    closeText="Close"
+    labelText="Maintenance notification"
+    buttonContent={<Button variant="secondary">Button</Button>}
+  >
+    <p>
```
LJKaski, line 37: What are your thoughts behind using the paragraph inside the main content in some examples?

src/core/Notification/BaseNotification/BaseNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+  status?: 'neutral' | 'success' | 'error';
+  /** Set aria-live mode for the Notication text content and label.
+   * @default 'assertive'
+   */
+  ariaLiveMode?: 'polite' | 'assertive' | 'off';
+  /** Label for the Notication */
+  labelText: string;
+  /** Main content of the Notication */
+  children?: ReactNode;
+  buttonContent?: ReactNode;
```
LJKaski, line 33: If the type is ReactNode, the name buttonContent might not be the best choice here. I think we need something more generic. It's a tricky one though. ´footerContent´ might work, but not sure if you can call that section a footer...

src/core/Notification/BaseNotification/BaseNotification.tsx
ketsappi: Alert is now fixed, but there is small typo in these: `Notication`, missing the `fi`.

src/core/Notification/Notification/Notification.md
valger11: Added Paragraph to every example

src/core/Notification/Notification/Notification.tsx
ketsappi: It seems that this is now fixed for Notification, but **InlineNotification** still has this issue.
If InlineNotification is about to be merged to be "same" as **Notification**, then this obviously is useless change for now.

src/core/Notification/Notification/Notification.tsx
```diff
@@ -0,0 +1,154 @@
...
+          </HtmlDiv>
+
+          <HtmlDiv
+            className={notificationClassNames.textContentWrapper}
+            id={id}
+            aria-live={ariaLiveMode}
+          >
+            <HtmlDiv className={notificationClassNames.content}>
+              {labelText && (
+                <Heading variant="h2" className={notificationClassNames.label}>
```
LJKaski, line 89: Could or should the heading level be an optional prop so that the users could control the heading level they want to use? What do you think?

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
```diff
@@ -0,0 +1,73 @@
...
+      flex-direction: column;
+      padding: 9px 20px 0 20px;
+      margin-top: ${theme.spacing.s};
+      margin-bottom: 0;
+      & .fi-notification_content {
+        vertical-align: middle;
+        ${font(theme)('bodyTextSmall')}
+      }
+    }
+    box-shadow: 0px 4px 8px 0px rgba(41, 41, 41, 0.2);
```
LJKaski, line 43: The internally used shadows are implemented in shadows.ts and used from there. Check if one of the existing ones would suit the purpose or add a new one in there if needed.

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
```diff
@@ -0,0 +1,73 @@
+import { css } from 'styled-components';
+import { SuomifiTheme } from '../../theme';
+import { element, font } from '../../theme/reset';
+
+export const baseNotificationBaseStyles = (theme: SuomifiTheme) => css`
+  &.fi-notification {
+    ${element(theme)}
+    ${font(theme)('bodyTextSmall')}
+    p {
+      font-size: 16px;
```
LJKaski, line 10: Can we really specify in our components that "if the user adds content of type X, it should have these styles"? I don't think so. What are your thoughts? Discuss before doing anything.

.styleguidist/styleguidist.sections.js
```diff
@@ -93,6 +93,13 @@ module.exports = {
               name: 'Text',
               components: getComponents(['Text', 'Paragraph']),
             },
+            {
```
LJKaski, line 96: Is this a primitive or more of a module?

src/core/Notification/InlineNotification/InlineNotification.md
```diff
@@ -0,0 +1,37 @@
+```js
+import { InlineNotification, Paragraph } from 'suomifi-ui-components';
+
+<>
+  <InlineNotification
+    closeText="Close"
+    labelText="Lorem ipsum dolor sit"
+  >
+    <Paragraph>
```
LJKaski, line 9: If we conclude that we can't adjust the paragraph styles as they are handled now, these will have to change somehow as well. I think the best course of action might be to just leave the semantics and styling for the projects and mention preferred semantics in the documentation. Any thoughts?

src/core/Notification/InlineNotification/InlineNotification.tsx
```diff
@@ -0,0 +1,113 @@
...
+  notificationClassNames,
+  baseClassName,
+} from '../BaseNotification/BaseNotification';
+import { AutoId } from '../../utils/AutoId/AutoId';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
+import { baseStyles } from './InlineNotification.baseStyles';
+
+export interface InlineNotificationProps extends BaseNotificationProps {
+  /** Label for the Notification */
+  labelText: string;
```
LJKaski, line 19: In the designs the success notification (which is what I think the InlineNotification is for) has no labelText, so we can't have that as a required prop. The need for it in general should probably be discussed.

src/core/Notification/InlineNotification/InlineNotification.tsx
```diff
@@ -0,0 +1,113 @@
...
+import {
+  BaseNotificationProps,
+  notificationClassNames,
+  baseClassName,
+} from '../BaseNotification/BaseNotification';
+import { AutoId } from '../../utils/AutoId/AutoId';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
+import { baseStyles } from './InlineNotification.baseStyles';
+
+export interface InlineNotificationProps extends BaseNotificationProps {
```
LJKaski, line 17: The InlineNotification name is still misleading in my opinion, since it does not exist for inline use.

src/core/Notification/InlineNotification/InlineNotification.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+      const { container } = render(notificationWithElementChild);
+      expect(container.firstChild).toHaveTextContent('Test notification');
+    });
+
+    it('should match snapshot', () => {
+      const { container } = render(notificationWithElementChild);
+      expect(container).toMatchSnapshot();
+    });
+  });
+  describe('labelText className Test', () => {
```
ketsappi, line 23: This `describe` block should be named after the prop name like others, so it should be just `labelText` to be consistent with other props.

src/core/Notification/InlineNotification/InlineNotification.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const InlineInlineNotificationComponent = (
+      <InlineNotification labelText="Inline Notification label">
+        Testcontent
+      </InlineNotification>
+    );
+    it('Inline component should contain given labelText', () => {
```
ketsappi, line 29: We could improve readability and remove unnecessary repetition from test name, so it could be shortened to be e.g:

```
it('should contain given labelText', () => {
```

src/core/Notification/Notification/Notification.tsx
ketsappi: If it is optional, it should be limited to `h2-h6`? So that it cannot be `h1` at all.

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
ketsappi: At least check if the shadow used in Expander is suitable for this case also; it seems to be close.

What you think about it, would it be suitable?

src/core/Notification/Notification/Notification.md
ketsappi: I'm not sure why we should have `p` or `Paragraph` used in examples? 
Could someone shed some light here for reasoning as it seems not to be obvious when doing the PR?

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
ketsappi: Yes, this kind of styling is not wanted. If paragraph is not used by the library user, Notification has broken styles.

src/core/Notification/InlineNotification/InlineNotification.tsx
ketsappi: Was there discussion that there is only `Notification` component because there was only one prop or so that was different? We have to discuss this.

LJKaski: Not a proper one yet. The styles are somewhat different with the success notification not having buttons, but it might be doable still. Would be much cleaner. I think this solution came from using a copy+paste of Alert as a base for the component.

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
```diff
@@ -0,0 +1,73 @@
...
+      margin-bottom: 0;
+      & .fi-notification_content {
+        vertical-align: middle;
+        ${font(theme)('bodyTextSmall')}
+      }
+    }
+    box-shadow: 0px 4px 8px 0px rgba(41, 41, 41, 0.2);
+    border-radius: 4px;
+    /* Status variant styles */
+    &--neutral {
```
LJKaski, line 46: If we end up keeping the two separate components, these styles should probably be divided among them according to which variants apply to which component.

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
```diff
@@ -0,0 +1,73 @@
...
+      font-size: 16px;
+    }
+    & .fi-notification_label {
+      ${font(theme)('bodySemiBold')}
+      margin-top: -4px;
+      margin-bottom: ${theme.spacing.xxs};
+    }
+
+    padding-bottom: 17px;
+    width: 100%;
```
ketsappi, line 19: All the same level rules should be at the same position, now it seems that they are scattered all over the file.
_Looking  `box-shadow` and `border-radius` rules that are introduced later in the file_

ketsappi: And these rules at this level should be in the top section before any others, e.g

```tsx
export const baseNotificationBaseStyles = (theme: SuomifiTheme) => css`
  &.fi-notification {
    ${element(theme)}
    ${font(theme)('bodyTextSmall')}

//  --> this place should be more appropriate for this level of styles <--

    p {
      font-size: 16px;
    }
    & .fi-notification_label {
      ${font(theme)('bodySemiBold')}
      margin-top: -4px;
// ... and the rest of styles file
```

src/core/Notification/BaseNotification/BaseNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+  styleWrapper: `${baseClassName}_style-wrapper`,
+  content: `${baseClassName}_content`,
+  label: `${baseClassName}_label`,
+  textContentWrapper: `${baseClassName}_text-content-wrapper`,
+  icon: `${baseClassName}_icon`,
+  iconWrapper: `${baseClassName}_icon-wrapper`,
+  closeButton: `${baseClassName}_close-button`,
+  closeButtonWrapper: `${baseClassName}_close-button-wrapper`,
+  smallScreen: `${baseClassName}--small-screen`,
+  inline: `${baseClassName}--inline`,
```
ketsappi, line 16: Hmm.. I think that class names that are not mutual for both versions are no supposed to be in this same `notificationClassNames` object.

E.g if only some are for inline version, they should be introduced only there.

What others will think about this approach? _(If accepted, Alert also need some slight changes regarding this)_

src/core/Notification/BaseNotification/BaseNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+  actionElementsWrapper: `${baseClassName}_actions-elements-wrapper`,
```
LJKaski, line 17: I would drop the plural from the name since it can be one or more elements. So just `actionElementWrapper` to be grammatically correct.

src/core/Notification/BaseNotification/BaseNotification.tsx
valger11: Now fixed to be Notification

src/core/Notification/Notification/Notification.tsx
valger11: Changed :status to be :'error'

src/core/Notification/Notification/Notification.md
valger11: I think semantically its paragraph specially if you have multiple "paragraphs".

src/core/Notification/BaseNotification/BaseNotification.tsx
valger11: now "buttonContent" is renamed "actionElementsWrapper".

src/core/Notification/Notification/Notification.tsx
valger11: Now it has props: headingVariant?: 'h2' | 'h3' | 'h4' | 'h5';
there was no option in heading for h6...

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
valger11: I created now one, because size of shadow is different. Now it is:
notificationBoxShadow: `0px 4px 8px 0px ${alphaHex(0.14)(colors.blackBase)}`,

.styleguidist/styleguidist.sections.js
valger11: Moved to under modules.

src/core/Notification/InlineNotification/InlineNotification.tsx
valger11: Fixed labelText to be optional.

src/core/Notification/InlineNotification/InlineNotification.test.tsx
valger11: renamed to be: 
describe('labelText', () => {

src/core/Notification/BaseNotification/BaseNotification.tsx
```diff
@@ -0,0 +1,34 @@
...
+  status?: 'neutral' | 'success' | 'error';
+  /** Set aria-live mode for the Notication text content and label.
+   * @default 'assertive'
+   */
+  ariaLiveMode?: 'polite' | 'assertive' | 'off';
+  /** Label for the Notication */
+  labelText?: string;
+  /** Main content of the Notication */
+  children?: ReactNode;
+  actionElements?: ReactNode;
```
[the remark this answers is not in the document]
valger11: JSdoc comments added

src/core/Notification/InlineNotification/InlineNotification.test.tsx
ketsappi: This seems to be now fixed.

src/core/Notification/BaseNotification/BaseNotification.tsx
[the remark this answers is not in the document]
ketsappi: ` /** Developer can add for example buttons to botton of notification */`

Maybe we could change it to somewhat different, let's discuss this in group what would be the best way to describe this prop.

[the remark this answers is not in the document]
LJKaski: I agree that this documentation doesn't really follow the style found elsewhere. Maybe something along the lines of 'elements to be shown under the notification text'.

src/core/Notification/Notification/Notification.tsx
LJKaski: Rather use hLevels found in HtmlH.tsx.

src/core/Notification/BaseNotification/BaseNotification.tsx
valger11: Plurals removed.

src/core/Notification/BaseNotification/BaseNotification.tsx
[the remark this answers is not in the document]
valger11: This has been fixed.

src/core/Notification/Notification/Notification.tsx
valger11: Fixed. now using hLevels.

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
valger11: Fixed, removed p{}

src/core/Notification/InlineNotification/InlineNotification.md
valger11: I using paragraph only in 1 example.

src/core/Notification/InlineNotification/InlineNotification.tsx
valger11: Inline removed

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
valger11: Now we have only one notification and Toast.

src/core/Notification/BaseNotification/BaseNotification.baseStyles.ts
valger11: fixed

src/core/Notification/BaseNotification/BaseNotification.tsx
valger11: Fixed, now we have only one component.

src/core/Notification/Notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+          <HtmlDiv
+            className={notificationClassNames.textContentWrapper}
+            id={id}
+            aria-live={ariaLiveMode}
+          >
+            <HtmlDiv className={notificationClassNames.content}>
+              {labelText && (
+                <Heading
+                  variant={headingVariant || 'h2'}
```
ketsappi, line 121: We want to be consistent and give the default values where we deconstruct the props, like done for `ariaLiveMode` and `status` props currently in this file.

src/core/Notification/Notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+  HtmlDivWithRefProps,
+} from '../../reset';
+import { hLevels } from '../../reset/HtmlH/HtmlH';
+import { Icon } from '../Icon/Icon';
+import { Heading } from '../Heading/Heading';
+
+import { AutoId } from '../utils/AutoId/AutoId';
+import { getConditionalAriaProp } from '../../utils/aria';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../theme';
+import { baseStyles } from './Notification.baseStyles';
```
ketsappi, line 19: 1. We prefer not having empty lines between imports.

2. And we would like to have imports ordered as how much much we have to traverse back; first with the longest e.g `../../` then with lower `../`

So the end in this case would be like

```tsx
import { getConditionalAriaProp } from '../../utils/aria';
import { Icon } from '../Icon/Icon';
import { Heading } from '../Heading/Heading';
import { AutoId } from '../utils/AutoId/AutoId';
import { SuomifiThemeProp, SuomifiThemeConsumer } from '../theme';
import { baseStyles } from './Notification.baseStyles';
```

src/core/Notification/Notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+  closeButtonWrapper: `${baseClassName}_close-button-wrapper`,
+  smallScreen: `${baseClassName}--small-screen`,
+  actionElementWrapper: `${baseClassName}_action-element-wrapper`,
+};
+
+export interface NotificationProps extends HtmlDivWithRefProps {
+  /** Style variant. Affects color and icon used.
+   * @default 'neutral'
+   */
+  status?: 'neutral' | 'success' | 'error';
```
ketsappi, line 40: I think we discussed that the `success` variant is removed as it will be used in new component, **Toast**.

src/core/Notification/Notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+  /** Text to label the close button. Visible + `aria-label` in regular size and only used as `aria-label` in small screen variant */
+  closeText: string;
+  /** Click handler for the close button */
+  onCloseButtonClick?: () => void;
+  /** Custom props passed to the close button */
+  closeButtonProps?: Omit<HtmlButtonProps, 'onClick' | 'aria-label'>;
+  /** Use small screen styling */
+  smallScreen?: boolean;
+  /** LabelText is header of component */
+}
```
ketsappi, line 62: Are we missing `labelText` prop or are we having unwanted JSDoc line for removed prop?

src/core/Notification/Notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+
+    const {
+      className: customCloseButtonClassName,
+      'aria-describedby': closeButtonPropsAriaDescribedBy,
+      ...closeButtonPassProps
+    } = closeButtonProps;
+    const variantIcon = status === 'neutral' ? 'info' : 'error';
+    const buttonAriaLabelSmallScreen = smallScreen
+      ? getConditionalAriaProp('aria-label', [closeText])
+      : false;
```
ketsappi, line 93: This seems odd.. as later when using this **const** in spread we are having `false` value there. The value returned from `getConditionalAriaProp` is an array that can be spread, but I do not think we would want to spread boolean value, `false`.


_That later place_
```tsx
{...buttonAriaLabelSmallScreen}
```

src/core/Notification/Notification.tsx
```diff
@@ -0,0 +1,188 @@
...
+   */
+  ariaLiveMode?: 'polite' | 'assertive' | 'off';
+  /** Label for the Notification */
+  labelText?: string;
+  /** Main content of the Notification */
+  children?: ReactNode;
+  /** Elements to be rendered under the notification text. Can be e.g. buttons, links etc. */
+  actionElements?: ReactNode;
+  /** Header variant for Notification */
+  headingVariant?: hLevels;
```
ketsappi, line 52: As we are giving default value later on, we need to communicate in the API. Look example how it is done for `ariaLiveMode` prop.

src/core/theme/SuomifiTheme/shadows.ts
```diff
@@ -10,4 +10,5 @@ export const shadows = (colors: ColorDesignTokens) => ({
   actionElementBoxShadow: `0 1px 2px 0 ${alphaHex(0.1)(
     colors.brandBase,
   )} inset`,
+  notificationBoxShadow: `0px 4px 8px 0px ${alphaHex(0.14)(colors.blackBase)}`,
```
ketsappi, line 13: Could we try to figure out a little more generic name for the this shadow to be more consistent with other shadows?

src/core/Notification/Notification.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+  });
+  it('Button should have given props', () => {
+    const { getByRole } = render(NotificationWithButtonProps);
+    expect(getByRole('button')).toHaveAttribute(
+      'aria-labelledby',
+      'test-element',
+    );
+    expect(getByRole('button')).toHaveAttribute('disabled');
+    expect(getByRole('button')).toHaveTextContent('Close');
+  });
```
ketsappi, line 94: As we are testing the `closeButtonProps` prop here it would be wise to have these test inside it's own `describe` block like e.g `className` prop above.

src/core/Notification/Notification.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+      expect(container.firstChild).toHaveTextContent('Test notification');
+    });
+
+    it('should match snapshot', () => {
+      const { container } = render(notificationWithElementChild);
+      expect(container).toMatchSnapshot();
+    });
+  });
+
+  describe('classnames', () => {
```
ketsappi, line 48: It would be better if the text would match the actual prop name, in this case `className`

src/core/Notification/Notification.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+    >
+      Testcontent
+    </Notification>
+  );
+
+  it('should have status and smallScreen prop specific classNames', () => {
+    const { container } = render(ErrorNotification);
+    expect(container.firstChild).toHaveClass('fi-notification--error');
+    expect(container.firstChild).toHaveClass('fi-notification--small-screen');
+  });
```
ketsappi, line 23: This would be better to change a little.

We are testing the `status` prop here with the **error** value.

So we want own `describe` block, named **status**, inside the existing props block like other tests in this file
And then appropriate name for the test that tells what kind of status variant we are testing.

src/core/Notification/Notification.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+  it('should have given className as props', () => {
+    const { container } = render(ErrorNotification);
+    expect(container.firstChild).toHaveClass('notification-test-class');
+  });
```
ketsappi, line 27: We already have test for `className` prop, so this test is not needed.

src/core/Notification/Notification.test.tsx
ketsappi: I don't see the reason for the `smallScreen` prop test to be here. It should have it's own describe block to test the effects of `smallScreen` prop.

src/core/Notification/Notification.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+  it('should have specified aria-live mode', () => {
+    const { container } = render(NotificationWithDefaultAriaLiveMode);
+    expect(container.querySelector('#testId')).toHaveClass(
+      'fi-notification_text-content-wrapper',
+    );
+    expect(container.querySelector('#testId')).toHaveAttribute(
+      'aria-live',
+      'off',
+    );
+  });
```
ketsappi, line 116: Here we are testing the `ariaLiveMode` prop, so own describe block for it would be nice.

src/core/Notification/Notification.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+        onClick={mockClick}
+        labelText="Lorem ipsum dolor sit"
+      >
+        Test content
+      </Notification>,
+    );
+    const button = getByRole('button');
+    fireEvent.click(button);
+    expect(mockClick).toHaveBeenCalledTimes(1);
+  });
```
ketsappi, line 131: `onClick` prop to own block would be nice.

src/core/Notification/Notification.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+});
+
+describe('accessibility', () => {
+  const TestNotification = (
+    <Notification closeText="Close" labelText="Lorem ipsum dolor sit">
```
ketsappi, line 136: As `labelText` prop is optional, then we could get rid of it from this test to be more consistent.

src/core/Notification/Notification.md
```diff
@@ -0,0 +1,103 @@
...
+    </Paragraph>
+    <Paragraph>
+      Et turpis eget tempor. Nam sed lacus ut lorem feugiat tincidunt.
+      Mauris quis nisi placerat, convallis turpis nec, auctor massa.
+      Nam ornare tortor id tortor tristique, in semper augue
+      fermentum. Nulla at pharetra leo, a ullamcorper metus. Donec
+      molestie velit tristique feugiat interdum. Donec ultrices
+      efficitur pellentesque. Sed luctus ac metus sed rhoncus.
+    </Paragraph>
+  </Notification>
```
ketsappi, line 61: Maybe we could drop this other long version as it takes so much screen space and gives no additional value.

Current is about 90 words or so, so maybe 30-50 is already enough to show long stuff?

LJKaski: I agree. I think only one with longer text is enought. All in all I'd try to cut the amount of examples here. We don't have to have every single combination of props portrayed here, but it's good to showcase the available variants and contents. Also, I don't think we can use the maintenance alert text here, since the notification is not meant for those kinds of messages.

src/core/Notification/Notification.md
```diff
@@ -0,0 +1,103 @@
...
+        <Button>Button</Button>
+        <Button variant="secondary">Button</Button>
+      </>
+    }
+  >
+    The service will be temporarily unavailable on 5.6.2021 at 21.00 –
+    23.59 due to maintenance.
+  </Notification>
+  <Notification closeText="Close">
+    <Paragraph>
```
LJKaski, line 46: I'd drop the other paragraph from here to save space. One is enough to showcase using elements as the main content. We also want to showcase using the correct styling even when using elements as the main content, so I'd add `style={{fontSize: '16px'}}` for the paragraph.

src/core/Notification/Notification.tsx
valger11: Fixed. Moved to deconstruct.

src/core/Notification/Notification.tsx
valger11: Fixed. Removed empty spaces.

src/core/Notification/Notification.tsx
valger11: Fixed. Removed success.

src/core/Notification/Notification.tsx
valger11: Fixed. Removed comment.

src/core/Notification/Notification.tsx
valger11: Fixed. If statement is now in jsx.

src/core/Notification/Notification.tsx
valger11: Fixed. added @default and exluded h1 from hLevels.

src/core/Notification/Notification.test.tsx
valger11: Fixed. Is now in own describe.

src/core/Notification/Notification.test.tsx
valger11: Fixed. is now classname

src/core/Notification/Notification.test.tsx
valger11: Fixed. Status and smallScreen have now own describe section

src/core/Notification/Notification.test.tsx
valger11: Fixed. Is now removed.

src/core/Notification/Notification.test.tsx
valger11: Fixed. Is now in own describe.

src/core/Notification/Notification.test.tsx
valger11: Fixed.  Is now in own describe.

src/core/Notification/Notification.test.tsx
valger11: Fixed. Removed labelText.

src/core/Notification/Notification.md
valger11: Fixed. Removed some of texts. Also removed Maintenance words.

src/core/Notification/Notification.md
valger11: Fixed. Added fontSize.

src/core/Notification/Notification.tsx
LJKaski: When did we decide to drop labelText from the api? Also, it's still part of the component below as well as the examples. Are you sure the prop is not needed?

src/core/Notification/Notification.tsx
ketsappi: Nitpicking here, but it seems that `import { getConditionalAriaProp } from '../../utils/aria';` is not as suggested in step 2; still in old order.

src/core/Notification/Notification.tsx
ketsappi: It seems that `labelText` is still in the API, but is now moved higher in the interface

https://github.com/vrk-kpa/suomifi-ui-components/blob/3e00e4c720163f7adf36cdd2cfebe40b1847f8fc/src/core/Notification/Notification.tsx#L44-L49

src/core/Notification/Notification.test.tsx
ketsappi: They now seem to be in own blocks, but those new describe blocks are not inside the existing `props` block like other props (e.g children and className).

Other way to say, they are not nested like others. They should be nested inside the **props**.

```tsx
 describe('props', () => {
  // other existing ones
  describe('status', () => {
  // ...
  });
  describe('smallScreen', () => {
  // ...
  });
  // other existing ones
});
```

src/core/Notification/Notification.test.tsx
ketsappi: Just nitpicking, but the describe block name does not match with the prop:
Currently it is `aria-live mode` when the prop name is actually  `ariaLiveMode`.

src/core/Notification/Notification.test.tsx
ketsappi: Just nitpicking, but the describe block name does not match with the prop:
Currently it is `aria-live mode` when the prop name is actually  `ariaLiveMode`.

src/core/Notification/Notification.test.tsx
ketsappi: Just a small typo in the describe block name: `aonClick`

src/core/Notification/Notification.md
ketsappi: Now the lengths are better.

As @LJKaski raised it might be good idea to look together which examples were are about to show. Maybe there are too many examples to show at this time.

src/core/Notification/Notification.tsx
ketsappi: Good job for excluding the h1!

src/core/Notification/Notification.test.tsx
valger11: Fixed. Those are now inside of props

src/core/Notification/Notification.test.tsx
valger11: Fixed. now it is describe('ariaLiveMode', () => {

src/core/Notification/Notification.test.tsx
valger11: fixed typo.

src/core/Notification/Notification.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * @default 'polite'
+   */
+  ariaLiveMode?: 'polite' | 'assertive' | 'off';
+  /** Label for the Notification */
+  labelText?: string;
+  /** Main content of the Notification */
+  children?: ReactNode;
+  /** Elements to be rendered under the notification text. Can be e.g. buttons, links etc. */
+  actionElements?: ReactNode;
+  /** Heading variant for Notification
```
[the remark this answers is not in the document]
valger11: Fixed comment

src/core/Notification/Notification.md
valger11: Fixed. now there are 3 examples

src/core/Notification/Notification.tsx
```diff
@@ -0,0 +1,185 @@
...
+   * @default 'polite'
+   */
+  ariaLiveMode?: 'polite' | 'assertive' | 'off';
+  /** Label for the Notification */
+  labelText?: string;
+  /** Main content of the Notification */
+  children?: ReactNode;
+  /** Elements to be rendered under the notification text. Can be e.g. buttons, links etc. */
+  actionElements?: ReactNode;
+  /** Heading variant for Notification. Prop for heading text is labelText.
```
LJKaski, line 48: I'd refrain from describing other props in different prop's descriptions. If the labelText is always going to be a heading, should that prop be headingText instead of labelText? Then this one could be described as `Heading variant for Notification heading`.

What do you think?

LJKaski: The other side is that in most other components the main label is behind the `labelText` prop, so in a sense it would be simpler to keep the same prop name.

src/core/Notification/Notification.tsx
valger11: Fixed. Changed order of imports.

src/core/Notification/Notification.tsx
valger11: Fixed it. Changed it to be headingText

.styleguidist/styleguidist.sections.js
```diff
@@ -167,6 +168,12 @@ module.exports = {
                 'LanguageMenuLink/LanguageMenuLink',
               ]),
             },
+            {
+              name: 'Notification',
+              components: getComponentWithVariants('Notification')([
```
LJKaski, line 173: the `getComponentWithVariants` method seems to generate an unnecessary subheading on the page now that the component has no variants presented on the same page.

valger11: Fixed. Now function call is getComponents.

LJKaski (approved):

PR #533, opened 2021-12-02
[Fix] Change node-sass to Dart Sass

ketsappi:
## Description

Change `node-sass` dev dependency to `sass`, Dart Sass.
Quoting the `node-sass` README:

> Warning: LibSass and Node Sass are deprecated. While they will continue to receive maintenance releases indefinitely, there are no plans to add additional features or compatibility with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.

node-sass was used by postcss and sass is direct placement to do the same for us.

## Related Issue

Fixes the issue we are having with CircleCI builds failing because of node-sass.

## Motivation and Context

- One of the service using comps library had to downgrade docker image used.
- No C++ requirement with Dart Sass, node-sass required it.

## How Has This Been Tested?

Locally in styleguidist and in separate CRA-TS test project. Styling seems to be in order when quickly glancing.

### Dependencies
- Change `node-sass` to `sass` (Dart Sass)


valger11 (approved):
Seems to work

LJKaski (approved):
Tested in CRA project, running in styleguidist and by glancing over the built styles. Seems to work as previously.

PR #536, opened 2021-12-09
[Fix] SingleSelect accessibility improvements

ketsappi:
## Description

This PR improves accessibility of the component when using with screen reader.

There was issue with the focus being in the item located in popover if there have been selection and therefore screen reader user was not getting correct information about the actual field (label, hintText etc.)
Also fixes the challenges to user what was caused by reading the options available amount too much.
Fix it to work in same logic when using with mouse or keyboard.

## Motivation and Context

These issues were found during accessibility clinic and we want to improve accessibility of the component.

## How Has This Been Tested?
Locally in styleguidist and in CRA-TS test project.

## Release notes

- SingleSelect accessibility improvements.


valger11 (dismissed):

LJKaski (commented):
Found an issue outside of the original scope of this PR, where the amount of list items wouldn't be read on Android + TalkBack. Suggested a possible solution on Slack, but needs checking.

src/core/Form/Select/BaseSelect/SelectItemList/SelectItemList.tsx
```diff
@@ -104,12 +104,12 @@ class BaseSelectItemList extends Component<
         role="listbox"
         onBlur={onBlur}
       >
-        <HtmlDivWithRef
-          forwardedRef={this.wrapperRef}
+        <HtmlSpanWithRef
```
valger11, line 107: This should be outside of ul, if is possible to make it that way!

ketsappi: Did you already tried if it does still work if doing that way?

LJKaski: image

Unfortunately does not seem to work with the span outside the list.

ketsappi: As earlier discussed, the actual fix will be done in separate PR.

LJKaski (approved):
Seems to work well in chrome + NVDA & Android + TalkBack. Keyboard use is good as well. As discussed, the element order issue in the list will be fixed in another PR. Good work 👍

LJKaski (approved):

PR #538, opened 2021-12-17
[Fix] Update internal aria util

ketsappi:
## Description

Update internal aria util. getConditionalAriaProp documentation and parameter naming changed.

## Motivation and Context

Util documentation says that it should receive one or multiple id's. But there are multiple use cases where we are giving just text value that needs to be read by screenreader and not the id of that element. See difference of using it with e.g `aria-describedby` vs `aria-label`.

Understanding how to use the util was confusing/harder than it should be.

## Release notes

No release notes as this is internal change.

### DIFF
--- a/src/utils/aria/aria.ts
+++ b/src/utils/aria/aria.ts
@@ -17,7 +17,8 @@ type ariaProp = { [key in ariaPropName]: string } | {};
 
 /**
  * Returns object with 'aria-' property which can be spread to props.
- * Using with different 'aria-' properties require different kind of use, see examples below.
+ * Using the util prevents generation of empty 'aria-' properties when no valid values are given.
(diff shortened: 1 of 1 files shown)


src/utils/aria/aria.ts
```diff
@@ -17,6 +17,7 @@ type ariaProp = { [key in ariaPropName]: string } | {};
 
 /**
  * Returns object with 'aria-' property which can be spread to props.
```
LJKaski, line 19: I'd maybe clarify the fact that using this util is a good way to prevent empty aria values when the conditions are not met. So I'd add something along the lines of: "prevents generation of empty 'aria-' properties when no valid values are given."

valger11 (approved):

PR #539, opened 2021-12-29
[Feature] Export inner form components

LJKaski:
## Description
This PR exports our inner `LabelText`, `HintText` and `StatusText` components for library users to use. LabelText `asProp` default value was also changed to `label` since it was used 100% of the time anyway internally and makes the component more intuitive to use.

**Still needs styleguidist documentation once we decide where the right place is to add it. Also needs tests for each of the components. Do not merge before they are added.** They could fit under several categories (Modules, Primitives or even under `Text`.)

## Motivation and Context

Library users also have custom components that might benefit from our accessible and styled Text components and exporting them has been in the discussions with service developers.

## How Has This Been Tested?

Tested locally in styleguidist as well as by using the newly exported components in a CRA project.

## Release notes
* Introduce LabelText component
* Introduce HintText component
* Introduce StatusText component


ketsappi:
Wrote this to remember checking later.

Is it required for the `StatusText` to have `display: inline` or could it be `block`?
And should the top margin to be `0`?

Here is quick test done that shows those values used so that it is not in the same line with input and do not have extra space between.

```tsx
import { Label, HintText, StatusText } from 'suomifi-ui-components';
import React from 'react';
<>
  <Label htmlFor="asd">This is a label for an adjacent input</Label>
  <HintText id="hint">This is a hint regarding an adjacent input</HintText>
  <input type="text" id="asd" aria-describedby="hint"/>
  <StatusText style={{display: 'block', marginTop: 0}} id="status" status="error">An error message</StatusText>
</>;
```

src/core/Form/HintText/HintText.md
```diff
@@ -0,0 +1,8 @@
+HintText is a component that generates an accessible hint text for an adjacent element. Give the ID of the HintText as an `aria-describedby` value to the related component. Applicable form components in the library already have HintText as an integral part.
+
+```js
+import { HintText } from 'suomifi-ui-components';
+import React from 'react';
+
+<HintText>This is a hint regarding an adjacent input</HintText>;
+```
```
valger11, line 8: Is span surely right html-tag for it?

aappoalander: Span is the most versatile option here as we cannot be sure of the needed semantics when this is used. Good question though and we could probably use something else as well at least internally. Might be different tag for different components though. But I propose we leave it as it is for now.

aappoalander:
> Wrote this to remember checking later.
> 
> Is it required for the `StatusText` to have `display: inline` or could it be `block`? And should the top margin to be `0`?
> 
> Here is quick test done that shows those values used so that it is not in the same line with input and do not have extra space between.
> 
> ```tsx
> import { Label, HintText, StatusText } from 'suomifi-ui-components';
> import React from 'react';
> <>
>   <Label htmlFor="asd">This is a label for an adjacent input</Label>
>   <HintText id="hint">This is a hint regarding an adjacent input</HintText>
>   <input type="text" id="asd" aria-describedby="hint"/>
>   <StatusText style={{display: 'block', marginTop: 0}} id="status" status="error">An error message</StatusText>
> </>;
> ```

These will be resolved in another PR.

ketsappi (approved):

valger11 (approved):

PR #540, opened 2021-12-30
[Feature] Toast

valger11:
## Description
This PR introduces the Toast component to the library. It is used to show Toasts for Suomifi users. Toast can inform user for example, that form is saved. 

## Motivation and Context
There was need to implement this component as it was designed. And there was need inform users with component.

## How Has This Been Tested?
On Mac with Chrome and Safari

Screenshot 2021-12-30 at 12 02 26

## Release notes
* Introduces Toast component


src/core/Toast/Toast.baseStyles.tsx
```diff
@@ -0,0 +1,42 @@
+import { css } from 'styled-components';
+import { SuomifiTheme } from '../theme';
+import { font, element } from '../theme/reset';
+
+export const baseStyles = (theme: SuomifiTheme) => css`
+  &.fi-toast {
```
aappoalander, line 6: Is there a reason we have adopted this kind of style def for the top level? Not sure if it's bad, but it kind of serves no purpose as these classnames will change during the build => these styles cannot be overridden using `.fi-toast` selector if I'm not mistaken. So the top level styles, preferably only default positioning and size, should be easy to override. Probably a style wrapper is needed as well.

src/core/Toast/Toast.baseStyles.tsx
```diff
@@ -0,0 +1,42 @@
...
+      & .fi-toast-content-wrapper {
+        vertical-align: middle;
+        ${font(theme)('bodyTextSmall')}
+      }
+      & .fi-toast-heading {
+        ${font(theme)('bodySemiBold')}
+        margin-top: -1px;
+        margin-bottom: ${theme.spacing.xxs};
+      }
+      .fi-toast_icon-wrapper {
```
aappoalander, line 32: Use of parent selector seems to be inconsistent. While not mandatory for these, it should be used in consistent manner.

src/core/Toast/Toast.test.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { Toast } from './Toast';
+import { axeTest } from '../../utils/test';
+
```
aappoalander, line 5: Additional tests are needed as all props in public API should be tested somehow.

src/core/Toast/Toast.tsx
```diff
@@ -0,0 +1,100 @@
...
+   * @default 'polite'
+   */
+  ariaLiveMode?: 'polite' | 'assertive' | 'off';
+  /** Label heading for the Toast */
+  headingText?: string;
+  /** Main content of the Toast */
+  children?: ReactNode;
+  /** Custom classname to extend or customize */
+  className?: string;
+  /** Elements to be rendered under the Toast text. Can be e.g. buttons, links etc. */
```
aappoalander, line 22: This description seems to be off. Children is not explicitly documented, but I assume it should be to give more information on allowed child types. Also, composition could have probably worked here as well, leaving all Heading props out of the main component API. Especially because headingText is not mandatory.

src/core/Toast/Toast.tsx
```diff
@@ -0,0 +1,100 @@
...
+        {...passProps}
+      >
+        <HtmlDiv className={toastClassNames.styleWrapper}>
+          <HtmlDiv className={toastClassNames.iconWrapper}>
+            <Icon icon="checkCircle" className={toastClassNames.icon} />
+          </HtmlDiv>
+
+          <HtmlDiv
+            className={toastClassNames.contentWrapper}
+            id={id}
```
aappoalander, line 63: The id should be either handled using AutoId or conditionally added. At the moment it can be empty as the prop is not mandatory.

src/core/Toast/Toast.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+describe('props', () => {
+  describe('children', () => {
+    const ToastWithElementChild = (
+      <Toast headingText="Lorem ipsum dolor sit">
```
ketsappi, line 9: It seems that `headingText` is an optional prop so there is no reason to include it in this test.

ketsappi: Same applies to `className` and `ariaLiveMode` tests where `headingText` props are given.

src/core/Toast/Toast.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+      const { container } = render(customClassToast);
+      expect(container.firstChild).toHaveClass('fi-toast');
+    });
+
+    it('contains custom classname', () => {
+      const { container } = render(customClassToast);
+      expect(container.firstChild).toHaveClass('custom-class');
+    });
+  });
+  describe('ariaLiveMode', () => {
```
ketsappi, line 42: Maybe also testing the **default** `ariaLiveMode`

src/core/Toast/Toast.tsx
ketsappi: And then remember to update the JSDoc of the `id` prop when it is using the **AutoId**.

src/core/Toast/Toast.baseStyles.tsx
ketsappi: At least there is no mutual conversation about to change to use this kind of approach. Same issue seems to be with the Notification, so should be fixed there also.

src/core/Toast/Toast.tsx
valger11: Do we need id? There is no button where we should move focus...

src/core/Toast/Toast.tsx
```diff
@@ -0,0 +1,100 @@
...
+  icon: `${baseClassName}_icon`,
+  iconWrapper: `${baseClassName}_icon-wrapper`,
+};
+class BaseToast extends Component<ToastProps & InnerRef> {
+  render() {
+    const {
+      ariaLiveMode = 'polite',
+      children,
+      className,
+      forwardedRef,
```
ketsappi, line 44: Why is `forwardedRef` deconstructed and never used?
Shouldn't it be allowed to be inside the `passProps` so that it will continue it's journey to **HtmlDivWithRef**?

src/core/Toast/Toast.md
```diff
@@ -0,0 +1,14 @@
+```js
+import { Toast } from 'suomifi-ui-components';
+
+<>
+  <Toast smallScreen>
```
ketsappi, line 5: Hmm.. there doesn't seem to be `smallScreen` prop for Toast. So this should be removed if there is no need for the actual different version for small screen, that is.

src/core/Toast/Toast.tsx
aappoalander: AutoId is not probably necessary here, but the whole prop should be conditional as there should be no empty id's. So if there is no id provided in props, there should be no id in DOM.

src/core/Toast/Toast.baseStyles.tsx
valger11: Fixed changed to:
  ${element(theme)}
  ${font(theme)('bodyTextSmall')}
  border-top: 4px solid ${theme.colors.successBase};
  width: 100%;
  box-shadow: ${theme.shadows.wideBoxShadow};
  border-radius: 4px;

  & .fi-toast-wrapper {

src/core/Toast/Toast.test.tsx
valger11: fixed All props tested

src/core/Toast/Toast.tsx
valger11: Fix. now JS Doc is following for children:
/** Text content of the Toast. */

src/core/Toast/Toast.tsx
valger11: There are no ids in dom if not given.

src/core/Toast/Toast.test.tsx
valger11: Fixed. removed those.

src/core/Toast/Toast.test.tsx
valger11: Added that test.

src/core/Toast/Toast.tsx
valger11: Should be fixed.

src/core/Toast/Toast.md
valger11: Removed smallScreen

src/core/Toast/Toast.tsx
ketsappi: Seems to be still same.

src/core/Toast/Toast.baseStyles.tsx
aappoalander: Looks good now in my opinion.

src/core/Toast/Toast.baseStyles.tsx
aappoalander: Seems to be fixed now.

src/core/Toast/Toast.test.tsx
aappoalander: Tests look good now in my opinion. One thing we could still add is testing forward refs but that might be an overkill for this component. Will resolve the conversation.

src/core/Toast/Toast.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    );
+    const heading = getByRole('heading', { level: 3 });
+    expect(heading).toHaveClass('fi-heading--h3');
+  });
+  describe('id', () => {
+    const ToastWithElementId = <Toast id="TestId">Test Toast</Toast>;
+    const { container } = render(ToastWithElementId);
+    expect(container.querySelector('#TestId')).toHaveClass(
+      'fi-toast-content-wrapper',
+    );
```
ketsappi, line 28: I suggest to look how the `id` prop is checked in the RadioButton. Currently this tests expects certain class instead of the id.

https://github.com/vrk-kpa/suomifi-ui-components/blob/340e7a3c0703308020e3dc38c59f4c5b5fb5e23e/src/core/Form/RadioButton/RadioButton.test.tsx#L118-L124

src/core/Toast/Toast.tsx
aappoalander: Seems to work, but is this defined somewhere with React or JSX that this is actually how it should work rather than adding an empty attribute to DOM?

src/core/Toast/Toast.test.tsx
valger11: I could get better way to get this id test working.

src/core/Toast/Toast.tsx
valger11: fixed

src/core/Toast/Toast.tsx
aappoalander: Looking this more thoroughly revealed a potential bug. The id is now referenced in the close buttons aria-described by so that when a screen reader user navigates to the button, also the contents of the text are read to help understand what the close button is about. If no id is provided for the component, only close button text is read. This component need AutoId in my opinion.

valger11:
Continues on branch https://github.com/vrk-kpa/suomifi-ui-components/pull/554

PR #541, opened 2021-12-30
[Fix] Notification layout

valger11:
## Description
Developer team found out that look and feel was't like it should be with short content. Close button was not aligned to right corner, with short content. PR makes look and feel as in designs. Also 2px removed from top margin of button.  Also Notification will be exported in project and will be visible in styleguidist.

## Related Issue
Fixes the issue that look and feel does not look correct.

## Motivation and Context
Look and feel was not according to layouts with short content.

## How Has This Been Tested?
On Mac and Chrome and Safari.

### DIFF
--- a/src/core/Notification/Notification.baseStyles.ts
+++ b/src/core/Notification/Notification.baseStyles.ts
@@ -11,17 +11,19 @@ export const baseStyles = (theme: SuomifiTheme) => css`
     border-radius: 4px;
 
     & .fi-notification_text-content-wrapper {
-      margin-top: 20px;
+      margin-top: 18px;
       padding-left: 20px;
       & .fi-notification_content {
         vertical-align: middle;
         ${font(theme)('bodyTextSmall')}
+        & .fi-notification_innerContent {
+          padding-top: 6px;
+        }
       }
     }
-    & .fi-notification_label {
+    & .fi-notification_heading {
       ${font(theme)('bodySemiBold')}
-      margin-top: -1px;
-      margin-bottom: ${theme.spacing.xxs};
+      margin-bottom: 0;
     }
     & .fi-notification_action-element-wrapper {
       margin: 10px 20px 22px 89px;
@@ -101,14 +103,11 @@ export const baseStyles = (theme: SuomifiTheme) => css`
     &.fi-notification--small-screen {
       padding-bottom: 20px;
       & .fi-notification_text-content-wrapper {
-        margin-top: 18px;
+        margin-top: 15px;
         padding-left: 12px;
         padding-right: 12px;
         display: flex;
         flex-direction: column;
-        & .fi-notification_label {
-          margin-top: -2px;
-        }
       }
       & .fi-notification_close-button-wrapper {
         justify-content: flex-end;
--- a/src/core/Notification/Notification.baseStyles.ts
+++ b/src/core/Notification/Notification.baseStyles.ts
@@ -3,13 +3,13 @@ import { SuomifiTheme } from '../theme';
 import { font, element } from '../theme/reset';
 
 export const baseStyles = (theme: SuomifiTheme) => css`
+  ${element(theme)}
+  ${font(theme)('bodyTextSmall')}
+  width: 100%;
+  box-shadow: ${theme.shadows.wideBoxShadow};
+  border-radius: 4px;
+  justify-content: space-between;
   &.fi-notification {
-    ${element(theme)}
-    ${font(theme)('bodyTextSmall')}
-    width: 100%;
-    box-shadow: ${theme.shadows.wideBoxShadow};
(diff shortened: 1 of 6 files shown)


LJKaski (changes requested):
There's some redundant code in the small screen variant styles. Let's get rid of that while this is open.

src/core/Notification/Notification.tsx
```diff
@@ -20,7 +20,8 @@ export const baseClassName = 'fi-notification';
 export const notificationClassNames = {
   styleWrapper: `${baseClassName}_style-wrapper`,
   content: `${baseClassName}_content`,
-  label: `${baseClassName}_label`,
+  innerContent: `${baseClassName}_innerContent`,
```
LJKaski, line 23: Would `mainContent` describe the element better? Might work with the current name just fine, but just thinking out loud here.

valger11: innerContent is child of content, thats why I selected that name...

aappoalander: The API description is `Main content of the Notification` and type is ReactNode. In addition, there is a parent node with className `textContentWrapper` which seems to be incorrect as ReactNode can contain almost anything. How about using just `contentWrapper`, `content` and, if needed, `mainContent` as described in the API. Another option would be to rename the contentWrapper to something like liveRegionWrapper etc.

valger11: I changed it to contentWrapper, that is better name.

aappoalander:
There seems to be certain cases where long text or words in e.g. close button can stretch the button so that the layout breaks due to text content having only minimal width. Also, action buttons layout is not optimal with long multi-line texts.

src/core/Notification/Notification.md
```diff
@@ -6,40 +6,42 @@ import {
 } from 'suomifi-ui-components';
 
 <>
-  <Notification closeText="Close Et turpis eget tempor. Et turpis eget tempor. Et turpis eget tempor.">
-    Et turpis eget tempor. Nam sed lacus ut lorem feugiat tincidunt.
-    Mauris quis nisi placerat.
+  <Notification closeText="This is very long close button text.">
```
aappoalander, line 9: I propose we have the first example with simple texts and change the closeText to "Close". Then the second example with error with error contents, so I would change the children to contain the word error also. Then a new, third example with long texts and the final one with simple texts and action buttons. How action button handle long texts is less irrelevant in this context.

valger11: Fixed

src/core/Notification/Notification.baseStyles.ts
```diff
@@ -3,63 +3,82 @@ import { SuomifiTheme } from '../theme';
...
       min-width: 40px;
-      padding: 7px 8px;
+      text-align: right;
+      padding: 7px 0 7px 8px;
       border: 1px solid transparent;
       border-radius: ${theme.radius.basic};
       text-transform: uppercase;
+      flex-grow: 0;
+      & svg {
+        flex-grow: 1;
```
ketsappi, line 70: Would it work as wanted when using the `fi-icon` instead of svg? Would be more consistent how the styles are defined in the library.

valger11: That is good improvement.

ketsappi: This is now changed and seems to work nicely.

ketsappi (approved):

src/core/Notification/Notification.baseStyles.ts
valger11: Refactored

src/core/Notification/Notification.baseStyles.ts
```diff
@@ -3,63 +3,75 @@ import { SuomifiTheme } from '../theme';
...
-    }
-    & .fi-notification_close-button {
-      ${font(theme)('bodyTextSmall')}
-      height: 40px;
+      margin-bottom: 0;
+      max-width: 50%;
       min-width: 40px;
+      text-align: right;
       padding: 7px 8px;
       border: 1px solid transparent;
```
LJKaski, line 65: Regarding the border: in alert the border reaches the edges of the parent element, but here there's a small margin. Is this something that has been discussed with design or should the styles be aligned?

LJKaski: Looks like this is because of the hidden overflow. Now the focus indicator fits inside the element too and is visible in its entirety.

src/core/Notification/Notification.md
ketsappi: Currently having three examples, instead of the four suggested. I think the now existing three are good to go.

ketsappi (approved):

PR #543, opened 2022-01-05
[Feature] Update suomifi-icons to 6.0.0

LJKaski:
## Description
This PR updates the suomifi-icons dependency to version 6.0.0. The new version includes changes to the baseIcon fill color, but shouldn't change anything in the ui-components library other than a very slight shade change in the icons themselves (even that won't manifest if the icons have colors specified by styles or they are accompanied by a certain colored text)

## Motivation and Context
The icons and library as a whole should reflect the style guide when it comes to colors and styles.

## How Has This Been Tested?
Tested locally in styleguidist on chrome. Might still be a good idea to use icons in a test project and see that all the styling options work there as well, just in case.

## Release notes
* Update suomifi-icons to 6.0.0


ketsappi (approved):
Tested locally in the styleguidist and in the CRA-TS test project.

Seems to work like previously. 
At quick glance in the places where the icon is used in components, `highlightDark1` token is used in e.g SearchInput.

aappoalander:
Included in #524 with other dependency upgrades.

PR #550, opened 2022-01-24
[Fix] Single- and Multiselect regexp escape

aappoalander:
## Description
This PR fixes Multi- and SingleSelect reg exp input sanitation and escapes reg exp special characters.

## Related Issue
6.1.0 Release had an issue with Single- and MultiSelect labels containing reg exp special characters and resulting into broken or not working reg exps and crashing the app.

## Motivation and Context
Single- and MultiSelect hight item label parts that match with the filtering text. This is achieved by using regexp, but typing e.g. `(` in the field crashed the application because special characters were not escaped for the reg exp.

In addition, SingleSelect uses reg exp for setting new filter values when there is a selection but no filter E.g. Hammer is selected and user presses S, results into S being used as a filter and shown in the field. Similar issue concerns this use case.

## How Has This Been Tested?
Tested using StyleGuidist build and by entering different reg exp special characters into Single- and MultiSelect input and as label texts for items.

## Release notes
**SingleSelect**:
- Fix crash when inputting special characters into input field
- Fix crash when starting to filter and item labels contain special characters

**MultiSelect**
- Fix crash when inputting special characters into input field

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,7 +139,6 @@
     "@reach/popover": "0.16.2",
     "@types/react-modal": "3.13.1",
     "classnames": "2.3.1",
-    "escape-string-regexp": "5.0.0",
     "polished": "4.1.3",
     "react-modal": "3.14.3",
     "react-popper": "2.2.5",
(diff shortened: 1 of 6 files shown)


ketsappi (approved):
Tested different regexps that caused the old version to crash but the new one did seem to escape them correctly and no issues therefore was found.

Ready to beta.

PR #555, opened 2022-01-27
[Fix] Align inner form component styles

aappoalander:
## Description
This PR removes outer margins from StatusText, HintText and Label components. Components affected by this change have been adjusted accordingly. In addition, default aria-live mode for StatusText is changed to polite.

## Motivation and Context
All exported components should follow the same principles and strategy for styling and API. This PR adjusts previously internal StatusText, HintText and Label components to be aligned with all other exported components.

## How Has This Been Tested?
Tested using Styleguidist using MacOS, Chrome and VoiceOver.

### Label
- Remove margins
- Add examples

### HintText
- Remove margins
- Add examples

### StatusText
- Remove margins
- Change aria-live default to polite
- Add examples

### TextInput
- Fix missing status text margin

### DIFF
--- a/src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
+++ b/src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
@@ -5,6 +5,10 @@ import { element, inputButton } from '../../theme/reset';
 export const baseStyles = (theme: SuomifiTheme) => css`
   &.fi-dropdown {
     display: inline-block;
+
+    & .fi-label-text {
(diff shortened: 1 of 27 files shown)


ketsappi (approved):
Tested locally in styleguidist and now it seems to work as previously, so that no visible changes are seen. _(Except the fix of the StatusText margin on TextInput.)_

LJKaski (approved):
Tested with styleguidist and CRA app. Seems to be working as intended. Good work :)

PR #556, opened 2022-02-03
[Fix] Alert layout

valger11:
## Description
Developer team found out that look and feel was't like it should be with long close buttons text. PR makes look and feel as in designs. 

## Related Issue
Fixes the issue that look and feel does not look correct.

## Motivation and Context
Look and feel was not according to layouts with long button content.

## How Has This Been Tested?
On Mac and Chrome and Safari.

## Screenshots
Screenshot 2022-02-02 at 12 36 59

### DIFF
--- a/src/core/Alert/Alert/Alert.baseStyles.ts
+++ b/src/core/Alert/Alert/Alert.baseStyles.ts
@@ -5,7 +5,6 @@ import { baseAlertBaseStyles } from '../BaseAlert/BaseAlert.baseStyles';
 
 export const baseStyles = (theme: SuomifiTheme) => css`
   ${baseAlertBaseStyles(theme)}
-  overflow: hidden;
   padding-bottom: 7px;
   & .fi-alert_icon-wrapper {
     flex: 0;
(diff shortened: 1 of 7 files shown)


src/core/Alert/Alert/Alert.baseStyles.ts
```diff
@@ -5,35 +5,31 @@ import { baseAlertBaseStyles } from '../BaseAlert/BaseAlert.baseStyles';
 
 export const baseStyles = (theme: SuomifiTheme) => css`
   ${baseAlertBaseStyles(theme)}
-
+  padding-bottom: 7px;
```
LJKaski, line 8: What is this for? It just seems to add padding that isn't there in the designs or in the previous version

valger11: This was there, because there was need to add overflow:hidden to parent. Focus ring was then cut. Padding-bottom made that visible, Later overflow:hidden was removed, but this padding, was forgot to be removed. I removed padding.

ketsappi (approved):

valger11:
There is no need to fix this

PR #561, opened 2022-02-11
[Fix] Modal text styles

valger11:
## Description
Suomi.fi services needed text styles to be correct without html-tags. This PR fixes &.fi-modal_content divs styles for plain text.

## Related Issue
Fixes the issue that look and feel does not look correct for plain text. Jira ticket is: SFIDS-574

## Motivation and Context
Look and feel does not look correct for plain text.

## How Has This Been Tested?
On Mac and Chrome and Safari with Styleguidist.

### DIFF
--- a/src/core/Modal/ModalContent/ModalContent.baseStyles.tsx
+++ b/src/core/Modal/ModalContent/ModalContent.baseStyles.tsx
@@ -1,7 +1,10 @@
 import { css } from 'styled-components';
 import { SuomifiTheme } from '../../theme';
+import { element, font } from '../../theme/reset';
 
 export const baseStyles = (theme: SuomifiTheme) => css`
+  ${element(theme)}
+  ${font(theme)('bodyText')}
   &.fi-modal_content {
     flex: 1 1 auto;
     max-height: 100%;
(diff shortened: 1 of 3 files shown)


ketsappi (dismissed):
Seems to fix the issue described and matching with the design in Abstract.

src/core/Modal/ModalContent/ModalContent.baseStyles.tsx
```diff
@@ -1,7 +1,10 @@
 import { css } from 'styled-components';
 import { SuomifiTheme } from '../../theme';
+import { element, font } from '../../theme/reset';
 
 export const baseStyles = (theme: SuomifiTheme) => css`
+  ${element(theme)}
+  ${font(theme)('bodyText')}
```
LJKaski, line 7: Should the body text size take into account the small screen option as well, since the component offers that in the API?

LJKaski: Apparently the current design guideline is that all body text is 18px regardless of viewport size although the documentation says otherwise. Let's go with the current guidelines and keep it 18px here as well :)

LJKaski (approved):

PR #563, opened 2022-02-14
[Fix/Feature] Change using of env variables to logger util

ketsappi:
Update logger. Logger to be exported util and not logging anything by default

## Description

With Webpack 5, users were having problems as the logger used inside the library uses environment variables.
This PR will remove the use of those env vars and instead will now use logger as an util.

By default, nothing is logged as we cannot know what environment and capabilities the user has.

To see the previous style of logging user can use it like this:

```tsx
import {  setLogger } from "suomifi-ui-components";

setLogger(console);
```

Or something like the following:

```tsx
import {  setLogger, Logger } from "suomifi-ui-components";

const logging: Logger = {
  log: (val: string) => {
  console.log(`[LOG] ${val}`);
  },
  warn: (val: string) => {
  console.warn(`[WAR] ${val}`);
  },
  error: (val: string) => {
  console.error(`[ERR] ${val}`);
  },
};
setLogger(logging);

```

User can alternatively use some logging library if they want to.

Styleguidist now has new page for the Logger. It will show how to use it.

## Motivation and Context

Users of the library had problems because the environment variables used that are no longer existing by default with Webpack 5.

## How Has This Been Tested?

Locally in styleguidist and CRA-TS test project

## Release notes

*Breaking change* (?)
- Fix logger. No longer using environment variables.
- Add Logger util

### DIFF
--- a/.styleguidist/logger.md
+++ b/.styleguidist/logger.md
@@ -1,0 +1,35 @@
+## Instructions
+
+Some of the components have additional logging to guide the right use of those and which was not able to be guided through API or Typescript.
+
+By default, nothing is logged. User of the library needs to enable logging when needed. User can use console or some 3rd party logging library for their needs.
+
+### Console
+
+Basic setup to log to your console.
+
+```jsx static
+import { setLogger } from 'suomifi-ui-components';
+
+setLogger(console);
+```
+
+Some customization.
+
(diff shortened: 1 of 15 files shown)


LJKaski (approved):
Looks good. Well done :)

PR #564, opened 2022-02-14
[Fix] Fixed width layout for Alert

LJKaski:
## Description
This PR changes regular Alert structure and styling in order to have a fixed width content area containing the icon, the main content and the close button. It also contains some previously overlooked changes in spacing for the small screen variant. Because of this, the `smallScreen` prop now appears in both Alert and InlineAlert.

In addition, the icon is removed from neutral variant InlineAlert as per the designs. 

## Motivation and Context
This was discussed with developers and the designers, and the intended usage was further clarified resulting in the need to change the way the Alert looks and behaves when resized.

## How Has This Been Tested?
So far tested in styleguidist + chrome on windows. Needs to be tested in a project to make sure no styles leak inside the component etc. Preferably with breakpoint-dependent styling as well.

### Alert
* Change internal structure and styling
### InlineAlert
* Add `smallScreen` prop
* Remove icon from neutral variant

### DIFF
--- a/src/core/Alert/Alert/Alert.baseStyles.ts
+++ b/src/core/Alert/Alert/Alert.baseStyles.ts
@@ -8,27 +8,24 @@ export const baseStyles = (theme: SuomifiTheme) => css`
 
   & .fi-alert_icon-wrapper {
     display: flex;
-    flex: 1 0 auto;
-    justify-content: flex-end;
   }
 
-  & .fi-alert_close-button-wrapper {
-    flex: 1 0 auto;
-    display: flex;
-    flex-wrap: nowrap;
-    box-sizing: border-box;
-    margin-top: 7px;
-    margin-right: ${theme.spacing.xs};
-    margin-bottom: ${theme.spacing.insetM};
+  & .fi-alert_style-wrapper {
+    max-width: 1110px;
+    margin: auto;
   }
 
   & .fi-alert_close-button {
     ${font(theme)('bodyTextSmall')}
     height: 40px;
-    min-width: 40px;
-    padding: 7px 8px;
+    display: inline-block;
+    padding: 7px;
+    margin-top: 7px;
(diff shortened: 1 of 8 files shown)


src/core/Alert/Alert/Alert.baseStyles.ts
```diff
@@ -8,27 +8,25 @@ export const baseStyles = (theme: SuomifiTheme) => css`
...
 
-  & .fi-alert_close-button-wrapper {
-    flex: 1 0 auto;
-    display: flex;
-    flex-wrap: nowrap;
-    box-sizing: border-box;
-    margin-top: 7px;
-    margin-right: ${theme.spacing.xs};
-    margin-bottom: ${theme.spacing.insetM};
+  & .fi-alert_style-wrapper {
```
LJKaski, line 13: Is it kosher to have the width here in the style wrapper or should we have a separate style wrapper and then e.g. a `main-content-wrapper` for the width?

ketsappi (dismissed):
Seems to work as wanted based on testing; items aligning to correct side and default max-width is found.


Tested to use breakpoints and see how Alert works with custom widths. Added the "custom-alert-width" className for Alert and then added following styles to sass.

```scss
@media (min-width: 962px) {
  .fi-alert.custom-alert-width {
    & .fi-alert_style-wrapper {
      max-width: 1100px;
    }
  }
}

@media (max-width: 767px) {
  .fi-alert.custom-alert-width {
    & .fi-alert_style-wrapper {
      max-width: 600px;
    }
  }
}
```

src/core/Alert/Alert/Alert.baseStyles.ts
ketsappi: In my opinion the current way of implementation is okay.

valger11 (approved):

PR #567, opened 2022-02-17
[Fix] Duplicate div was removed from LanguageMenu

ketsappi:
## Description

Duplicate use of MenuItems removed from the LanguageMenu. Causing problems at least because having duplicate ids and other possible issues.

## Related Issue

We got report from a service about the accessibility issue with the LanguageMenu, v6.1.1, that the duplicate ids were causing.
Issue existed in the latest develop branch also.

## Motivation and Context

We would want to have the component to be as accessible as possible.

## How Has This Been Tested?
Locally in styleguidist.
_Issue can be verified from the DOM before fix and how it is corrected after the fix, was no need to test in CRA-TS or similar test projects._

## Release notes

- Fix LanguageMenu issue with duplicate id.


LJKaski (approved):
Seems to fix the issue with no side effects. Well done :)

PR #571, opened 2022-02-24
[Fix] MultiSelect's item selection in non-controlled state

ketsappi:
## Description

Fix the MultiSelect issue for not able to select items in non-controlled state.

## Motivation and Context

We would like the component to work as intended.

## How Has This Been Tested?
Locally in styleguidist and cra-ts. 
Added new test that should ensure it is behaving as wanted in this case.

## Release notes

- Fix MultiSelect's item selection in non-controlled state


LJKaski (approved):
Seems to fix the issue well, and the tests look good as well. Great work! 💪

PR #580, opened 2022-03-10
[Fix] SingleSelect. Open list by clicking field

valger11:
## Description
In SingleSelect list didn't open from clicking field after user removed selection once.
## Motivation and Context
There was bug before and this fixes it.
## How Has This Been Tested?
Tested in StyleGuidist with mac and chrome.

### SingleSelect
* Fixes bug. In SingleSelect list didn't open from clicking field after user removed selection once.

### DIFF
--- a/src/core/Form/Select/SingleSelect/SingleSelect.tsx
+++ b/src/core/Form/Select/SingleSelect/SingleSelect.tsx
@@ -209,7 +209,7 @@ class BaseSingleSelect<T> extends Component<
     });
   };
 
-  private focusToInputAndSelectText = (openMenu: any) => {
+  private focusToInputAndSelectText = (openMenu: boolean) => {
     if (!!this.filterInputRef && this.filterInputRef.current) {
       if (openMenu) {
         this.setState({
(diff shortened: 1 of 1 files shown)


src/core/Form/Select/SingleSelect/SingleSelect.tsx
```diff
@@ -211,6 +211,9 @@ class BaseSingleSelect<T> extends Component<
 
   private focusToInputAndSelectText = () => {
     if (!!this.filterInputRef && this.filterInputRef.current) {
+      this.setState({
+        showPopover: true,
+      });
```
LJKaski, line 216: Could there be a cleaner way to do this? With only this in place, for example in the next method `focusToInputAndCloseMenu` this is now called before the menu is closed again that method. Seems like opening the menu does not suit all the situations where this method is called even though it does solve the problem in question.

valger11: Fixed.

LJKaski (approved):
Seems to do the trick and couldn't find any unwanted side effects. Well done :) Let's give Sami a chance to test this tomorrow before merging.

One possibility is that services start complaining that they only want to show the results and menu when there are at least a few characters in the filter input. But we'll cross that bridge when we get there.

ketsappi (approved):
Works nicely!

Surgically and efficiently fixed the bug.

Good job 👍

PR #582, opened 2022-03-11
[Fix] External link's screen reader text and asProp

valger11:
## Description
* In External link screen reader read labelNewWindow text even if it was not opening to new window. This is fixed.
* Prop was wrongly used in example. as={Component} is renamed to asProp={Component}
## Motivation and Context
There were bugs before and this fixes them.
## How Has This Been Tested?
Tested in StyleGuidist with mac and chrome.

### SingleSelect
* In External link screen reader read labelNewWindow text even if it was not opening to new window. This is fixed.
* Prop was wrongly used in example. as={Component} is renamed to asProp={Component}

### DIFF
--- a/src/core/Link/Link/Link.md
+++ b/src/core/Link/Link/Link.md
@@ -51,14 +51,14 @@ import { ExternalLink } from 'suomifi-ui-components';
     toNewWindow
     labelNewWindow="Opens to a new window"
   >
-    External link
+    External link opens to new window
   </ExternalLink>
   <ExternalLink
     href="https://designsystem.suomi.fi/fi/"
     toNewWindow={false}
     labelNewWindow="Opens to same window"
   >
-    External link
+    External link in same window
   </ExternalLink>
 </>;
 ```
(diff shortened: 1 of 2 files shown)


src/core/Link/Link/Link.md
```diff
@@ -45,12 +45,22 @@ import { SkipLink } from 'suomifi-ui-components';
...
+  >
+    External link
+  </ExternalLink>
+  <ExternalLink
+    href="https://designsystem.suomi.fi/fi/"
+    toNewWindow={false}
+    labelNewWindow="Opens to same window"
+  >
+    External link
+  </ExternalLink>
```
ketsappi, line 62: Should be wise to somehow separate the examples with different texts.

LJKaski: I agree. Other than this the change looks good :)

valger11: Fixed in md-file

LJKaski (approved):

ketsappi (approved):

PR #584, opened 2022-03-17
[Fix] Upgrade dependencies and do migrations

ketsappi:
## Description

- Updates the dependencies to latest available.
- Now using Webpack 5 with styleguidist.
- Migrate configs to remove removed properties and use the equivalent new ones.

Not updating `eslint-plugin-react ` yet as it requires to go through all the new rules it brings and should we fix what it will suggest, ignore those or what to do.

## Motivation and Context

To minimise the risk of more vulnerable dependencies. Unfortunately the majority of the issues still come from the `react-styleguidist` and it is already at the latest version.

## How Has This Been Tested?

Locally in styleguidist and CRA-TS test project.

## Release notes

- Update dependencies.


LJKaski (dismissed):
Tested in styleguidist, CRA project and dev mode in next.js project, and didn't encounter any problems. Good work :)

LJKaski (approved):

PR #586, opened 2022-03-21
[Fix] SingleSelect async support

LJKaski:
## Description
SingleSelect had some undesired behavior when loading list of options asynchronously. When the new list was loaded, the previous selection appeared in the input even if there was some filter text in there. This PR fixes the issue. 

As changes were made to FilterInput, MultiSelect was possibly affected as well, but I did not encounter any adverse side effects upon testing.

## Motivation and Context
This was a bug reported by a service using the components.

## How Has This Been Tested?
So far in styleguidist (Win + Chrome) with delayed changes to the item list applied to simulate asynchronous content.

### SingleSelect
* Improve behaviour with asynchronous content


ketsappi (approved):
Tested locally in styleguidist and seems to work now better than before.

valger11 (approved):

PR #588, opened 2022-03-25
[Fix] Consistent class names in SSR

jukkah:
## Description

Change rollup config by enabling styled-components babel plugin with SSR support. It generates some code on transpile time that ensures that class names are rendered consistently between renders and even in server/SSR and client. It is actually using some hash function so they will not change in every build.

Without this config class names are generated first in server and then in client but the results won't match. Then the following warning is thrown:

> Warning: Prop `className` did not match. Server: "sc-iqseJM jEyOLi sc-fFeiMQ hCxPtA fi-visually-hidden" Client: "sc-1c2dd3r-0 fSrnQw sc-1s98eq1-0 cNJqcF fi-visually-hidden"

## Motivation and Context

In vrk-yti/yti-terminology-ui we are using Next.js with SSR. This bug prevents us from updating the library so we are stuck in v6.1.1.

## How Has This Been Tested?

Tested by inspecting diff in dist directory. Also tested with vrk-yti/yti-terminology-ui.

Here is also an example diff of dist/esm/core/VisuallyHidden/VisuallyHidden.js: https://www.diffchecker.com/UgetIvED

### General Changes

Fix class names in SSR


ketsappi (dismissed):
Thanks for the contribution! 🙇 
Working nicely 👍 



## How it was tested
Tested first how to reproduce the error without this fix. I could confirm that it complained about the className:
```
Warning: Prop `className` did not match. Server: "sc-gsnTZi hXhpJV fi-button sc-cxabCf dDHYHf fi-button--secondary" Client: "sc-papXJ kNmoWv fi-button sc-fEOsli exmCqE fi-button--secondary"
```
After this tested with having this PR applied and no warnings  anymore and everything seems to work.

ketsappi (approved):

LJKaski (approved):
Works well! Thank you for the contribution! 💪

PR #589, opened 2022-03-28
[Fix] StaticChip property extension

LJKaski:
## Description
StaticChip is semantically a `span` element, yet it did not extend `HtmlSpanProps` as it should have. This PR fixes that issue, so the properties and typings are correct in the future. 

Couldn't figure out if there are some other span props aside from children that we need to omit. Spend a moment to think about that when reviewing.

## Motivation and Context
This was bug that needed to be fixed.

## How Has This Been Tested?
Tested in a Next.js test project in order to check typings and to see if the IDE suggestions work correctly. Worked as it should.

### StaticChip
* Update props to extend span props

### DIFF
--- a/src/core/Chip/StaticChip/StaticChip.tsx
+++ b/src/core/Chip/StaticChip/StaticChip.tsx
@@ -1,7 +1,7 @@
 import React, { Component } from 'react';
 import classnames from 'classnames';
 import { default as styled } from 'styled-components';
-import { HtmlSpan } from '../../../reset';
+import { HtmlSpan, HtmlSpanProps } from '../../../reset';
 import {
   BaseChipProps,
   baseClassName,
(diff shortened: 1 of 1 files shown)


ketsappi:
> Set props to extend span props

Maybe to use "Update" word instead to be more consistent the verbs used in the past in the release notes.

> Update props to extend HtmlSpan's props

ketsappi (dismissed):
Tested in CRA-TS test project and seems to work as wanted.

LJKaski:
> > Set props to extend span props
> 
> Maybe to use "Update" word instead to be more consistent the verbs used in the past in the release notes.
> 
> > Update props to extend HtmlSpan's props

Good point. Edited the commit message when rebasing. Should now be in order :)

ketsappi (approved):

PR #594, opened 2022-03-31
[Feature] Disable coverage collection fo Jest

ketsappi:
We were not utilizing the coverage reports and it just caused unwanted
delays for testing. It can be enabled again if there will rise actual
need for it.

## Description

Disable Jest coverage report.

## Motivation and Context

We were not utilizing the coverage reports and it just caused unwanted delays for testing. It can be enabled again if there will rise actual need for it.

## How This Can Be Tested?

Before you checkout this PR branch, run `yarn test` in some other branch, e.g `develop`.
You can see that the coverage report will be run after the tests.

Then checkout this PR and after running the tests you can no longer see the coverage report and it will finish sooner.

## Release notes

_No need to have in release notes as this affects only internal developing of the library._

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -7,7 +7,7 @@ module.exports = {
   // Automatically clear mock calls and instances between every test
   clearMocks: true,
 
-  collectCoverage: true,
+  collectCoverage: false,
   // The directory where Jest should output its coverage files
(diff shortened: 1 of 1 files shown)


LJKaski (approved):
Good initiative! This makes the results much easier to read.

PR #595, opened 2022-04-04
[Fix] Toast, Notification and Textarea background color

LJKaski:
## Description
This PR adds background `whiteBase` as background color to Toast, Notification and the input element in Textarea. Previously those took on the background color of their parent, which was not in line with how the rest of the components work. 

## Motivation and Context
Component styles should behave predictably across the library.

## How Has This Been Tested?
Tested in styleguidist on Chrome on windows by putting the components inside a div with a background color.

### Toast
* Add `whiteBase` as default background color 
### Notification
* Add `whiteBase` as default background color
### Textarea
* Add background color for the input element

### DIFF
--- a/src/core/Notification/Notification.baseStyles.ts
+++ b/src/core/Notification/Notification.baseStyles.ts
@@ -11,6 +11,7 @@ export const baseStyles = (theme: SuomifiTheme) => css`
   display: block;
   justify-content: space-between;
   padding-bottom: 10px;
+  background-color: ${theme.colors.whiteBase};
   &.fi-notification {
     & .fi-notification_style-wrapper {
       padding: 0 32px 10px 40px;
--- a/src/core/Notification/Notification.baseStyles.ts
+++ b/src/core/Notification/Notification.baseStyles.ts
@@ -11,8 +11,9 @@ export const baseStyles = (theme: SuomifiTheme) => css`
   display: block;
   justify-content: space-between;
   padding-bottom: 10px;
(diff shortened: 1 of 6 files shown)


src/core/Notification/Notification.baseStyles.ts
```diff
@@ -11,6 +11,7 @@ export const baseStyles = (theme: SuomifiTheme) => css`
   display: block;
   justify-content: space-between;
   padding-bottom: 10px;
+  background-color: ${theme.colors.whiteBase};
```
LJKaski, line 14: Not quite sure if this should be inside the main class or not. Could this get overridden too easily by accident?

ketsappi: Hmm.. not quite sure but maybe under the `fi-notification` would be better place.

And if we decide to do that then same change is to be done with Toast.

ketsappi (commented):
Works nicely.

Tested in styleguidist and added color for parent element to see if color leaks. It didn't leak anymore when compared to older version.

```tsx
<div style={{background: "red"}}>
  {/* Elements that are to be tested */}
</div>
```

src/core/Notification/Notification.baseStyles.ts
LJKaski: I moved the styles inside the main classes. There were no styles inside the main class in Toast, so created the selector as well.

ketsappi (approved):

PR #596, opened 2022-04-06
[Feature] Tooltip

aappoalander:
## Description
This PR adds the Tooltip component and support for Tooltip use with Label component.

## Motivation and Context
Labels may need additional information for some user groups, while others will have better experience with less information. Tooltips solve this issue by implementing a sort of inline expander that adds content between the label and the following component the label refers to.

## How Has This Been Tested?
TODO, only tested quickly with Styleguidist / Chrome / VoiceOver on MacOS.

## Release notes
Tooltip
- Add Tooltip component

Label component
- Add tooltipComponent, contentClassName, wrapperProps and style props
- Breaking Changes: 
  - LabelProps now extend HtmlSpanProps instead of HtmlDivProps
  - Props are passed to inner content wrapper instead of the top most element, except for className and wrapperProps.
  - Remove labelSpanProps
  - ClassName is passed to top most element instead of inner content wrapper


src/core/Form/Label/Label.tsx
```diff
@@ -40,42 +60,70 @@ const labelTextClassNames = {
...
+    return (
+      <HtmlDivWithRef
+        className={classnames(className, baseClassName)}
+        {...wrapperProps}
+        forwardedRef={(ref: SetStateAction<HTMLDivElement | null>) =>
+          setWrapperRef(ref)
+        }
+      >
+        {labelMode === 'hidden' ? (
+          <VisuallyHidden>
```
aappoalander, line 98: Do we need to pass some props to VisuallyHidden as well?

src/core/Form/Label/Label.tsx
```diff
@@ -40,42 +60,70 @@ const labelTextClassNames = {
...
+    const [wrapperRef, setWrapperRef] = useState<HTMLDivElement | null>(null);
+
+    function getTooltipComponent(
+      tooltipComponent: ReactElement | undefined,
+    ): ReactNode {
+      if (isValidElement(tooltipComponent)) {
+        return cloneElement(tooltipComponent, {
+          anchorElement: wrapperRef,
+          // trick to force tooltip to rerender every time when label changes.
+          key: Date.now(),
```
aappoalander, line 83: This needs further investigation. In cases the Label is displayed inline, and it's children change, re-render might not be automatically triggered and the indication arrow position is not updated. Optimal solution would be to leave out the key, but if it does not work in all cases, maybe use something like derived state and set an uuid everytime children change?

src/core/Tooltip/Tooltip.text.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { render } from '@testing-library/react';
+import { Tooltip } from './Tooltip';
+import { axeTest } from '../../utils/test';
+
+describe('props', () => {
```
aappoalander, line 6: Further tests are needed to cover all necessary props.

src/core/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,196 @@
...
+      forwardedRef,
+    } = this.props;
+
+    const open = 'open' in this.props ? propsOpen : this.state.open;
+
+    return (
+      <>
+        <TooltipButton
+          className={classnames(baseClassName, toggleButtonClassName)}
+          ref={forkRefs(this.toggleButtonRef, forwardedRef)}
```
aappoalander, line 170: forkRef needs thorough testing. Seems to be working, but better to test well.

src/core/Tooltip/TooltipContent/TooltipContent.baseStyles.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+export const baseStyles = (arrowOffsetPx: number, theme: SuomifiTheme) => css`
+  margin-top: 10px;
+  margin-bottom: 10px;
+  ${theme.typography.bodyTextSmall};
+  position: relative;
+  border: 1px solid ${theme.colors.depthDark2};
+  border-radius: ${theme.radius.basic};
+  background-color: ${theme.colors.highlightLight4};
+  padding: 20px 43px 20px 20px;
```
aappoalander, line 12: Should we use inset spacing tokens?

src/core/Tooltip/Tooltip.text.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
```
ketsappi, line 1: Just a note to remember for us to rename the file to be `Tooltip.test.tsx`, small typo in name as it is not `.text.`.

src/core/Tooltip/TooltipContent/TooltipContent.baseStyles.tsx
ketsappi: True, maybe we could use `insetXxl` token here.

src/core/Form/Label/Label.tsx
```diff
@@ -1,34 +1,54 @@
...
+  /** Wrapper class name for styling and customizing */
+  className?: string;
   /** Label element content */
   children: ReactNode;
-  /** Custom class name for styling and customizing  */
-  className?: string;
+  /** Content class name for styling and customizing */
+  contentClassName?: string;
+  /** Content style for styling and customizing */
+  style?: CSSProperties;
```
ketsappi, line 36: As the `style` is for the content, should it be `contentStyle` to be consistent?

src/core/Form/Label/Label.tsx
```diff
@@ -1,34 +1,54 @@
...
+  wrapperProps?: Omit<HtmlDivProps, 'as' | 'className'>;
   /** Render the wrapping element as another element
    *
    * @default 'label'
    */
   asProp?: asPropType;
   /** Text to mark the field as optional. Shown after labelText and wrapped in parentheses. */
   optionalText?: string;
+  /** Tooltip component for label */
+  tooltipComponent?: ReactElement;
```
ketsappi, line 51: Instead of `ReactElement` we might be able to be more specific and to support just our own tooltip.

```tsx
tooltipComponent?: typeof Tooltip;
```

src/core/Form/Label/Label.tsx
```diff
@@ -40,42 +60,70 @@ const labelTextClassNames = {
...
+                </HtmlSpan>
+              )}
             </HtmlSpan>
-          )}
-        </HtmlSpan>
-      )}
-    </HtmlDiv>
-  ),
+            {!!tooltipComponentProp &&
+              getTooltipComponent(tooltipComponentProp)}
```
ketsappi, line 121: Quickly tested to just use like the following, **IF** using the `typeof Tooltip` for the prop as in my other comment.
```tsx
{!!tooltipComponentProp && tooltipComponentProp}
```

So far it seems to work without the complex logic for checking validity and need for cloning. But needs more testing to see if hiccups will occur.

src/core/Tooltip/Tooltip.md
```diff
@@ -0,0 +1,53 @@
...
+        >
+          <Heading variant="h5" as="h2">
+            Tooltip
+          </Heading>
+          <Text>Tooltip text for a text that requires a tooltip</Text>
+        </Tooltip>
+      </div>
+    </>
+  );
+}
```
ketsappi, line 50: Pasting the typed version here for anyone testing this to save time as CRA-TS will not happily run when tried to test with copy-pasting this as typings are missed AND styleguidist doesn't seem to like if adding them to example.

```tsx
  // Tooltip should always render when tooltip button position changes
  const [longText, setLongText] = React.useState(false);
  const [anchorElement, setAnchorElement] = React.useState<any>(null);
  const [open, setOpen] = React.useState(false);

  const getLabelText = (long: boolean) =>
    long
      ? "Some longer text that requires a tooltip"
      : "Some text that requires a tooltip";

  return (
    <>
      <Button onClick={() => setLongText(!longText)}>Toggle text</Button>
      <div ref={(ref) => setAnchorElement(ref)}>
        <Text style={{ verticalAlign: "middle" }}>
          {getLabelText(longText)}
        </Text>
        <Tooltip
          anchorElement={anchorElement}
          ariaToggleButtonLabelText={`${getLabelText(
            longText
          )}, additional information`}
          ariaCloseButtonLabelText={`${getLabelText(
            longText
          )}, close additional information`}
          open={open}
          onToggleButtonClick={() => setOpen(!open)}
          onCloseButtonClick={() => setOpen(false)}
          key={String(longText)}
        >
          <Heading variant="h5" as="h2">
            Tooltip
          </Heading>
          <Text>Tooltip text for a text that requires a tooltip</Text>
        </Tooltip>
      </div>
    </>
  );

```

src/core/Form/Label/Label.tsx
ketsappi: I assume that this means, that should the tooltip's text to be read by screenreader even if the label is visually hidden.

src/core/Tooltip/TooltipContent/TooltipContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { baseStyles } from './TooltipContent.baseStyles';
+
+const baseClassName = 'fi-tooltip';
+
+const tooltipContentClassNames = {
+  closeButton: `${baseClassName}_close-button`,
+  closeButtonIcon: `${baseClassName}_close-button_icon`,
+};
+
+interface TooltipContentProps {
```
ketsappi, line 22: Should we extend `HtmlDiv` in the interface?

src/core/Tooltip/TooltipContent/TooltipContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+class BaseTooltipContent extends Component<TooltipContentProps & InnerRef> {
+  render() {
+    const {
+      children,
+      arrowOffsetPx,
+      onCloseButtonClick,
+      forwardedRef,
+      ariaCloseButtonLabelText,
+      ...rest
```
ketsappi, line 45: We should rename this to `...passProps` just to be consistent.

src/utils/common/common.ts
```diff
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { MutableRefObject, Ref } from 'react';
+import { getLogger } from '../..';
```
ketsappi, line 2: I suggest that we should use explicit path for the logger.

```tsx
import { getLogger } from '../log/logger';
```

src/core/Tooltip/TooltipContent/TooltipContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const TooltipContent = forwardRef(
+  (props: TooltipContentProps, ref: RefObject<HTMLDivElement>) => {
+    const { ...passProps } = props;
+    return (
+      <SuomifiThemeConsumer>
+        {({ suomifiTheme }) => (
+          <StyledTooltipContent
+            theme={suomifiTheme}
+            forwardedRef={ref}
+            {...passProps}
```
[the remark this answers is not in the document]
LJKaski: Same seems to go for TooltipButton.

src/core/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,196 @@
...
+    return 0;
+  };
+
+  private setContentArrowOffset = () => {
+    this.setState({ contentArrowOffsetPx: this.calculateContentArrowOffset() });
+  };
+
+  componentDidMount() {
+    window.addEventListener('resize', this.setContentArrowOffset);
+    const anchorRefObserver = new ResizeObserver(() => {
```
ketsappi, line 100: It seems that at least the tests don't like the ResizeObserver.

>  ReferenceError: ResizeObserver is not defined

Need to look a little how this can be fixed.

ketsappi: Couldn't yet another way than mock the ResizeObserver in the tests.

Adding the following to our ` jest/jest.setup.js`:
```js
global.ResizeObserver = jest.fn().mockImplementation(() => ({
    observe: jest.fn(),
    unobserve: jest.fn(),
    disconnect: jest.fn(),
}));

```

src/core/Form/Label/Label.tsx
aappoalander: Not sure what exactly this need would be, but it looks like VisuallyHidden takes no props, so maybe this can be tackled later on. I'd just make sure that everything works as expected after these changes - including screen readers. Tooltip itself should not be shown imo if label mode is hidden.

src/core/Form/Label/Label.tsx
aappoalander: Shouldn't you support JS as well? If yes, you probably should not remove validity checks and cannot rely on TS types, but rather something that works with JS and React too. Cloning is needed for injecting / merging props so it's probably not possible to remove it.

src/core/Form/Label/Label.tsx
aappoalander: There are cases where tooltips will not be correctly updated if tooltip anchors (text length, container size or what ever the anchor element is, might be inline, block, flex etc) gets updated, so there needs to be a way to force a rerender. Key was used to achieve this and reposition the css arrow. If there is an alternative way that works 100% of the time, the additional logic can and should be removed.

src/core/Tooltip/TooltipContent/TooltipContent.tsx
aappoalander: when extending, it might be worthwhile to make sure potentially breaking layout props are left out as the whole structure relies on being inline + block / inline-block elements.

src/core/Tooltip/TooltipContent/TooltipContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const {
+      children,
+      arrowOffsetPx,
+      onCloseButtonClick,
+      forwardedRef,
+      ariaCloseButtonLabelText,
+      ...rest
+    } = this.props;
+    return (
+      <div {...rest} ref={forwardedRef}>
```
ketsappi, line 48: Maybe we should use `HtmlDiv` instead?

src/core/Tooltip/TooltipContent/TooltipContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+  MouseEvent,
+  KeyboardEvent,
+} from 'react';
+import { default as styled } from 'styled-components';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
+import { HtmlButton } from '../../../reset/HtmlButton/HtmlButton';
+import { Icon } from '../../Icon/Icon';
+import { baseStyles } from './TooltipContent.baseStyles';
+
+const baseClassName = 'fi-tooltip';
```
ketsappi, line 15: Might be good idea that we will have some className for the "component", so we would have `fi-tooltip_content`.

src/core/Tooltip/Tooltip.text.tsx
ketsappi: Renamed.

src/core/Tooltip/TooltipContent/TooltipContent.tsx
ketsappi: Renamed.

src/core/Tooltip/TooltipContent/TooltipContent.tsx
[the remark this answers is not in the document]
ketsappi: These are now changed.

src/utils/common/common.ts
ketsappi: Hmm.. interesting. When changing this to suggested `import { getLogger } from '../log/logger';` following snapshots will have changes:

```
src/core/Expander/Expander/__snapshots__/Expander.test.tsx.snap
src/core/Expander/ExpanderGroup/__snapshots__/ExpanderGroup.test.tsx.snap
src/core/Expander/ExpanderTitle/__snapshots__/ExpanderTitle.test.tsx.snap
src/core/Form/Checkbox/__snapshots__/CheckboxGroup.test.tsx.snap
src/core/Form/RadioButton/__snapshots__/RadioButtonGroup.test.tsx.snap
src/core/Form/SearchInput/__snapshots__/SearchInput.test.tsx.snap
src/core/Form/Select/MultiSelect/MultiSelect/__snapshots__/MultiSelect.test.tsx.snap
src/core/Form/Select/SingleSelect/__snapshots__/SingleSelect.test.tsx.snap
```

src/core/Tooltip/TooltipContent/TooltipContent.tsx
ketsappi: TooltipContent now has the `fi-tooltip_content` added.

src/core/Tooltip/Tooltip.text.tsx
ketsappi: More tests are now added.

src/core/Form/Label/Label.tsx
ketsappi: Okay, good notes. Will then leave as it is.

src/core/Tooltip/TooltipContent/TooltipContent.tsx
ketsappi: Changed it now to extend `HtmlDiv`.

src/core/Form/Label/Label.tsx
LJKaski: Makes sense. In most cases the style prop is passed to the outermost element, and with that not being the case here, it would make sense to communicate that.

src/core/Form/Label/Label.tsx
ketsappi: Leaving this unchanged.

src/core/Tooltip/TooltipContent/TooltipContent.tsx
ketsappi: This is now changed.

src/core/Form/Label/Label.tsx
ketsappi: This is now changed.

src/core/Form/Label/Label.tsx
ketsappi: We will leave it like this for now.

ketsappi (approved):

PR #598, opened 2022-04-12
[Fix] Update babel-plugin-styled-components to 2.0.7

LJKaski:
## Description
The recent changes to the build configuration caused some styles to be broken. The cause was tracked down to be `babel-plugin-styled-components` which didn't handle some styles correctly. The bug was reported in an issue and the developers released a patch shortly after fixing the issue.

This PR updates the `babel-plugin-styled-components` to its latest version of 2.0.7, which seems to be fixing the issue.

## Motivation and Context
There was a bug and this update fixes it.

## How Has This Been Tested?
Tested in a CRA project on Windows with Chrome and Firefox.

### Dependencies
* Resolve `babel-plugin-styled-components` version to 2.0.7

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "suomifi-ui-components",
-  "version": "7.0.0-beta.1-babel",
+  "version": "7.0.0-beta.1",
   "description": "Suomi.fi UI component library",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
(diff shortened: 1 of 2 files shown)


jukkah:
LGTM :+1: Also tested with [vrk-yti/yti-terminology-ui](https://github.com/VRK-YTI/yti-terminology-ui/tree/v2).

ketsappi (approved):
Seems to work as wanted.

## How it was tested
First I installed latest beta where the issue exists, `7.0.0-beta.1`. And I could confirm on CRA-TS test project that SingleSelect's clear button was not in correct place.

Applied this PR and issue seems to be fixed.

PR #600, opened 2022-04-14
[Feature] Make MultiSelect use InputToggleButton

riitasointi:
## Description

This PR changes the MultiSelect component so that it uses our InputToggleButton component instead of a CSS-based way to display the caret icon.

This adds a new behavior to the component: the popover can now be closed (in addition to opening) by pressing the caret icon. This was not possible previously.

The height of the InputToggleButton was also set to 100% of parent height to avoid "dead zones"

Also fixed a flaky test in SingleSelect. Used waitFor() when determining what happens when the input is blurred. 

## Motivation and Context

We wanted to unify the implementations of caret icons across the different select components. The use of the InputToggleButton was copied from SingleSelect. 

## How Has This Been Tested?

With Chrome & Firefox, Mac VoiceOver in Styleguidist and a CRA app

## Screenshots:

https://user-images.githubusercontent.com/17459942/163395066-6a5e86d4-0edf-4c50-983a-f184f4aadeee.mov

### MultiSelect
* Use internal InputToggleButton component to display the caret icon. 
* This adds a new behavior to MultiSelect: the popover can now be closed (in addition to opening) by pressing the caret icon
* Changed cursor from default to pointer in popover item list

### SingleSelect
* Changed cursor from default to pointer in popover item list

### Dropdown
* Changed cursor from default to pointer in popover item list

### DIFF
--- a/src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.baseStyles.tsx
+++ b/src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.baseStyles.tsx
@@ -9,19 +9,6 @@ export const baseStyles = (theme: SuomifiTheme) => css`
   &.fi-multiselect {
     & .fi-filter-input_input-element-container {
       position: relative;
-
-      &:before {
-        content: '';
-        position: absolute;
-        top: 50%;
-        right: 14px;
-        margin-top: -3px;
-        border-style: solid;
-        border-color: ${theme.colors.blackBase} transparent transparent
-          transparent;
-        border-width: 6px 4px 0 4px;
-        pointer-events: none;
-      }
     }
     & .fi-filter-input_input {
       padding-right: 36px;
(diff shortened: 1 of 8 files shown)


src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.baseStyles.tsx
```diff
@@ -9,19 +9,6 @@ export const baseStyles = (theme: SuomifiTheme) => css`
   &.fi-multiselect {
     & .fi-filter-input_input-element-container {
       position: relative;
-
```
LJKaski, line 12: Why are these removed?

LJKaski, line 12: Why are these removed?

LJKaski (approved):
Looks good to me. Good work 👍

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "suomifi-ui-components",
-  "version": "7.0.0-beta.1",
+  "version": "7.0.0-beta.2",
```
ketsappi, line 3: Normally we wouldn't want to change the version during "normal" PRs.
Separate PR for increasing version number.

But this might be just some mishap when merging/rebasing and does not cause any real problem.

ketsappi (approved):
Seems to work great.

Good job!

ketsappi:
Maybe release notes should be changed a little in the description.

Thinking from the perspective of what the library user would want to read and what is really needed to be told for them.

package.json
riitasointi: Yes, this is a bit weird since it's a change made in a previous PR, not this one

LJKaski (approved):

PR #602, opened 2022-04-21
[Fix] RadioButton & RadioButtonGroup adjustments

ketsappi:
## Description

- Adjust spacings for RadioButtonGroup
- Decrease font size of RadioButton
- Renamed RadioButtonGroup's `hintText` to be `groupHinText` to not confuse with the RadioButton's one and to be in same way as in CheckboxGroup.
- Use the Label and HintText components in RadioButtonGroup  instead of custom implementations
- Use the HintText components in RadioButton instead of custom implementations
- Add `optionalText` prop for RadioButtonGroup 

## Motivation and Context

While doing CheckboxGroup, some differences came up that this PR will fix to be more consistent with each other. 
_And something new also came up while doing this PR that will need to be fixed to CheckboxGroup. Will do that in separate PR._

## How Has This Been Tested?
Locally in styleguidist so far.

### Radiobutton
- Decrease font size

### RadiobuttonGroup

- **Breaking change:** Rename `hintText` prop to `groupHintText`.
- Update spacings to match with the design.
- New `optionalText` prop added.

### DIFF
--- a/src/core/Form/RadioButton/RadioButton.baseStyles.tsx
+++ b/src/core/Form/RadioButton/RadioButton.baseStyles.tsx
@@ -71,6 +71,7 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       ${theme.typography.bodyTextSmall};
     }
     & .fi-radio-button_label {
+      ${font(theme)('bodyTextSmall')}
       position: relative;
       display: inline-block;
       cursor: pointer;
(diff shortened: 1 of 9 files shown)


src/core/Form/RadioButton/RadioButton.baseStyles.tsx
```diff
@@ -71,11 +70,14 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       ${theme.typography.bodyTextSmall};
     }
     & .fi-radio-button_label {
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
```
LJKaski, line 75: Is this just for the vertical alignment? Is there a reason for using this method instead of, for example, setting the line height to match the element height? I think that one is used elsewhere.

ketsappi: Yes, it seems it was just for aligning the label vertically.

This is now changed to use `line-height` instead; same for the large variant.

LJKaski (approved):
Looks good! Great work :)

PR #603, opened 2022-04-25
[Feature] Disabled state to SingleSelect & MultiSelect

riitasointi:
## Description

This PR adds the disabled state to SingleSelect and MultiSelect. The disabled prop is optional and is passed further into subcomponents like FIlterInput, InputClearButton and InputToggleButton to provide correct behavior. 

When disabled, the components are non-focusable and out of the tab order. In MultiSelect, however, the chip list items are still focusable (assuming that there are preselected values in the input).

## Related Issue

This PR is built on top the commits of PR #600 which adds the InputToggleButton to MultiSelect.

## Motivation and Context

There is a need to use these components in a disabled state.

## How Has This Been Tested?

On Mac. Chrome, Safari, Firefox. VoiceOver. In Styleguidist. 

## Screenshots

SingleSelect
image

MultiSelect
image
image

### SingleSelect
* Add optional prop `disabled`

### MultiSelect
* Add optional prop `disabled`

### DIFF
--- a/src/core/Form/InputClearButton/InputClearButton.baseStyles.ts
+++ b/src/core/Form/InputClearButton/InputClearButton.baseStyles.ts
@@ -25,4 +25,16 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       }
     }
   }
+
+  &[disabled],
+  &:disabled {
+    cursor: not-allowed;
+    &.fi-input-clear-button {
+      & .fi-input-clear-button_icon {
+        & .fi-icon-base-fill {
+          fill: ${theme.colors.depthBase};
+        }
+      }
+    }
+  }
 `;
--- a/src/core/Form/InputClearButton/InputClearButton.baseStyles.ts
+++ b/src/core/Form/InputClearButton/InputClearButton.baseStyles.ts
@@ -24,12 +24,10 @@ export const baseStyles = (theme: SuomifiTheme) => css`
         fill: ${theme.colors.highlightDark1};
       }
     }
-  }
 
-  &[disabled],
-  &:disabled {
-    cursor: not-allowed;
-    &.fi-input-clear-button {
+    &[disabled],
+    &:disabled {
+      cursor: not-allowed;
       & .fi-input-clear-button_icon {
         & .fi-icon-base-fill {
           fill: ${theme.colors.depthBase};
--- a/src/core/Form/InputToggleButton/InputToggleButton.baseStyles.ts
+++ b/src/core/Form/InputToggleButton/InputToggleButton.baseStyles.ts
@@ -20,12 +20,10 @@ export const baseStyles = (theme: SuomifiTheme) => css`
         fill: ${theme.colors.blackBase};
       }
     }
-  }
 
-  &[disabled],
-  &:disabled {
-    cursor: not-allowed;
-    &.fi-input-toggle-button {
+    &[disabled],
+    &:disabled {
+      cursor: not-allowed;
       & .fi-input-toggle-button_icon {
         & .fi-icon-base-fill {
           fill: ${theme.colors.depthBase};
--- a/src/core/Form/InputToggleButton/InputToggleButton.baseStyles.ts
+++ b/src/core/Form/InputToggleButton/InputToggleButton.baseStyles.ts
@@ -21,4 +21,16 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       }
     }
   }
+
+  &[disabled],
+  &:disabled {
+    cursor: not-allowed;
+    &.fi-input-toggle-button {
+      & .fi-input-toggle-button_icon {
+        & .fi-icon-base-fill {
+          fill: ${theme.colors.depthBase};
+        }
+      }
+    }
+  }
 `;
--- a/src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.md
+++ b/src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.md
@@ -169,9 +169,9 @@ const foods = [
       items={foods}
       labelText="Food"
       visualPlaceholder="Selection disabled"
-      ariaSelectedAmountText=""
-      ariaOptionsAvailableText=""
-      ariaOptionChipRemovedText=""
+      ariaSelectedAmountText="foods selected"
+      ariaOptionsAvailableText="options available"
+      ariaOptionChipRemovedText="removed"
     />
 </>
 ```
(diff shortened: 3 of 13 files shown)


src/core/Form/InputToggleButton/InputToggleButton.baseStyles.ts
```diff
@@ -20,4 +21,16 @@ export const baseStyles = (theme: SuomifiTheme) => css`
...
+  &:disabled {
+    cursor: not-allowed;
+    &.fi-input-toggle-button {
+      & .fi-input-toggle-button_icon {
+        & .fi-icon-base-fill {
+          fill: ${theme.colors.depthBase};
+        }
+      }
+    }
+  }
```
ketsappi, line 35: Not a big deal, but I'll suggest that we move this inside of the so called base class, to `fi-input-toggle-button` that is few lines above.

src/core/Form/InputClearButton/InputClearButton.baseStyles.ts
```diff
@@ -25,4 +25,16 @@ export const baseStyles = (theme: SuomifiTheme) => css`
...
+  &:disabled {
+    cursor: not-allowed;
+    &.fi-input-clear-button {
+      & .fi-input-clear-button_icon {
+        & .fi-icon-base-fill {
+          fill: ${theme.colors.depthBase};
+        }
+      }
+    }
+  }
```
ketsappi, line 39: Same as with toggle button, I'll suggest that we move this inside of the so called base class, to `fi-input-clear-button` that is few lines above.

src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.md
```diff
@@ -148,3 +148,30 @@ const animals = [
...
+
+<>
+    <MultiSelect
+      disabled={true}
+      items={foods}
+      labelText="Food"
+      visualPlaceholder="Selection disabled"
+      ariaSelectedAmountText=""
+      ariaOptionsAvailableText=""
+      ariaOptionChipRemovedText=""
```
ketsappi, line 174: Just a little snag, but maybe we could add some values here and show good example use case with no empty values.
Of course in this case it does not actually matter, but maybe it looks little better and complete example. 🤔

src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.baseStyles.tsx
```diff
@@ -34,13 +21,6 @@ export const baseStyles = (theme: SuomifiTheme) => css`
   }
 
   &.fi-multiselect--open {
-    & .fi-filter-input_input-element-container {
-      &:before {
-        border-color: transparent transparent ${theme.colors.blackBase}
-          transparent;
-        border-width: 0 4px 6px 4px;
-      }
-    }
```
ketsappi, line 43: I think you mentioned about the removal of some styles, but not sure was it this PR.
Can you tell again why this style is no longer needed for me and maybe for other reviewers also.

riitasointi: This is actually from a commit that was in the previous PR #600. That pseudo element was the CSS-based caret-icon

src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.md
riitasointi: Added!

src/core/Form/InputToggleButton/InputToggleButton.baseStyles.ts
riitasointi: I don't think it works though? `.fi-input-toggle-button` is the class for the button element, not the SVG inside it.

src/core/Form/InputClearButton/InputClearButton.baseStyles.ts
riitasointi: See my comment above

src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.tsx
```diff
@@ -345,6 +359,34 @@ class BaseMultiSelect<T> extends Component<
...
+      }
+    }
+  }
+
+  private disableItems(items: MultiSelectData[]): MultiSelectData[] {
+    return items.map((item) => ({
+      ...item,
+      disabled: true,
+    }));
+  }
```
LJKaski, line 388: I think this could be cleaner and more uniform with other functions, if it was an arrow function, like this: 

```suggestion
   private disableItems = (items: MultiSelectData[]): MultiSelectData[] =>
    items.map((item) => ({
      ...item,
      disabled: true,
    }));
```

src/core/Form/InputToggleButton/InputToggleButton.baseStyles.ts
ketsappi: Seems to work, when I tested to move it like the following:

```tsx
&.fi-input-toggle-button {
  display: flex;
  height: 100%;
  & .fi-input-toggle-button_icon {
    pointer-events: none;
    width: 10px;
    height: 10px;
    & .fi-icon-base-fill {
      fill: ${theme.colors.blackBase};
    }
  }

  &[disabled],
  &:disabled {
    cursor: not-allowed;
    & .fi-input-toggle-button_icon {
      & .fi-icon-base-fill {
        fill: ${theme.colors.depthBase};
      }
    }
  }
}
```

riitasointi: Ah you meant it like that. Makes very much sense, changed it like so!

src/core/Form/InputClearButton/InputClearButton.baseStyles.ts
riitasointi: Changed this as well based on comments above

src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.tsx
riitasointi: Sure, changed it!

src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.md
```diff
@@ -148,3 +148,30 @@ const animals = [
...
+  }
+];
+
+<>
+    <MultiSelect
+      disabled={true}
+      items={foods}
+      labelText="Food"
+      visualPlaceholder="Selection disabled"
+      ariaSelectedAmountText="foods selected"
```
LJKaski, line 172: I think this should be more generic, because it should probably be uniform across the service using MultiSelect despite the context of the individual MultiSelect. So probably just 'items selected'.

riitasointi: Done!

LJKaski (approved):

PR #604, opened 2022-04-26
[Fix] Expander group - CSS selector for .fi-icon

riitasointi:
## Description

This PR fixes an issue where a generic CSS selector in Expander group component targeted all icons (.fi-icon) inside the Expander component and changed their color. This misbehaving CSS selector can be removed altogether since the correct color for the toggle icon is inherited from its parent. 

## Motivation and Context

Component stylings need to be robust and predictable and not cause side effects.

## How Has This Been Tested?

On MacOS: Chrome, Safari & Firefox. Manually in Styleguidist.

### Expander group
* Fix CSS selector which targeted all `.fi-icon` elements inside the component and determined their color

### DIFF
--- a/src/core/Expander/ExpanderGroup/ExpanderGroup.baseStyles.tsx
+++ b/src/core/Expander/ExpanderGroup/ExpanderGroup.baseStyles.tsx
@@ -22,9 +22,6 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       transition: margin ${`${theme.transitions.basicTime}
         ${theme.transitions.basicTimingFunction}`};
 
-      & .fi-icon {
-        color: ${theme.colors.highlightBase};
-      }
       & > {
         border-radius: 0;
       }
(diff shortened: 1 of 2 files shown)


src/core/Expander/ExpanderGroup/ExpanderGroup.baseStyles.tsx
```diff
@@ -22,9 +22,6 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       transition: margin ${`${theme.transitions.basicTime}
         ${theme.transitions.basicTimingFunction}`};
 
-      & .fi-icon {
```
LJKaski, line 25: Since after the removal the correct styling is passed through the color prop of the parent element (ExpanderTitleButton), would it be more robust to explicitly add the correct styling via the `fi-icon` class name? Probably not gonna make a big difference in practice, but wanted to put the thought out there.

LJKaski: As discussed elsewhere, the proposed change would not provide much additional value, so we can leave this as it is with the current changes.

LJKaski (approved):

LJKaski (approved):
Seems to work well :)

PR #606, opened 2022-04-26
[Release] 7.0.0

LJKaski:
## Description

This PR merges the current develop branch with the master branch in anticipation of the release of version 7.0.0

React version 18 is not yet fully supported because of issues with components using Reach UI. This needs to be and will be stated in the release notes.

Other release notes available in PR #605


ketsappi (approved):
Seems to work, except the case with Dropdowns when using with React v18 in development mode.
Under the hood it uses reach-ui where the issue remains unsolved at the moment.

We should highlight this possible issue in the release notes here in this PR already.

PR #607, opened 2022-04-27
Feature/loading spinner

Riippi:
## Related Issue

Closes #


src/core/LoadingSpinner/LoadingSpinner.md
```diff
@@ -0,0 +1,95 @@
+```js
+import { LoadingSpinner } from 'suomifi-ui-components';
+
+<>
+  <div aria-live="assertive">
```
LJKaski, line 5: Does this serve any purpose here as things presented inside it appear to be static?

src/core/LoadingSpinner/LoadingSpinner.md
```diff
@@ -0,0 +1,95 @@
...
+import {
+  LoadingSpinner,
+  VisuallyHidden
+} from 'suomifi-ui-components';
+
+const [visible, setVisible] = useState(false);
+const [loaded, setLoaded] = useState(0);
+const [status, setStatus] = useState('loading');
+
+const timeout = () => {
```
LJKaski, line 54: Could this be named something a bit more descriptive? Not a big deal, but caught my eye.

src/core/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { Icon } from '../../core/Icon/Icon';
+
+export type LoadingSpinnerStatus = 'loading' | 'success' | 'failed';
+export interface LoadingSpinnerProps {
+  /** Custom class name for styling and customizing */
+  className?: string;
+  /** Unique id */
+  id?: string;
+  /** Status text */
+  text: string;
```
LJKaski, line 18: Should this be named `statusText` since that's what it essentially is? Or does it clash with the `statusText` props used elsewhere for a slightly different purpose? In any case it might make sense to make the prop name and/or description a bit more informative.

src/core/LoadingSpinner/LoadingSpinner.tsx
```diff
@@ -0,0 +1,126 @@
...
+  /** Align text on bottom or on the right side of animation svg
+   * @default 'bottom'
+   */
+  textAlign?: 'bottom' | 'right';
+  /** Is the text visible or hidden
```
LJKaski, line 23: This wording differs from the descriptions in rest of the library somewhat. Maybe "Show or hide status text" or something along those lines could work better.

src/core/LoadingSpinner/LoadingSpinner.tsx
ketsappi: We had the discussion about this yesterday and the previous label was not suitable as this text is changing based on the status text.
But as you noted the `statusText` as a prop name sounded confusing as it behaves differently than other components.

src/core/LoadingSpinner/LoadingSpinner.md
Riippi: No. I will remove it.

src/core/LoadingSpinner/LoadingSpinner.md
Riippi: Yes it could, but I have no good ideas. startTimer ?

LJKaski: Yeah, it's not the easiest to name... Maybe `runLoader` or something along those lines?

src/core/LoadingSpinner/LoadingSpinner.tsx
Riippi: It used to be LabelText.  That was confusing so in a daily meeting we decided to change it to "text". statusText was also in the table, but somehow we ended up with "text".

LJKaski: Makes sense. I think text conveys the purpose well enough :)

src/core/LoadingSpinner/LoadingSpinner.md
Riippi: That sounds good. Changed it now.

src/core/LoadingSpinner/LoadingSpinner.tsx
Riippi: Fixed

riitasointi (approved):
Good work!

PR #611, opened 2022-05-03
[Fix] Link styles specificity

riitasointi:
## Description

This PR moves the styles of `<Link>` component inside `.fi-link` class to avoid same specificity issues with `<a>` element reset styles. 

Also fixed issue with a redundant file `src/core/Link/Link/Link.baseStyles.ts` which now exports correctly named styles. 

## Motivation and Context

Suomi-fi library stylings should be predictable and robust. Component styles should be more important than HTML reset styles. 

## How Has This Been Tested?

On MacOS, Styleguidist, Chrome, Safari & Firefox

### Link
* Fix CSS specificity issue which caused unpredictable component styling

### DIFF
--- a/src/core/Link/BaseLink/BaseLink.baseStyles.ts
+++ b/src/core/Link/BaseLink/BaseLink.baseStyles.ts
@@ -4,26 +4,28 @@ import { element, font } from '../../theme/reset';
 import { allStates } from '../../../utils/css';
 
 export const baseStyles = (theme: SuomifiTheme) => css`
-  ${element(theme)}
   ${font(theme)('bodyText')}
-  ${allStates(`color: ${theme.colors.highlightBase};`)}
-  color: ${theme.colors.highlightBase};
-  text-decoration: none;
 
-  &:focus,
-  &:focus-within {
+  &.fi-link {
+    ${allStates(`color: ${theme.colors.highlightBase};`)}
+    color: ${theme.colors.highlightBase};
     text-decoration: none;
-  }
 
-  &:focus {
-    ${theme.focus.boxShadowFocus}
-  }
+    &:focus,
+    &:focus-within {
+      text-decoration: none;
+    }
 
-  &:hover,
-  &:active {
-    text-decoration: underline;
-  }
-  &:visited {
-    color: ${theme.colors.accentTertiaryDark1};
+    &:focus {
+      ${theme.focus.boxShadowFocus}
+    }
+
+    &:hover,
+    &:active {
+      text-decoration: underline;
+    }
+    &:visited {
+      color: ${theme.colors.accentTertiaryDark1};
+    }
   }
 `;
--- a/src/core/Link/BaseLink/BaseLink.baseStyles.ts
+++ b/src/core/Link/BaseLink/BaseLink.baseStyles.ts
@@ -1,6 +1,6 @@
 import { css } from 'styled-components';
 import { SuomifiTheme } from '../../theme';
-import { element, font } from '../../theme/reset';
+import { font } from '../../theme/reset';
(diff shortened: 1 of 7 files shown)


src/core/Link/BaseLink/BaseLink.baseStyles.ts
```diff
@@ -1,29 +1,31 @@
 import { css } from 'styled-components';
 import { SuomifiTheme } from '../../theme';
-import { element, font } from '../../theme/reset';
```
riitasointi, line 3: The `element` function seems to only add `color: ${theme.colors.blackBase};`? I think it could be removed in this case

src/core/Link/Link/Link.baseStyles.ts
```diff
@@ -2,6 +2,6 @@ import { SuomifiTheme } from '../../theme';
 import { css } from 'styled-components';
 import { baseStyles } from '../BaseLink/BaseLink.baseStyles';
 
-export const ExternalLinkStyles = (theme: SuomifiTheme) => css`
+export const LinkStyles = (theme: SuomifiTheme) => css`
   ${baseStyles(theme)}
```
riitasointi, line 6: Using the same pattern from `StaticChip`. This file only passes forward the `BaseLink` styles because nothing else is needed. Another possibility would be to remove this file altogether and use `/BaseLink/BaseLink.baseStyles` directly in `Link.tsx`

src/core/Link/Link/Link.tsx
```diff
@@ -22,7 +22,7 @@ const StyledLink = styled(
     />
   ),
 )`
-  ${({ theme }) => baseStyles(theme)}
+  ${({ theme }) => LinkStyles(theme)}
```
LJKaski, line 25: I actually find this naming quite informative compared to the bare `baseStyles` used throughout the library for all components. A more uniform naming would be to use baseStyles for now, but I could be discussed if we should change these names to be more informative in general.

src/core/Link/Link/Link.baseStyles.ts
LJKaski: I think this approach is pretty clean 👍

ketsappi: I'm okay with either way.

If following the existing, separate `*.baseStyles.ts` it might help as every component are (or should) be with same pattern.
But it is also little redundant to just keep the `Link.baseStyles.ts` hanging with no actual value given and just to be "for future" --> we can always add it back if we really need some customisation where the baseStyles is needed.

src/core/Link/Link/Link.tsx
ketsappi: That is true. The current `baseStyles` gives thoughts that it is some base that should be appended with more styles into it.
I vote for for the naming changes all around the components to get this kind of more describing names. 👍

src/core/Link/BaseLink/BaseLink.baseStyles.ts
LJKaski: I think so too. I didn't see any adverse effects from it, since the color seems to be defined by other means in all cases.

LJKaski (approved):
Seems to be working as inteded. Beta testing will tell if it solves the issue in the services.

src/core/Link/Link/Link.baseStyles.ts
riitasointi: No clear consensus against my choice, so I'll leave it :)

ketsappi (approved):

src/core/Link/Link/Link.tsx
ketsappi: Other components will be changed in later PR.

PR #614, opened 2022-05-05
[Fix] SingleSelect controlled state bug

riitasointi:
## Description

This PR fixes a bug in SingleSelect. The bug was as thus: When the component was in controlled state, and its `selectedItem` was changed from a defined value to `null`, the previous value was still present in the text input field. This was due to `filterInputValue` not being handled correctly in `getDerivedStateFromProps` hook. 

## Motivation and Context

Components should work predictably and without bugs in all use cases.

## How Has This Been Tested?

On MacOS: Chrome, Safari, Firefox. Styleguidist. 

### SingleSelect
* Fix an issue regarding controlled state nullification
* Add `onClearSelection()` prop

### DIFF
--- a/src/core/Form/Select/SingleSelect/SingleSelect.tsx
+++ b/src/core/Form/Select/SingleSelect/SingleSelect.tsx
@@ -86,6 +86,8 @@ export interface SingleSelectProps<T extends SingleSelectData> {
   onItemSelect?: (uniqueItemId: string | null) => void;
   /** Disable the input */
   disabled?: boolean;
+  /** Event sent when the button which clears the input is clicked */
+  onClearSelection?: () => void;
 }
 
 interface SingleSelectState<T extends SingleSelectData> {
@@ -357,6 +359,7 @@ class BaseSingleSelect<T> extends Component<
       ariaOptionsAvailableText,
       onItemSelect,
       disabled,
+      onClearSelection,
       ...passProps
     } = this.props;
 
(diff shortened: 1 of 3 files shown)


src/core/Form/Select/SingleSelect/SingleSelect.tsx
```diff
@@ -146,7 +146,7 @@ class BaseSingleSelect<T> extends Component<
       const resolvedSelectedItem =
         'selectedItem' in nextProps ? selectedItem : prevState.selectedItem;
       const resolvedInputValue = selectedItemChanged
-        ? resolvedSelectedItem?.labelText
+        ? resolvedSelectedItem?.labelText || ''
```
LJKaski, line 149: Just a thought: should we add a test for this?

riitasointi: Added!

riitasointi: Added!

src/core/Form/Select/SingleSelect/SingleSelect.tsx
```diff
@@ -81,11 +81,13 @@ export interface SingleSelectProps<T extends SingleSelectData> {
   /** Status text to be shown below the component and hint text. Use e.g. for validation error */
   statusText?: string;
   /** Controlled items; if item is in array, it is selected. If item has disabled: true, it will be disabled. */
-  selectedItem?: T & SingleSelectData;
+  selectedItem?: (T & SingleSelectData) | null;
```
riitasointi, line 84: Had to add `null` here to be able to write the test I wanted

ketsappi (approved):
Seems to work as we want. Good job!

PR #616, opened 2022-05-05
[Fix] Checkbox label alignment to vertical middle

Riippi:
## Description

Label of the checkbox is moved down by 1 pixel. Now it looks more closer to the center.

## Motivation and Context
It looked like the label was not in the vertical center.  

## How Has This Been Tested?
Chrome, Firefox and Safari.

Screenshot 2022-05-05 at 13 13 14

## Release notes
- Fix issue with Checkbox label alignment


Riippi:
Adjusted line heights to 30px and 27px. Looks like they are on the middle now.

ketsappi (dismissed):
Seems to work nicely. Good job!

_It seems that only the large variant had the change or at least I can see visual change only with large variant._

src/core/Form/Checkbox/Checkbox.baseStyles.tsx
```diff
@@ -57,6 +57,7 @@ const largeVariantStyles = (theme: SuomifiTheme) => css`
     & .fi-checkbox_label {
       padding-left: ${theme.spacing.xxl};
       min-height: 30px;
```
LJKaski, line 59: Does the min-height style serve a purpose now that the line-height is set to 30 as well?

LJKaski: Same goes for the smaller variant.

Riippi: Seems that there is no effect with min-height or without. So I removed it now.

Riippi:
Updated the screenshot. Not sure if it has any differences, but it's now from the current PR state.

LJKaski (approved):

PR #619, opened 2022-05-11
[Feature] Add new team members as code owners

ketsappi:
## Description
Add new team members as code owners. **Waiting first for new team members to be in correct group before merging.**


riitasointi (approved):
Let's do it!

PR #620, opened 2022-05-16
[Fix] Update engines field in package.json

ketsappi:
## Description

Updated to minimum version that is required to build the project and run the styleguidist.
Should have no effect on services using the library.

## Motivation and Context

`engines` field have not been updated since start.

When using older than v14 of the **node** , building the project was no longer possible as the dependencies required newer version of the node to be installed.

## How Has This Been Tested?
Building the project locally.

## Release notes

_I think there is no need to mention in release notes as this should be internal change._

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   ],
   "sideEffects": false,
   "engines": {
-    "node": ">=8.0.0"
+    "node": ">=14.0.0"
   },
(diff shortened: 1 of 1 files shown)


riitasointi (approved):
Good little update!

riitasointi (approved):

PR #622, opened 2022-05-17
[Fix] ExternalLink new window props

LJKaski:
## Description
This PR addresses an issue where ExternalLink would have a required prop `labelNewWindow` even when `toNewWindow` prop was set to false. The solution utilizes conditional props to change the logic as follows:
  * if `toNewWindow` is true, which is also the default value, `labelNewWindow` must be set as well
  * if `toNewWindow` is false, `labelNewWindow` cannot be set.

## Motivation and Context
A redundant prop was required when `toNewWindow` was set to false

## How Has This Been Tested?
Tested in styleguidist on Chrome as well as in a CRA project on Chrome. Also checked that VSCode tooltips show and support the new logic.

### ExternalLink
  * **Breaking change:** `labelNewWindow` no longer required or allowed when `toNewWindow` is set to false

### DIFF
--- a/src/core/Link/ExternalLink/ExternalLink.test.tsx
+++ b/src/core/Link/ExternalLink/ExternalLink.test.tsx
@@ -6,7 +6,6 @@ import { ExternalLink } from './ExternalLink';
 
 const TestExternalLink = (
   <ExternalLink
-    toNewWindow
     href="/"
     data-testid="test-link"
     labelNewWindow="Opens in a new window"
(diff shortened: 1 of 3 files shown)


ketsappi (dismissed):
Tested locally in CRA-TS and seems to work nicely as it complains when it should be.
Good job! 👍

riitasointi (approved):
Works as intended! Nice 👍

ketsappi (dismissed):

ketsappi (approved):

PR #623, opened 2022-05-20
[Fix] API documentation styling

LJKaski:
## Description
This PR improves the readability of the API descriptions in styleguidist. Previously there was an unnecessary fade at the bottom of the prop descriptions. This PR removes that and adds lines between props as well as vertically aligns content in rows.

## Motivation and Context
API descriptions should be clear and easily readable and skimmable.

## How Has This Been Tested?
By running styleguidist on Chrome on Windows

image

### General changes
* Improve API documentation readability in Styleguidist

### DIFF
--- a/.styleguidist/styleguidist.styles.js
+++ b/.styleguidist/styleguidist.styles.js
@@ -71,7 +71,7 @@ module.exports = {
     type: {
       display: 'block',
       overflowY: 'auto',
-      maxHeight: '6em',
+      maxHeight: '144px',
       minWidth: '20vw',
     },
   },
--- a/.styleguidist/styleguidist.styles.js
+++ b/.styleguidist/styleguidist.styles.js
@@ -78,17 +78,8 @@ module.exports = {
   Table: {
     cell: {
       position: 'relative',
-      '&:after': {
-        display: 'block',
-        content: '""',
-        position: 'absolute',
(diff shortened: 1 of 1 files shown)


.styleguidist/styleguidist.styles.js
```diff
@@ -71,24 +71,15 @@ module.exports = {
...
-        content: '""',
-        position: 'absolute',
-        top: '0',
-        height: '100%',
-        width: '100%',
-        background:
-          'linear-gradient(180deg, rgba(255, 255, 255, 0) 0, rgba(255, 255, 255, 0) 4.5em, rgba(255, 255, 255, 1) 6em)',
-        pointerEvents: 'none',
-      },
+      borderBottom: '1px solid #e8e8e8',
```
LJKaski, line 81: This is the same color that seemed to be in use in other borders on the page.

riitasointi:
Idea: Could we also modify the `<code>` element styles a bit so it pops out from the text more?

Here is the current style:
image

And maybe we could have something like this:
image

ketsappi (dismissed):
Improves the experience with the styleguidist by good amount.
No more faded out description for e.g MultiSelect's interface.

Good job!

ketsappi (approved):

PR #624, opened 2022-05-23
[Feature] CheckboxGroup status & status text

ketsappi:
## Description

Add `groupStatus` and `groupStatusText` props.
`groupStatus` will affect the children Checkboxes by passing the status to them.

If individual Checkbox has given `status` already it will be overriding the group's `groupStatus`.

## Motivation and Context

Able to give status for every Checkbox that are children of the group.

## How Has This Been Tested?
So far locally in styleguidist

## Screenshots:
```
// Props given
groupStatus="error"
groupStatusText="Please choose at least one"
```

image

### CheckboxGroup (#624)
- Add `groupStatus` and `groupStatusText` props.

### DIFF
--- a/src/core/Form/Checkbox/Checkbox.tsx
+++ b/src/core/Form/Checkbox/Checkbox.tsx
@@ -253,19 +253,19 @@ const StyledCheckbox = styled(
 
 export const Checkbox = forwardRef(
   (props: CheckboxProps, ref: React.RefObject<HTMLInputElement>) => {
-    const { id: propId, ...passProps } = props;
+    const { id: propId, status: propStatus, ...passProps } = props;
     return (
       <SuomifiThemeConsumer>
         {({ suomifiTheme }) => (
           <AutoId id={propId}>
             {(id) => (
               <CheckboxGroupConsumer>
-                {({ status }) => (
+                {({ status: groupStatus }) => (
                   <StyledCheckbox
                     theme={suomifiTheme}
                     id={id}
                     forwardedRef={ref}
-                    status={status}
+                    status={!!propStatus ? propStatus : groupStatus}
                     {...passProps}
                   />
                 )}
--- a/src/core/Form/Checkbox/Checkbox.tsx
+++ b/src/core/Form/Checkbox/Checkbox.tsx
@@ -10,6 +10,7 @@ import { HtmlLabel, HtmlDiv, HtmlInput } from '../../../reset';
 import { Icon } from '../../Icon/Icon';
 import { StatusText } from '../StatusText/StatusText';
 import { HintText } from '../HintText/HintText';
+import { CheckboxGroupConsumer } from './CheckboxGroup';
 import { baseStyles } from './Checkbox.baseStyles';
 
 const baseClassName = 'fi-checkbox';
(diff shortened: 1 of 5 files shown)


src/core/Form/Checkbox/Checkbox.tsx
```diff
@@ -252,18 +253,23 @@ const StyledCheckbox = styled(
...
+                {({ status: groupStatus }) => (
+                  <StyledCheckbox
+                    theme={suomifiTheme}
+                    id={id}
+                    forwardedRef={ref}
+                    status={!!propStatus ? propStatus : groupStatus}
+                    {...passProps}
+                  />
+                )}
+              </CheckboxGroupConsumer>
```
riitasointi, line 272: Makes sense to use the React context pattern here I think. Same thing seems to be used in other places as well, like ExpanderGroup and RadioButtonGroup.

src/core/Form/Checkbox/CheckboxGroup.tsx
```diff
@@ -76,9 +102,24 @@ class BaseCheckboxGroup extends Component<
...
         </HtmlFieldSet>
+        <StatusText
+          className={classnames({
+            [checkboxGroupClassNames.statusTextHasContent]: !!groupStatusText,
+          })}
+          id={statusTextId}
+          status={groupStatus}
+        >
+          {groupStatusText}
+        </StatusText>
```
riitasointi, line 122: This more a comment on the visual design: The group status text could be a bit difficult to connect to the right context when it's just under the last checkbox. At least I could confuse it to only be connected to the last box, not the entire group

ketsappi: As discussed elsewhere, will leave the decision to be made by the user of the library.

In normal cases when using Checkboxes in a group, the `groupStatusText` will be the only one used.

src/core/Form/Checkbox/CheckboxGroup.tsx
```diff
@@ -74,11 +102,31 @@ class BaseCheckboxGroup extends Component<
               {labelText}
             </Label>
             {groupHintText && <HintText>{groupHintText}</HintText>}
+            {groupStatusText && (
+              <VisuallyHidden
+                {...getConditionalAriaProp('aria-labelledby', [statusTextId])}
```
LJKaski, line 107: Is there a reason this is defined as an aria attribute instead of just content in the visually hidden? Both seem to work the same, just wondering.

ketsappi: Tried to follow the pattern how **StatusText** was used elsewhere, so that it is referenced by it's **id**. 
So that the actual content is written only once, inside StatusText.

LJKaski: Yeah, makes sense. Thanks for clarifying. With this approach I think the component it's defined in could be almost anything, but `VisuallyHidden` is probably the safest bet.

ketsappi: True, it could be almost any element. I will leave it to use `VisuallyHidden` as it tells the purpose quite clearly.

riitasointi (approved):
MacOS Voiceover gives quite an "onslaught" of information when both statusText and hintText are used for a checkbox 😅  But at least all the info is there. 

Seems to work!

ketsappi:
Yeah, we can mention in the DS-site for the CheckboxGroup that library users should consider short texts only to prevent the "flood".

PR #625, opened 2022-05-23
[BreakingChange] Add ref suport for components

Riippi:
## Description
Add ref support for components that don't yet have it

## Related Issue

Closes https://jira.dvv.fi/browse/SFIDS-610

## Motivation and Context
Adding ref forwarding to components where it seems viable.
Having unified system of forwarding refs should make using the components easier and also reduces complexity of the codebase a bit.

- Ref forwarding with `ref` attribute is added to components where it seemed viable
- Also the `forwardedRef` prop is kept in or added to the api even though it's not actually needed. It works just like `ref` does.
  -  Because of the components inner structure the `forwardedRef` prop would be available to use even if it's not "visible". 
  - It would have needed additional complexity to prevent usage of the `forwardedRef`
  - React/Styleguidist don't show ref or it's type. With `forwardedRef` in the api we can show the ref type in Styleguidist documentation.
- Some of the component examples are updated to include usage of `ref`
- Ref is usually passed to the most "important" part of the component. For example button or input that can have focus. When there is none the ref is passed to the root element. This is decided case by case.

**Added forward ref support**
Block
StaticChip
Expander
ExpanderGroup
ExpanderTitleButton
CheckboxGroup
HintText
RadioButtonGroup
SearchInput
MultiSelect
SingleSelect
StatusText
Textarea
ToggleInput
LanguageMenu
ExternalLink
Link
SkipLink
Paragraph
Text
VisuallyHidden

**Components that previously had support for React ref now also have forwardedRef prop visible.**
Button
Heading
Tooltip
ToggleButton
LoadingSpinner
CheckBox
RadioButton
Chip
Toast
Dropdown

## How Has This Been Tested?

With CRA project, Styleguidist, unit tests

Example of testing Text component in the project. Add these to some other component:
```
// in constructor
    this.textRef = React.createRef();

// in template
        <Text
          color="depthBase"
          ref={this.textRef}
          onClick={() => console.log(this.textRef?.current)}
        >
          foo
        </Text>
```

## Release notes
- React ref forwarding support is added to multiple components

### DIFF
--- a/src/core/Form/Select/SingleSelect/SingleSelect.test.tsx
+++ b/src/core/Form/Select/SingleSelect/SingleSelect.test.tsx
@@ -462,3 +462,25 @@ describe('custom item addition mode', () => {
     }
   });
 });
+
+describe('forward ref', () => {
+  it('ref is forwarded to input', () => {
+    const ref = React.createRef<HTMLInputElement>();
+
+    render(
+      <SingleSelect
+        labelText="SingleSelect"
+        clearButtonLabel="Clear selection"
+        items={tools}
+        defaultSelectedItem={defaultSelectedTool}
+        noItemsText="No items"
+        visualPlaceholder="Select item"
+        ariaOptionsAvailableText="Options available"
+        ref={ref}
+      />,
+    );
+
+    expect(ref.current?.tagName).toBe('INPUT');
+    expect(ref.current?.value).toBe('Hammer');
+  });
+});
--- a/src/core/Text/Text.tsx
+++ b/src/core/Text/Text.tsx
@@ -29,58 +29,33 @@ export interface TextProps extends InternalTextProps {
   ref?: React.RefObject<HTMLSpanElement>;
 }
 
-class BaseText extends Component<TextProps & InnerRef> {
-  render() {
-    const {
-      id,
-      className,
-      children,
-      forwardedRef,
-      variant = 'body',
-      smallScreen,
-      color,
-
-      ...passProps
-    } = this.props;
-
-    return (
-      <HtmlSpan
-        {...passProps}
-        className={classnames(
-          baseClassName,
-          className,
-          [`${baseClassName}--${variant}`],
-          {
-            [smallScreenClassName]: smallScreen,
-          },
-        )}
-      />
-    );
-  }
-}
-
 const StyledText = styled(
   ({
     theme,
+    className,
+    variant,
+    smallScreen,
     ...passProps
   }: InternalTextProps & InnerRef & SuomifiThemeProp) => (
-    <BaseText {...passProps} />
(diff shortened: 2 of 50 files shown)


src/core/Text/Text.tsx
```diff
@@ -20,6 +20,15 @@ export interface TextProps extends HtmlSpanProps {
...
 
+interface InnerRef {
+  forwardedRef: React.RefObject<HTMLSpanElement>;
+}
+
+export interface TextProps extends InternalTextProps {
+  /** Ref object to be passed to the element */
+  ref?: React.RefObject<HTMLSpanElement>;
+}
+
```
riitasointi, line 31: I'm wondering about these interfaces. `<HintText>` does not have these and for all intents and purposes it seems to work the same way in the end: the ref gets passed all the way to the `<span>` element. Do we need these interfaces?

Also, even though `ref` is defined here as part of the `TextProps`, it's not visible on the Styleguidist page as part of the `<Text>` component's API. I think it would be nice to have it there so developers know it can be used. Maybe also documented more specifically like: "Ref object which gets passed to the underlying span-element"

Riippi: Outdated, but in the end when forwardRef was left to be visible in the api most of these inner interfaces were removed. 
Documenting the ref is now done with forwardedRef prop.

riitasointi (dismissed):
Tested some of the components in a CRA app, the refs seem to function as intended. Only component I was wondering about was ExpanderGroup where the ref is passed to the closeAll/openAll button instead of the wrapping div, but I guess that's a good choice still. 

Good work!

src/core/Form/Select/SingleSelect/SingleSelect.test.tsx
```diff
@@ -462,3 +462,25 @@ describe('custom item addition mode', () => {
...
+        items={tools}
+        defaultSelectedItem={defaultSelectedTool}
+        noItemsText="No items"
+        visualPlaceholder="Select item"
+        ariaOptionsAvailableText="Options available"
+        ref={ref}
+      />,
+    );
+
+    expect(ref.current?.tagName).toBe('INPUT');
```
LJKaski, line 483: Nitpicking department, but does the question mark serve a purpose here?

LJKaski (approved):
Looks nice and clean to me, and seems to work as intended. Great work! 👍

src/core/Form/Select/SingleSelect/SingleSelect.test.tsx
Riippi: It makes the linter happy.  Without it it's `Object is possibly 'null'`

riitasointi (approved):

PR #626, opened 2022-05-24
|Feature] Possibility to add an item outside of list to SingleSelect & MultiSelect

riitasointi:
## Description

This PR adds functionality to SingleSelect & MultiSelect components. Setting a boolean prop `allowItemAddition` as true, it is possible for the user to add their own custom items as the input's selected value ("vapaa asiasana"). 

In this mode, a new prop called `itemAdditionHelpText` is also required. The existing prop `noItemsText` is not required in this case. 

Under the hood, there is a new component called `<SelectItemAddition>` which is rendered below `<SelectItemList>` when appropriate. 

Added couple of new examples to Styleguidist: basic and controlled use. 

This PR also fixes a bug where keyboard navigation didn't go to the last element when up arrow was pressed when no item was previously focused. 

## Motivation and Context

We want to provide an option to add custom items to the select components.

## How Has This Been Tested?

In Styleguidist so far. Chrome, Safari, Firefox 

## Screenshots:

https://user-images.githubusercontent.com/17459942/170021417-76d88d6d-0625-4a75-a7f3-ce43687ed479.mov

### SingleSelect
* Add possibility to allow user to add their own custom item as the selected item
  - This mode requires a new prop `allowItemAddition` to be `true`. A string prop `itemAdditionHelpText` is then also required
* Fix a bug with keyboard navigation

### SingleSelect
* Add possibility to allow user to add their own custom items as selected items
  - This mode requires a new prop ´allowItemAddition´ to be `true`. A string prop `itemAdditionHelpText` is then also required
* Fix a bug with keyboard navigation

### DIFF
--- a/src/core/Form/Select/BaseSelect/SelectItemAddition/SelectItemAddition.tsx
+++ b/src/core/Form/Select/BaseSelect/SelectItemAddition/SelectItemAddition.tsx
@@ -8,7 +8,7 @@ import classnames from 'classnames';
 const baseClassName = 'fi-select-item-addition';
 
 const classNames = {
-  hint_text: `${baseClassName}_hint_text`,
+  hint_text: `${baseClassName}_hint-text`,
   item: `${baseClassName}_item`,
   hasKeyboardFocus: `${baseClassName}_item--hasKeyboardFocus`,
 };
--- a/src/core/Form/Select/BaseSelect/SelectItemAddition/SelectItemAddition.tsx
+++ b/src/core/Form/Select/BaseSelect/SelectItemAddition/SelectItemAddition.tsx
@@ -1,0 +1,85 @@
+import React, { Component } from 'react';
+import { default as styled } from 'styled-components';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../../../theme';
+import { HtmlDiv, HtmlLi } from '../../../../../reset';
+import { selectItemAdditionStyles } from './SelectItemAddition.baseStyles';
+import classnames from 'classnames';
+
+const baseClassName = 'fi-select-item-addition';
+
+const classNames = {
+  hint_text: `${baseClassName}_hint_text`,
+  item: `${baseClassName}_item`,
+  hasKeyboardFocus: `${baseClassName}_item--hasKeyboardFocus`,
+};
+
+export interface SelectItemAdditionProps {
+  /** Text displayed above the item. Will also be read by screen readers when focusing on the item itself */
+  hintText: string;
+  /** onClick event handler */
+  onClick: (event: React.MouseEvent<HTMLLIElement>) => void;
+  /** Indicates if the current item has keyboard focus. */
+  hasKeyboardFocus: boolean;
+  /** Unique id for the item */
+  id: string;
+  /** Text value for the item to be added */
+  children: string;
+}
+
+class BaseSelectItemAddition extends Component<
+  SelectItemAdditionProps & SuomifiThemeProp
+> {
+  render() {
+    const {
+      hintText,
+      onClick,
+      hasKeyboardFocus,
+      id,
+      theme,
+      children,
+      ...passProps
+    } = this.props;
+
+    return (
+      <HtmlDiv {...passProps}>
+        <HtmlDiv className={classNames.hint_text}>{hintText}</HtmlDiv>
+        <HtmlLi
+          className={classnames(classNames.item, {
+            [classNames.hasKeyboardFocus]: hasKeyboardFocus,
+          })}
+          tabIndex={-1}
+          onMouseDown={(event) => {
+            // prevent focusing the li element
+            event.preventDefault();
+          }}
+          onClick={(event) => {
+            if (!!onClick) {
+              onClick(event);
+            }
+          }}
+          aria-label={`${hintText}. ${children}.`}
+          role="option"
+          id={id}
+        >
+          {children}
+        </HtmlLi>
+      </HtmlDiv>
+    );
+  }
+}
+
+const StyledSelectItemAddition = styled(BaseSelectItemAddition)`
+  ${({ theme }) => selectItemAdditionStyles(theme)}
+`;
+
+export class SelectItemAddition extends Component<SelectItemAdditionProps> {
+  render() {
+    return (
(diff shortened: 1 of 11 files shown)


src/core/Form/Select/BaseSelect/SelectItemList/SelectItemList.tsx
```diff
@@ -23,9 +22,7 @@ export interface SelectItemListProps {
   /** SelectItemList container div class name for custom styling. */
   className?: string;
   /** List items */
-  children:
-    | Array<ReactElement<SelectItemProps>>
-    | ReactElement<SelectItemProps>;
+  children: ReactElement | Array<ReactElement>;
```
riitasointi, line 25: I simplified this a little bit since the different types of elements this can take as children is already quite long with the addition of <SelectItemAddition>. Also it was quite difficult to make the types work with conditional rendering logic. 

I say this should be fine since this interface is only for our internal use.

src/core/Form/Select/BaseSelect/SelectItemAddition/SelectItemAddition.tsx
```diff
@@ -0,0 +1,85 @@
...
+import { default as styled } from 'styled-components';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../../../theme';
+import { HtmlDiv, HtmlLi } from '../../../../../reset';
+import { selectItemAdditionStyles } from './SelectItemAddition.baseStyles';
+import classnames from 'classnames';
+
+const baseClassName = 'fi-select-item-addition';
+
+const classNames = {
+  hint_text: `${baseClassName}_hint_text`,
```
LJKaski, line 11: This naming seems slightly iffy. Is it in line of the general naming convention of these? It's the underscore between hint and text that seems a bit off. Would a hyphen or camelcase be more suitable?

src/core/Form/Select/MultiSelect/MultiSelect/__snapshots__/MultiSelect.test.tsx.snap
```diff
@@ -872,6 +872,7 @@ exports[`has matching snapshot 1`] = `
 
 <div
   class="c0 fi-multiselect c1"
+  items="[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object],[object Object]"
```
LJKaski, line 875: This seems to be leaking into the DOM due to not being destructured from props in MultiSelect after the changes.

src/core/Form/Select/SingleSelect/SingleSelect.tsx
```diff
@@ -335,13 +427,13 @@ class BaseSingleSelect<T> extends Component<
...
       filterInputValue,
       selectedItem,
+      computedItems,
     } = this.state;
 
     const {
       id,
       className,
       theme,
-      items: propItems,
```
LJKaski, line 344: Removing this leads to the item list being leaked into the DOM.

src/core/Form/Select/BaseSelect/SelectItemAddition/SelectItemAddition.tsx
riitasointi: True, should use hyphen as separator there. 

Fixed!

src/core/Form/Select/SingleSelect/SingleSelect.tsx
riitasointi: Good catch. Added it back in, although `items` is no longer used in rendering because there is now `computedItems` state.

src/core/Form/Select/MultiSelect/MultiSelect/__snapshots__/MultiSelect.test.tsx.snap
riitasointi: See comment below

src/core/Form/Select/MultiSelect/MultiSelect/MultiSelect.tsx
```diff
@@ -61,7 +62,7 @@ export interface MultiSelectProps<T extends MultiSelectData> {
   optionalText?: string;
   /** Hint text to be shown below the label */
   hintText?: string;
-  /** Event that is fired when item selections change. Not fired when in controlled state. */
+  /** Event that is fired when item selections change */
   onItemSelectionsChange?: (selectedItems: Array<T>) => void;
```
LJKaski, line 66: Is there a reason this array is only of type T and not T & MultiSelectData like most other selected items related arrays? Not changed in this PR but just hit my eye.

riitasointi: Yes, I think it should be the same type as used everywhere else with items. 

Changed it for both SingleSelect & MultiSelect

src/core/Form/Select/SingleSelect/SingleSelect.tsx
```diff
@@ -164,6 +187,14 @@ class BaseSingleSelect<T> extends Component<
...
   }
 
+  componentDidUpdate(prevProps: SingleSelectProps<T & SingleSelectData>): void {
+    if (JSON.stringify(this.props.items) !== JSON.stringify(prevProps.items)) {
+      this.setState({
+        computedItems: this.props.items,
+      });
+    }
+  }
+
```
riitasointi, line 197: I know comparing object equality with stringify is maybe not the most elegant solution. But it works 🤷‍♂️

ketsappi: I have a hunch that this might be problematic if we use `JSON.stringify` performance wise.

We might have a some similar case in existing components where the prop change is wanted to detect. Will try to find it.

ketsappi: I searched around and couldn't also found any more elegant solution that fits the case at hand. 
I will suggest that we put it to beta and and test it more. Keeping eyes open if it causes any issues while testing more down the line.

ketsappi (dismissed):
Seems to work as intended and this wanted feature is ready to go to the beta stage!

Good job! 💪

Riippi (approved):
LGTM

PR #634, opened 2022-06-13
[Feature] Add polymorphic RouterLink component

riitasointi:
## Description

This PR adds a new component called `<RouterLink`>. The solution is based on the work started in the branch https://github.com/vrk-kpa/suomifi-ui-components/tree/fix/as-prop

As the name suggests, the component is intended to be used with external (routing) libraries. It takes a prop called `asComponent` and renders itself as that component while applying the correct Suomi.fi styles on top of it. 

Features:
- Full Typescript support for props based on the component passed to `asComponent`
- Full theme customisation support through `SuomifiThemeProvider`

Usage example:

```
import { Link as ReactRouterLink } from "react-router-dom";
import { RouterLink } from "suomifi-ui-components";

<RouterLink asComponent={ReactRouterLink} to="/">
  Home
</RouterLink>
```

## Motivation and Context

We want to have a polymorphic link component like this so it can easily be used with all kinds of frameworks while retaining the correct Suomi.fi styles and responding to overrides of the theme (via `SuomifiThemeProvider`). 

This component could also be used in navigation/menu components in this library the future. 

## How Has This Been Tested?

In Gatsby (as GatsbyLink) and in CRA as ReactRouterLink. Various Typescript and rendering scenarios. 

Caveat: This component is pretty much redundant for NextJS projects. Next's Link component is a high order component and does not allow CSS classes to be passed to it which makes the logic of this component dead on arrival. The correct behavior in NextJS projects can be achieved using the basic Link component from suomifi-ui-components like this

```
import Link from "next/link";
import { Link as SuomiFiLink } from "suomifi-ui-components";

<Link href="/someroute" passHref>
  <SuomiFiLink href="/someroute">Suomi.fi styled link with NextJS link behavior</SuomiFiLink>
</Link>
```

### RouterLink
* Introduce polymorphic `<RouterLink>` component
   - Renders as the component of your choice, for example React Router Link but with proper Suomi.fi link styles
   - Full Typescript support for props based on the component passed to `asComponent`
   - Full theme customisation support through `SuomifiThemeProvider`

### DIFF
--- a/src/core/Link/RouterLink/RouterLink.baseStyles.ts
+++ b/src/core/Link/RouterLink/RouterLink.baseStyles.ts
@@ -1,0 +1,31 @@
+import { css } from 'styled-components';
+import { SuomifiTheme } from '../../theme';
+import { font } from '../../theme/reset';
+import { allStates } from '../../../utils/css';
+
+export const RouterLinkStyles = (theme: SuomifiTheme) => css`
+  ${font(theme)('bodyText')}
+
+  &.fi-link--router {
+    ${allStates(`color: ${theme.colors.highlightBase};`)}
+    color: ${theme.colors.highlightBase};
+    text-decoration: none;
+
+    &:focus,
+    &:focus-within {
+      text-decoration: none;
+    }
+
+    &:focus {
+      ${theme.focus.boxShadowFocus}
+    }
+
+    &:hover,
+    &:active {
+      text-decoration: underline;
+    }
+    &:visited {
+      color: ${theme.colors.accentTertiaryDark1};
+    }
+  }
+`;
--- a/src/core/Link/RouterLink/RouterLink.tsx
+++ b/src/core/Link/RouterLink/RouterLink.tsx
@@ -1,0 +1,101 @@
+// import React, { Component, ReactNode } from 'react';
+import React, { ReactNode } from 'react';
+import { default as styled } from 'styled-components';
+import { RouterLinkStyles } from './RouterLink.baseStyles';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
+import { baseClassName } from '../BaseLink/BaseLink';
+import classnames from 'classnames';
+import { HtmlA } from '../../../reset';
+
+//
+// Source: https://www.benmvp.com/blog/polymorphic-react-components-typescript/
+//
+// Source: https://github.com/emotion-js/emotion/blob/master/packages/styled-base/types/helper.d.ts
+// A more precise version of just React.ComponentPropsWithoutRef on its own
+export type PropsOf<
+  C extends keyof JSX.IntrinsicElements | React.JSXElementConstructor<any>,
+> = JSX.LibraryManagedAttributes<C, React.ComponentPropsWithoutRef<C>>;
+
+type AsProp<C extends React.ElementType> = {
+  /**
+   * An override of the default HTML tag.
+   * Can also be another React component.
+   * @default <a>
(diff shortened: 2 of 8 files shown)


src/core/Link/RouterLink/RouterLink.baseStyles.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    &:hover,
+    &:active {
+      text-decoration: underline;
+    }
+    &:visited {
+      color: ${theme.colors.accentTertiaryDark1};
+    }
+  }
+`;
```
riitasointi, line 31: This is basically a copy-paste from BaseLink styles with just the class name changed. I think it's a good idea to have a separate class for this component so it can be overridden separately from .fi-link if needed.

src/core/Link/RouterLink/RouterLink.tsx
```diff
@@ -0,0 +1,101 @@
...
+import React, { ReactNode } from 'react';
+import { default as styled } from 'styled-components';
+import { RouterLinkStyles } from './RouterLink.baseStyles';
+import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
+import { baseClassName } from '../BaseLink/BaseLink';
+import classnames from 'classnames';
+import { HtmlA } from '../../../reset';
+
+//
+// Source: https://www.benmvp.com/blog/polymorphic-react-components-typescript/
```
riitasointi, line 11: Everything in this file until line 72 is copied directly from https://github.com/vrk-kpa/suomifi-ui-components/tree/fix/as-prop branch (work done by @ketsappi)

src/core/Link/RouterLink/RouterLink.tsx
```diff
@@ -0,0 +1,101 @@
...
+  C extends keyof JSX.IntrinsicElements | React.JSXElementConstructor<any>,
+> = JSX.LibraryManagedAttributes<C, React.ComponentPropsWithoutRef<C>>;
+
+type AsProp<C extends React.ElementType> = {
+  /**
+   * An override of the default HTML tag.
+   * Can also be another React component.
+   * @default <a>
+   */
+  asComponent?: C;
```
riitasointi, line 25: This prop was intended to be named just `as` but I found out it messes up styled-components somehow. When I renamed it to something else like this it started to work perfectly.

src/core/Link/RouterLink/RouterLink.tsx
```diff
@@ -0,0 +1,101 @@
...
+      {children}
+    </Component>
+  );
+};
+
+const StyledRouterLink = styled(PolymorphicLink)`
+  ${({ theme }) => RouterLinkStyles(theme)}
+`;
+
+export const RouterLink = <C extends React.ElementType = 'a'>(
```
riitasointi, line 95: Is this a good name for the component? It describes the intended use case but could it just be `PolymorphicLink`? Then we could maybe remove the support for `asProp` from the basic Link component

ketsappi: Hmm.. maybe we could have it as `RouterLink` that is exported out of the library.

We can then possibly use the `PolymorphicLink` internally like you said.

ketsappi (dismissed):
Tested with CRA-TS testproject.
And with a freshly made Gatsby site using the Gatsby's local link, `Link`.
```tsx
import { Link } from "gatsby";
```

Works nicely. Good job Ilkka! 💯

ketsappi (approved):

PR #635, opened 2022-06-13
[Fix] Tooltip arrow borders

riitasointi:
## Description

This PR fixes an issue with Tooltip's arrow borders.

## Motivation and Context

We want our components to have a professional and polished visual look.

## How Has This Been Tested?

Styleguidist on MacOS Chrome, Safari & Firefox

## Screenshots
Before:
Screenshot 2022-06-09 at 14 59 16

After:
image

### Tooltip
* Fix arrow borders visual glitch

### DIFF
--- a/src/core/Tooltip/TooltipContent/TooltipContent.baseStyles.tsx
+++ b/src/core/Tooltip/TooltipContent/TooltipContent.baseStyles.tsx
@@ -25,7 +25,7 @@ export const baseStyles = (arrowOffsetPx: number, theme: SuomifiTheme) => css`
       pointer-events: none;
     }
     &:before {
-      border-bottom-color: ${theme.colors.depthDark2};
+      border-bottom-color: ${theme.colors.blackBase};
       border-width: 9px;
       margin-right: -9px;
       bottom: 100%;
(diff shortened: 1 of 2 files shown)


src/core/Tooltip/TooltipContent/TooltipContent.baseStyles.tsx
```diff
@@ -20,20 +20,21 @@ export const baseStyles = (arrowOffsetPx: number, theme: SuomifiTheme) => css`
       position: absolute;
       height: 0;
       width: 0;
-      bottom: 100%;
       left: ${arrowOffsetPx}px;
       border: solid transparent;
       pointer-events: none;
     }
     &:before {
       border-bottom-color: ${theme.colors.depthDark2};
```
Riippi, line 28: Would it make sense to use a bit darker color here? The line color looks lighter when it's in an angle, compared to the box borders.  For example depthDark1 is a bit darker.

Riippi: Looks better now. On macbook screen you can notice that it's darker than other borders. Then in the external HP screen it's a bit lighter than others. I suppose in average it's good 😄

Riippi (dismissed):
LGTM

src/core/Tooltip/TooltipContent/TooltipContent.baseStyles.tsx
ketsappi: This seems interesting problem; when using same colour, `depthDark2` it actually looks a lot lighter as mentioned. But when using the `blackBase` it looks nice until zooming closer.

So as a compromise I would vote also for `blackBase` for it to look visually pleasant in normal use. We can change it later if this is a real issue or we can figure some other solution for it.

ketsappi (dismissed):

ketsappi (approved):

PR #636, opened 2022-06-13
[Fix] Improve typings for MouseEvent and KeyboardEvent

Riippi:
## Description

Button onClick function was showing the `MouseEvent` type as a "native" version instead of the React version. Actually the React version was used. This was already fixed in the tooltip PR #596 
Now also the typings in the code are fixed. 

Also the type `KeyboardEvent` is removed from places where it's not actually used.

## Motivation and Context

Removes unnecessary code and makes it a bit easier to read and use.

## How Has This Been Tested?
Typescript CRA app

### Button
- **Breaking Changes:** 
  - `onClick` event is now React `MouseEvent` instead of native `MouseEvent`
  - `KeyboardEvent` removed from `onClick` event type alternative

### DIFF
--- a/src/core/Button/Button.tsx
+++ b/src/core/Button/Button.tsx
@@ -54,7 +54,7 @@ interface InternalButtonProps
   /** Event handler to execute when clicked
    *  @default void
    */
-  onClick?: (event: MouseEvent | KeyboardEvent) => void;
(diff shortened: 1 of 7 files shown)


ketsappi (dismissed):
LGTM! 👍

riitasointi (approved):
These changes make things clearer. Good job!

PR #639, opened 2022-06-27
[Hotfix] Keyboard event propagation

riitasointi:
## Description

This PR addresses an issue regarding keyboard events in certain component. The propagation of these events was not addressed which caused unexpected behavior in certain cases. For example, when SingleSelect was used inside a Modal and the escape key was used to close the popover items list, the event was propagated all the way to the Modal component and the Modal was closed. 

## Motivation and Context

We don't want to have any unexpected side effects when using the keyboard to interact with our components

## How Has This Been Tested?

In a CRA app:
 - When used inside a `<Modal>`, make sure escape key only closes the popover, the modal: SingleSelect, MultiSelect, Dropdown, LanguageMenu
 - When used inside a `<form>`, make sure enter keypress does not submit the form: SingleSelect, MultiSelect

### SingleSelect & MultiSelect
* Prevent onKeyDown event from propagating upwards
* Prevent event default when pressing Enter key

### Dropdown & LanguageMenu
* Prevent event propagation when Escape key is pressed

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "suomifi-ui-components",
-  "version": "8.0.0",
+  "version": "8.0.1",
   "description": "Suomi.fi UI component library",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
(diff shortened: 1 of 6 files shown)


ketsappi (approved):
Tested to have Dropdown, LanguageMenu, SingleSelect and MultiSelect on Modal. 
If these components have their menus open, it now stops the key event from bubbling up.

Good work! 👍

PR #643, opened 2022-06-30
[Feature] Add displayNames to components

riitasointi:
## Description

This PR adds the `displayName` property to all exported components of the library.
Also convert all outer components to function components (in favor of class components) for a more unified codebase.

## Motivation and Context

This came up when creating a demo for a refurbished DS-site (https://github.com/riitasointi/ds-site-poc). When crafting component showcases, it became apparent the displayNames are missing from our components. 

## How Has This Been Tested?

In a CRA app and the DS-site demo NextJS project. 

## Release notes

* Add the `displayName` property to all components

### DIFF
--- a/src/core/Expander/Expander/Expander.tsx
+++ b/src/core/Expander/Expander/Expander.tsx
@@ -216,4 +216,5 @@ const Expander = (props: ExpanderProps) => {
   );
 };
 
+Expander.displayName = 'Expander';
 export { Expander, ExpanderConsumer, ExpanderProvider };
--- a/src/core/Expander/Expander/Expander.tsx
+++ b/src/core/Expander/Expander/Expander.tsx
@@ -192,30 +192,28 @@ interface ExpanderState {
  * <i class="semantics" />
  * Hide or show content with always visible title
  */
-export class Expander extends Component<ExpanderProps> {
-  render() {
-    const { id: propId, ...passProps } = this.props;
-    return (
-      <SuomifiThemeConsumer>
-        {({ suomifiTheme }) => (
-          <AutoId id={propId}>
-            {(id) => (
-              <ExpanderGroupConsumer>
-                {(consumer) => (
-                  <StyledExpander
-                    theme={suomifiTheme}
-                    id={id}
-                    {...passProps}
-                    consumer={consumer}
-                  />
-                )}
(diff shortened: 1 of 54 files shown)


src/core/Icon/Icon.tsx
```diff
@@ -60,20 +44,7 @@ const StyledSuomifiIcon = styled(
...
-export class Icon extends Component<IconProps> {
-  render() {
-    const { icon, ...passProps } = this.props;
-    const { className, ariaLabel } = this.props;
-
-    if (icon !== undefined) {
-      return <StyledSuomifiIcon {...passProps} icon={icon} />;
-    }
-
-    iconLogger(ariaLabel, className);
```
riitasointi, line 75: For the `<Icon>` component, I simplified the outer layer a lot. Since the `icon` prop is mandatory, it really makes no sense to have this if-condition there and have it return undefined if it's not provided. Same goes for `<StaticIcon>` and `<LogoIcon>`.

BTW, tested this without TS and if the icon prop is left out it renders as the login-icon

src/core/Expander/Expander/Expander.tsx
```diff
@@ -192,30 +192,28 @@ interface ExpanderState {
...
+                  consumer={consumer}
+                />
+              )}
+            </ExpanderGroupConsumer>
+          )}
+        </AutoId>
+      )}
+    </SuomifiThemeConsumer>
+  );
+};
```
ketsappi, line 217: This seems to miss the `displayName`.

src/core/Icon/Icon.tsx
```diff
@@ -27,21 +26,6 @@ export interface IconBaseProps {
 
 export interface IconProps extends IconBaseProps, SuomifiIconInterface {}
 
-export const iconLogger = (
```
ketsappi, line 30: I agree that we should remove this icon logger. I think that it does not really give as any real value.

src/core/Expander/Expander/Expander.tsx
ketsappi: `displayName` is now added.

ketsappi (approved):
LGTM.

Did test with styleguidist and CRA-TS testproject. Seems to work like before the change.

Riippi (approved):
LGTM

PR #645, opened 2022-07-20
[Fix] Circular dependency

Riippi:
## Description
Fixes a circular dependency warning

## Related Issue

Closes #

## Motivation and Context
The warning is probably not very serious as everything has been working fine. But this fix makes the build output shorter by one line and it's nice to have less warnings. 

## How Has This Been Tested?
Running build, tests, styleguidist.

Screenshot 2022-07-20 at 14 52 21

### DIFF
--- a/src/core/Expander/Expander/__snapshots__/Expander.test.tsx.snap
+++ b/src/core/Expander/Expander/__snapshots__/Expander.test.tsx.snap
@@ -112,28 +112,37 @@ exports[`Basic expander shoud match snapshot 1`] = `
   box-sizing: border-box;
 }
 
-.c5 {
-  vertical-align: baseline;
+.c1 {
+  color: hsl(0,0%,16%);
+  -webkit-letter-spacing: 0;
+  -moz-letter-spacing: 0;
+  -ms-letter-spacing: 0;
+  letter-spacing: 0;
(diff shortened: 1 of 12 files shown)


src/core/Expander/Expander/__snapshots__/Expander.test.tsx.snap
```diff
@@ -112,28 +112,37 @@ exports[`Basic expander shoud match snapshot 1`] = `
...
-.c5.fi-icon {
-  display: inline-block;
-}
-
-.c5 .fi-icon-base-fill {
-  fill: currentColor;
-}
-
-.c5 .fi-icon-base-stroke {
-  stroke: currentColor;
```
riitasointi, line 128: Why have all these snapshots changed? 🤔

Riippi: I don't know. This has happened at least once before with some random PR.
I guess once in a while something mysterious happens with the system that creates the snapshots and css classes are spit out in a different order.  

Googling hints that this might have something to do with styled components
https://github.com/styled-components/jest-styled-components/issues/289

riitasointi (approved):
LGTM

PR #646, opened 2022-07-20
[Fix] Remove prettier-check dependency

Riippi:
## Description
Removed prettier-check dependency and start using `prettier --check` instead. 
The output is very similar so nothing is missing. Also doesn't add any new information to the developer, but seems that the current status is ok.  
Adding prettylint would add a bit more informative output, but seems that it's not worth it to add a new dependency for that. It's easy to check the issue details with IDE.

## Related Issue

Closes [#SFIDS-202](https://jira.dvv.fi/browse/SFIDS-202)

## Motivation and Context
Removes a dependency that is not needed anymore.

## How Has This Been Tested?
Comparing the checks from cli. 

## Release notes
* Remove prettier-check dependency


riitasointi (dismissed):
Seems to work well 👍

package.json
```diff
@@ -55,7 +55,7 @@
     "lint:eslint": "eslint src",
     "lint:stylelint": "stylelint \"./src/**/*.{ts,tsx}\"",
     "prettier": "prettier --write \"src/**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
-    "prettier:check": "prettier-check \"src/**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
+    "prettier:check": "prettier --check \"src/**/*.{js,jsx,ts,tsx,json,css,scss,md}\"",
```
riitasointi, line 58: This should probably be changed for the `pre-commit` hook as well

Riippi: Good catch. Fixed now.

riitasointi (approved):

PR #647, opened 2022-07-21
[Fix] Update outdated dependencies

Riippi:
## Description
Updates selected npm dependencies, where the update causes only minimal changes or no changes at all.

## Related Issue

Closes #

## Motivation and Context
It's good to update the dependencies once in a while.

## How Has This Been Tested?
Styleguidist Firefox, Chrome, Safari. CRA Firefox, Chrome, Safari.

## Release notes
* Update dependencies


.eslintrc
```diff
@@ -127,6 +127,7 @@
     "react/jsx-wrap-multilines": "off",
     "react/jsx-one-expression-per-line": "warn",
     "react/jsx-boolean-value": "off",
+    "react/require-default-props": "off",
```
riitasointi, line 130: What is the idea behind this?

package.json
```diff
@@ -70,99 +70,100 @@
...
-    "@testing-library/jest-dom": "5.16.2",
+    "@rollup/plugin-commonjs": "22.0.1",
+    "@rollup/plugin-node-resolve": "13.3.0",
+    "@stylelint/postcss-css-in-js": "0.38.0",
+    "@testing-library/jest-dom": "5.16.4",
     "@testing-library/react": "12.1.4",
     "@types/classnames": "2.3.1",
-    "@types/jest": "27.4.1",
-    "@types/jest-axe": "3.5.3",
+    "@types/jest": "28.1.6",
```
riitasointi, line 85: There are some major release updates in these dependencies, how has this been tested?

.eslintrc
Riippi: That rule has not been active before, but in updated eslint it is and it shows a lot of errors in the interfaces. It would want all the optional props to have a default value. I think that is not something we want to have. 
```
export interface Something {
   text?: string
}
```
in this example there should be some default value for `text`

package.json
Riippi: Styleguidist Firefox, Chrome, Safari. CRA Firefox, Chrome, Safari.

riitasointi (approved):

PR #651, opened 2022-08-22
[Feature] ServiceNavigation

riitasointi:
## Description

This PR introduces the `<ServiceNavigation>` component. As children, it takes `<ServiceNavigationItem>` components which in turn takes a ReactNode as child. In practice, the child of a `<ServiceNavigationItem>` is supposed to be a `<RouterLink>` which is polymorphic and can thus be rendered as any semantic HTML element as well as third party library elements such as React Router Link, Gatsby Link etc. 

## Motivation and Context

## How Has This Been Tested?

Styleguidist, CRA app, NextJS app

## Screenshots
image

### ServiceNavigation
* Introduce `<ServiceNavigation>` component

### DIFF
--- a/src/core/Navigation/ServiceNavigation/ServiceNavigation/ServiceNavigation.tsx
+++ b/src/core/Navigation/ServiceNavigation/ServiceNavigation/ServiceNavigation.tsx
@@ -25,8 +25,7 @@ export interface ServiceNavigationProps {
    * @default true
    */
   initiallyExpanded?: boolean;
-  /** When using smallScreen variant, this is the text showed in the expander button.
-   * Intended to be the same text as the currently selected nav item.  */
+  /** When using smallScreen variant, this is the text showed in the expander button */
   smallScreenExpandButtonText?: string | ReactNode;
   /** Custom classname to extend or customize */
   className?: string;
--- a/src/core/Navigation/ServiceNavigation/ServiceNavigation/ServiceNavigation.tsx
+++ b/src/core/Navigation/ServiceNavigation/ServiceNavigation/ServiceNavigation.tsx
@@ -11,7 +11,7 @@ export interface ServiceNavigationProps {
    * Normal or small screen variant
    * @default normal
    */
-  variant: 'normal' | 'smallScreen';
+  variant?: 'normal' | 'smallScreen';
   /** Custom classname to extend or customize */
   className?: string;
 }
--- a/src/core/Navigation/ServiceNavigation/ServiceNavigation/ServiceNavigation.tsx
+++ b/src/core/Navigation/ServiceNavigation/ServiceNavigation/ServiceNavigation.tsx
@@ -1,9 +1,16 @@
-import React, { ReactNode } from 'react';
-import { HtmlNav, HtmlUl } from '../../../../reset';
+import React, { ReactNode, useState } from 'react';
+import {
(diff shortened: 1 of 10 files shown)


src/core/Navigation/ServiceNavigation/ServiceNavigationItem/ServiceNavigationItem.baseStyles.tsx
```diff
@@ -0,0 +1,103 @@
...
+        &:visited {
+          text-decoration: none;
+          color: ${theme.colors.blackBase};
+        }
+      }
+    }
+  }
+`;
+
+/* stylelint-enable no-descending-specificity */
```
Riippi, line 103: Is there a reason why this is outside of the block when the stylelint-disable is inside the block ?

riitasointi: This line was actually completely redundant so I removed it

Riippi:
Seems that going through the links with Tab doesn't work on Styleguidist  + Firefox. Chrome works better.

Riippi (commented):
Not sure how serious this is, but focus highlight is "blocked" by the current ServiceNavigationItem. On small screen variant the line is 1px thinner? Maybe the current one has higher z-index?
Screenshot 2022-10-10 at 10 42 08
Screenshot 2022-10-10 at 12 19 17

riitasointi:
> Not sure how serious this is, but focus highlight is "blocked" by the current ServiceNavigationItem. On small screen variant the line is 1px thinner? Maybe the current one has higher z-index? Screenshot 2022-10-10 at 10 42 08 Screenshot 2022-10-10 at 12 19 17

This has now been adjusted!

src/core/Navigation/ServiceNavigation/ServiceNavigation/ServiceNavigation.tsx
```diff
@@ -0,0 +1,126 @@
...
+        [smallScreenClassName]: variant === 'smallScreen',
+      })}
+    >
+      {variant === 'smallScreen' && (
+        <HtmlButton
+          className={smallScreenExpandButtonClassName}
+          onClick={() => setSmallScreenNavOpen(!smallScreenNavOpen)}
+          aria-expanded={smallScreenNavOpen}
+        >
+          <HtmlSpan
```
jenkrisu, line 74: Should these styles come as class from baseStyles?

riitasointi: Not sure what you mean here? Line 71 toggles between showing and hiding the nav, whereas line 72 renders the aria-expanded attribute based on whether the nav is opened or not

jenkrisu: Comment section shows this a bit strangely, but I meant rows 75-79 that have HtmlSpan component's styles inlined.

riitasointi: Ah, I see. These were indeed meant to be moved to a CSS class but I forgot about it. Changed it now!

Riippi (approved):
LGTM

PR #652, opened 2022-08-23
[Fix] Label content types

LJKaski:
## Description
This PR changes the prop type of label content from `string` to `ReactNode` in all applicable components. This allows the labels to contain other elements than just text, such as tooltip icons etc. 

The Label component itself that's used by many of the components already supports ReactNode as content, but the APIs of the components were dragging behind.

## Motivation and Context
There is current and future need for more varied content in the labels.

## How Has This Been Tested?
Tested by running styleguidist locally on Chrome and adding different kinds of content to labels. Also checked that screen readers behave predictably. Optional text works as before in most cases as well.

### Dropdown, CheckboxGroup, RadiobuttonGroup, FilterInput, SearchInput, MultiSelect, SingleSelect, TextInput, TextArea
* **Breaking change:** Change label content type from `string` to `ReactNode` (#652)

### DIFF
--- a/src/core/Alert/InlineAlert/InlineAlert.tsx
+++ b/src/core/Alert/InlineAlert/InlineAlert.tsx
@@ -27,7 +27,7 @@ export interface InlineAlertProps extends HtmlDivWithRefProps {
   /** Use small screen styling */
   smallScreen?: boolean;
   /** Label for the alert */
-  labelText?: string;
+  labelText?: ReactNode;
   /** Set aria-live mode for the alert text content and label.
    * @default 'assertive'
    */
--- a/src/core/Dropdown/Dropdown/Dropdown.tsx
+++ b/src/core/Dropdown/Dropdown/Dropdown.tsx
@@ -56,7 +56,7 @@ interface InternalDropdownProps {
   /** Controlled selected value, overrides defaultValue if provided. */
   value?: string;
   /** Label for the Dropdown component. */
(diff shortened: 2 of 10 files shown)


riitasointi (commented):
Should we have an example somewhere which shows a more complicated use of label? Like some custom component

ketsappi:
Would this be good chance to unify the way we have JSDocs for the `labelText` component?

Here are the different types of `labelText` JSDocs (just from this PR). Would the ` /** Label */` be most appropriate as it can now be ReactNode? (Of course the actual prop name says text, but still)

```diff
   /** Label for the Dropdown component. */
-  labelText: string;
+  labelText: ReactNode;

   /** Label for the group */
-  labelText: string;
+  labelText: ReactNode;

   /** Label */
-  labelText: string;
+  labelText: ReactNode;

   /** Label for the group */
-  labelText: string;
+  labelText: ReactNode;

   /** Label text */
-  labelText: string;
+  labelText: ReactNode;

   /** Label text */
-  labelText: string;
+  labelText: ReactNode;

   /** Label */
-  labelText: string;
+  labelText: ReactNode;

   /** Label */
-  labelText: string;
+  labelText: ReactNode;

   /** Label */
-  labelText: string;
+  labelText: ReactNode;
```

riitasointi (approved):
Looks good!

ketsappi (approved):

PR #656, opened 2022-09-13
[Feature] WizardNavigation

riitasointi:
## Description

This PR introduces the `<WizardNavigation>` component. As children, it takes `<WizardNavigationItem>` components which in turn takes a ReactNode as child. In practice, the child of a `<WizardNavigationItem>` is supposed to be a `<RouterLink>` which is polymorphic and can thus be rendered as any semantic HTML element as well as third party library elements such as React Router Link, Gatsby Link etc. 

A `<WizardNavigationItem>` can have one of these 5 statuses (affects styling and link availability):

- default: reachable & incomplete step
- current: currently active step
- completed: a step where the user has filled all necessary information
- current-completed: a combination of current and completed statuses
- coming: a step which is unreachable at the moment (but will become available when e.g. the previous steps have been completed)
- disabled: disabled step, not reachable. We do not recommend the use of this status unless absolutely necessary

## Motivation and Context

## How Has This Been Tested?

Styleguidist, CRA app

## Screenshots
image

### WizardNavigation
* Introduce `<WizardNavigation>` component

### DIFF
--- a/src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.md
+++ b/src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.md
@@ -1,6 +1,6 @@
 ### Basic usage
 
-A `<WizardNavigationItem>` can have one of these 5 statuses:
+A `<WizardNavigationItem>` can have one of these 6 statuses:
 
 - default: reachable & incomplete step
 - current: currently active step
--- a/src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.md
+++ b/src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.md
@@ -24,17 +24,17 @@ const Comp = (props) => {
 };
 
 <div style={{ width: '350px' }}>
-  <WizardNavigation heading="Steps">
+  <WizardNavigation heading="Steps" aria-label="Main">
     <WizardNavigationItem status="completed">
-      <RouterLink href="https://suomi.fi">Step 1</RouterLink>
+      <RouterLink
+        href="https://suomi.fi"
+        aria-label="Step 1. This step is completed"
+      >
+        Step 1
+      </RouterLink>
     </WizardNavigationItem>
     <WizardNavigationItem status="default">
-      <RouterLink
-        href="#"
-        aria-label="Step 2. This step is completed"
-      >
-        Step 2
-      </RouterLink>
+      <RouterLink href="#">Step 2</RouterLink>
     </WizardNavigationItem>
     <WizardNavigationItem status="current">
       <RouterLink aria-current="step" href="#">
@@ -42,28 +42,23 @@ const Comp = (props) => {
       </RouterLink>
     </WizardNavigationItem>
     <WizardNavigationItem status="coming">
-      <RouterLink aria-disabled tabIndex={-1} href="#">
+      <RouterLink aria-disabled role="link">
         Step 4 with a long text that wraps to the second line like
         this
       </RouterLink>
     </WizardNavigationItem>
     <WizardNavigationItem status="coming">
-      <RouterLink aria-disabled tabIndex={-1} href="#">
+      <RouterLink aria-disabled role="link">
         Step 5
       </RouterLink>
     </WizardNavigationItem>
     <WizardNavigationItem status="disabled">
-      <RouterLink aria-disabled tabIndex={-1} href="#">
+      <RouterLink aria-disabled role="link">
         Step 6
       </RouterLink>
     </WizardNavigationItem>
     <WizardNavigationItem status="coming">
-      <RouterLink
-        asComponent={Comp}
-        role="button"
-        aria-disabled
-        tabIndex={-1}
-      >
+      <RouterLink asComponent={Comp} role="button" aria-disabled>
         Step 7
       </RouterLink>
     </WizardNavigationItem>
@@ -100,6 +95,7 @@ const Comp = (props) => {
     variant="smallScreen"
     heading="Steps"
     initiallyExpanded={false}
+    aria-label="Main"
   >
     <WizardNavigationItem status="default">
       <RouterLink href="https://suomi.fi">Step 1</RouterLink>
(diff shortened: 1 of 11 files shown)


src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { ReactNode, useState } from 'react';
+import { HtmlButton, HtmlDiv, HtmlNav, HtmlUl } from '../../../../reset';
+import styled from 'styled-components';
+import { SuomifiThemeConsumer } from '../../../theme';
+import { baseStyles } from './WizardNavigation.baseStyles';
+import classnames from 'classnames';
+import { Icon } from '../../../Icon/Icon';
+
+export interface WizardNavigationProps {
+  children: ReactNode | ReactNode[];
```
LJKaski, line 10: Is the array option really needed? ReactNode should support ReactFragment and an array or ReactNodes by default.

src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.tsx
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Normal or small screen variant
+   * @default normal
+   */
+  variant?: 'default' | 'smallScreen';
+  /**
+   * Whether the menu is initially expanded. Only applies to smallScreen variant.
+   * @default true
+   */
+  initiallyExpanded?: boolean;
```
LJKaski, line 20: Since the `initiallyExpanded` only applies to the smallScreen variant, should this be a conditional prop? Probably not a big deal but worth a discussion I think.

riitasointi: Fair point, but as far as I understand the only benefit you would get out of it is Typescript not accepting the `initiallyExpanded` prop if the variant !== smallScreen. So not a huge thing IMO

src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.tsx
riitasointi: I didn't realise ReactNode actually accepts array by default. Removed the redundant part.

src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.tsx
LJKaski: Yeah, it's more of a matter of principle with these. And judging from past experiences this more strict typing can be more of an annoyance for the users. So I'm fine with keeping it as it is.

src/reset/HtmlNav/HtmlNav.tsx
```diff
@@ -6,6 +6,7 @@ import { asPropType } from '../../utils/typescript';
 export interface HtmlNavProps
   extends Omit<HTMLProps<HTMLElement>, 'ref' | 'as'> {
   as?: asPropType;
+  forwardedRef?: React.RefObject<HTMLDivElement>;
```
Riippi, line 9: I think the type should be `<HTMLElement>`.  If I have understood correctly nav doesn't have its own type, so `<HTMLElement>` would be the proper one.
And if there is no specific reason for it to be `RefObject` the whole thing should be `React.Ref<HTMLElement>`

src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.tsx
```diff
@@ -0,0 +1,114 @@
...
+  /** Navigation component heading text */
+  heading: string;
+  /** Custom classname to extend or customize */
+  className?: string;
+  /** Ref is forwarded to nav element. Alternative for React `ref` attribute. */
+  forwardedRef?: React.RefObject<HTMLDivElement>;
```
Riippi, line 32: Also in here the type could be `React.Ref<HTMLElement>` if it's changed in HtmlNav

riitasointi: Fixed this!

src/reset/HtmlNav/HtmlNav.tsx
riitasointi: Fixed this!

Riippi (commented):
Is the focus highlight ok?  Seems that it has maybe 1px empty padding inside it that is not in Figma
Screenshot 2022-09-30 at 9 02 34

riitasointi:
> 

Ah, there was `boxShadowFocus` style available which does not overflow the div. Changed!

src/core/Navigation/WizardNavigation/WizardNavigation/__snapshots__/WizardNavigation.test.tsx.snap
```diff
@@ -816,6 +816,7 @@ exports[`calling render with the same component on the same container does not r
     />
     <ul
       class="c3 fi-wizard-navigation_list"
+      role="list"
```
jenkrisu, line 819: As <ul> implicit ARIA role is list, adding it just in case seems a bit redundant. Or does it make some difference with assistive technologies?

src/core/Navigation/WizardNavigation/WizardNavigation/WizardNavigation.md
```diff
@@ -0,0 +1,138 @@
+### Basic usage
+
+A `<WizardNavigationItem>` can have one of these 5 statuses:
```
jenkrisu, line 3: 6 statuses instead of 5?

src/core/Navigation/WizardNavigation/WizardNavigation/__snapshots__/WizardNavigation.test.tsx.snap
riitasointi: This is on purpose, just as a precaution due to this: https://developer.mozilla.org/en-US/docs/Web/CSS/list-style#accessibility_concerns

jenkrisu: Ok, good to know!

Riippi (approved):
LGTM

PR #657, opened 2022-09-14
[Feature] Add plain div and span as variants for Block

riitasointi:
## Description

This PR adds two new variant options for the `<Block>` component:
- div
- span

## Motivation and Context
The existing `default` variant renders the component as a `<div>`, but it applies the reset styles from the suomifiTheme (including margin: 0 and padding: 0). This has been deemed cumbersome by our users. Now the `<Block>` component can be rendered as a plain div without the reset styles. 

In addition, there was a request to add `span` as one of the available variants. Span renders as an inline-block without reset styles

## How Has This Been Tested?
Styleguidist

### Block
* Add `div` and `span` as a possible variants. "Div" renders as a plain HTML div without suomifiTheme div reset styles. "Span" renders as an inline-block without suomifiTheme reset styles

### DIFF
--- a/src/core/Block/Block.baseStyles.ts
+++ b/src/core/Block/Block.baseStyles.ts
@@ -2,8 +2,9 @@ import { css } from 'styled-components';
 import { SuomifiTheme } from '../theme';
 import { element, font } from '../theme/reset';
 import { spacingModifiers } from '../theme/utils';
+import { BlockVariant } from './Block';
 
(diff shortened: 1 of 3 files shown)


Riippi (dismissed):
LGTM

LJKaski (approved):
Seems to work well 👍

LJKaski (dismissed):
LGTM :)

LJKaski (approved):

PR #658, opened 2022-09-16
[Feature] SideNavigation

riitasointi:
## Description

This PR introduces the `<SideNavigation>` component. As children, it takes `<SideNavigationItem>` components.

Differing from the other two nav components, a `<SideNavigationItem>` can have other `<SideNavigationItem>`s as children. The tree can go up to 3 levels deep. Because of this structure, the API of this component also differs slightly compared to `<WizardNavigation>` and `<ServiceNavigation>`: There is a mandatory `subLevel` prop (1-3) and the actual link content of the item must be given into the `content` prop (ReactNode). In practice, the content prop is supposed to be a `<RouterLink>` which is polymorphic and can thus be rendered as any semantic HTML element as well as third party library elements such as React Router Link, Gatsby Link etc. 

Example: 
```
<SideNavigation heading="Economy" icon="piggyBank">
  <SideNavigationItem
    subLevel={1}
      content={
        <RouterLink href="/" aria-current="location">
          Personal economy
        </RouterLink>
      }
    >
      <SideNavigationItem
        subLevel={2}
        content={
          <RouterLink href="/" aria-current="location">
            Crisis situations in personal finances
          </RouterLink>
        }
      >
        <SideNavigationItem
          subLevel={3}
          content={
            <RouterLink href="/">
              If you are unable to pay your debts
            </RouterLink>
          }
        />
        <SideNavigationItem
          subLevel={3}
          selected
          content={
            <RouterLink href="/" aria-current="page">
              Advice on banking and insurance matters
            </RouterLink>
          }
        />
      </SideNavigationItem>
  </SideNavigationItem>
</SideNavigation>
```

## Motivation and Context

## How Has This Been Tested?

Styleguidist, CRA app

## Screenshots
image

### SideNavigation
* Introduce `<SideNavigation>` component

### DIFF
--- a/src/core/Navigation/SideNavigation/SideNavigation/SideNavigation.tsx
+++ b/src/core/Navigation/SideNavigation/SideNavigation/SideNavigation.tsx
@@ -50,6 +50,7 @@ const smallScreenClassName = `${baseClassName}--small-screen`;
 const listClassName = `${baseClassName}_list`;
 const headingClassName = `${baseClassName}_heading`;
 const headingInnerWrapperClassName = `${baseClassName}_heading_inner`;
+const dividerClassName = `${baseClassName}_divider`;
 
 const BaseSideNavigation = ({
   children,
@@ -100,6 +101,7 @@ const BaseSideNavigation = ({
           id={id}
           {...getConditionalAriaProp('aria-label', [ariaLabel])}
         >
+          <HtmlDiv className={dividerClassName} />
           <HtmlUl className={listClassName}>{children}</HtmlUl>
         </HtmlNav>
       )}
--- a/src/core/Navigation/SideNavigation/SideNavigation/SideNavigation.tsx
+++ b/src/core/Navigation/SideNavigation/SideNavigation/SideNavigation.tsx
@@ -1,0 +1,108 @@
+import React, { ReactNode, useState } from 'react';
+import {
+  HtmlButton,
+  HtmlDiv,
+  HtmlNav,
+  HtmlSpan,
+  HtmlUl,
+} from '../../../../reset';
+import styled from 'styled-components';
+import { SuomifiThemeConsumer } from '../../../theme';
+import { baseStyles } from './SideNavigation.baseStyles';
+import classnames from 'classnames';
+import { Icon } from '../../../Icon/Icon';
+import {
+  StaticIcon,
+  IllustrativeIconKeys,
+  DoctypeIconKeys,
+} from '../../../StaticIcon/StaticIcon';
+
+export interface SideNavigationProps {
+  children: ReactNode | ReactNode[];
+  /**
+   * Normal or small screen variant
+   * @default normal
+   */
+  variant?: 'default' | 'smallScreen';
+  /**
+   * Whether the menu is initially expanded. Only applies to smallScreen variant.
+   * @default true
+   */
+  initiallyExpanded?: boolean;
+  /** Navigation component heading text */
+  heading: string;
+  /** Icon in the nav heading. Options include suomifi static icons. */
+  icon?: IllustrativeIconKeys | DoctypeIconKeys;
+  /* Custom class to extend or customise */
+  className?: string;
+}
+
+const baseClassName = 'fi-side-navigation';
+const smallScreenClassName = `${baseClassName}--small-screen`;
+const listClassName = `${baseClassName}_list`;
+const headingClassName = `${baseClassName}_heading`;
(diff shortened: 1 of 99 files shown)


src/core/Navigation/SideNavigation/SideNavigationItem/SideNavigationItem.tsx
```diff
@@ -0,0 +1,117 @@
...
+          if (c.props) {
+            if (c.props.selected) {
+              childIsSelected = true;
+              return;
+            }
+          }
+        });
+      }
+    }
+  });
```
LJKaski, line 62: I feel like this could be more readable done otherwise. Maybe somewhere along the lines of:

  React.Children.forEach(children, (element) => {
    if (!React.isValidElement(element) || !element.props) return;

    if (element.props.selected) {
      childIsSelected = true;
      return;
    }

    if (!element.props.children) return;

    element.props.children.forEach((child: ReactElement) => {
      if (child.props?.selected) {
        childIsSelected = true;
        return;
      }
    });
  });

riitasointi: Fair enough, fewer if-statements and they are not nested. 

Changed it like you suggested!

src/core/Navigation/SideNavigation/SideNavigation/SideNavigation.tsx
```diff
@@ -0,0 +1,134 @@
...
+  DoctypeIconKeys,
+} from '../../../StaticIcon/StaticIcon';
+import { getConditionalAriaProp } from '../../../../utils/aria';
+
+export interface SideNavigationProps {
+  /** Use the `<SideNavigationItem>` components as children */
+  children: ReactNode;
+  /** Name for the navigation element. Don't use the word "navigation" since it will be read by screen reader regardless. */
+  'aria-label': string;
+  /** HTML nav element will receive this id */
```
LJKaski, line 26: Just a style difference compared to the rest of the library, but caught my eye. Could this be "id passed to the HTML nav element" or "id for the HTML nav element"?

src/core/Navigation/SideNavigation/SideNavigation/SideNavigation.tsx
```diff
@@ -0,0 +1,134 @@
...
+  id?: string;
+  /**
+   * Normal or small screen variant
+   * @default normal
+   */
+  variant?: 'default' | 'smallScreen';
+  /**
+   * Whether the menu is initially expanded. Only applies to smallScreen variant.
```
LJKaski, line 34: Same as above. Maybe "Initial expand status for the menu. Only applies to smallScreen variant".

src/core/Navigation/SideNavigation/SideNavigation/SideNavigation.tsx
riitasointi: Good point. Changed it!

src/core/Navigation/SideNavigation/SideNavigation/SideNavigation.tsx
riitasointi: Good point. Changed it!

src/core/Navigation/SideNavigation/SideNavigationItem/SideNavigationItem.baseStyles.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+    &:after {
+      content: '';
+      position: absolute;
+      bottom: -1px;
+      left: 2.5%;
+      width: 95%;
+      height: 1px;
+      background: ${theme.colors.depthSecondary};
+    }
```
LJKaski, line 24: What is the purpose of this exactly? When colored red, it looks like this: 
image

LJKaski: Seems to have been a display issue for me. Couldn't distinguish the lines with laptop screen at first. However, this colorization seems to have highlighted a small issue with the separator lines.

Riippi (changes requested):
SideNavigationItem.md  file is empty. Should it be removed?

riitasointi:
> SideNavigationItem.md file is empty. Should it be removed?

.md files are needed for all components which we want to show the props list in Styleguidist. The codebase is actually filled with empty .md files in other components as well

Riippi (approved):

PR #662, opened 2022-09-27
[Feature] Ref to RouterLink

riitasointi:
## Description
This PR introduces support for React ref to `<RouterLink>`. It complies with the polymorphic nature of the component, maintaining Typescript type integrity on the ref side as well. 

Example: `<RouterLink asComponent="button" />` only accepts a ref the type of which is `HTMLButtonElement`.

Like our other components, this one now also has a prop called `forwardedRef` in its API. It works exactly the same as `ref`.

## Motivation and Context
We want to have support for React ref in all our components. This one was still missing it. 

## How Has This Been Tested?
Typescript manual tests, Styleguidist

### RouterLink
* Add support for React ref

### DIFF
--- a/src/core/Link/Link/Link.md
+++ b/src/core/Link/Link/Link.md
@@ -71,16 +71,28 @@ This component is mainly intended to be used with external libraries/frameworks.
 
 ```js
 import { RouterLink } from 'suomifi-ui-components';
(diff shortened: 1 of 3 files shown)


Riippi (approved):
LGTM

Riippi (approved):

PR #663, opened 2022-09-29
[Bugfix] Hidden Label semantics

LJKaski:
## Description
This PR addresses a bug where the hidden variant of Label was not semantically a label.

## Motivation and Context
The bug was reported by a service using the library. It's a usability and accessibility problem.

## How Has This Been Tested?
Tested in styleguidist on Chrome + Windows. 

### Label
* Fix hidden variant semantics


riitasointi (approved):
Seems to have the desired effect! 👍

PR #664, opened 2022-10-07
[Fix] Styleguidist HMR

jenkrisu:
## Description

Enables hot module replacement for styleguidist dev server.

## Motivation and Context

Styleguidist's HMR has stopped working at some point. No updates were being sent to browser after compiling.

## How Has This Been Tested?

Running dev server on Windows and Mac.

## Release notes

\-

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -12,7 +12,7 @@ After cloning suomifi-ui-components, run `yarn` to fetch its dependencies. Then,
 
 2. `yarn test` runs written tests.
 
-3. `yarn test:lint` checks TypeScript code for readability, maintainability, and functionality errors.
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -12,7 +12,7 @@ After cloning suomifi-ui-components, run `yarn` to fetch its dependencies. Then,
 
 2. `yarn test` runs written tests.
 
-3. `yarn test:lint` checks TypeScript code for readability, maintainability, and functionality errors.
+3. `yarn lint` checks TypeScript code for readability, maintainability, and functionality errors.
```
riitasointi, line 15: This change is unrelated to the actual subject matter of this PR, and we usually want to avoid that. But in this case it's fine because the change is so small and does not affect the library users (internal change)

PR #665, opened 2022-10-11
[Feature] Link underline option

jenkrisu:
## Description

Adds property `underline` to components `Link`, `ExternalLink`, `SkipLink`, and `RouterLink`.

## Motivation and Context

Links that have no underline can be hard to distinguish from other text content. Underlining links improves accessibility for users with decreased (or no) color discrimination.

## How Has This Been Tested?

Styleguidist and CRA.

### Link
links

### ExternalLink
external-link

### SkipLink
skip-link

## Release notes

- Add `underline` property to Link, ExternalLink, SkipLink, and RouterLink.

### DIFF
--- a/src/core/Link/SkipLink/SkipLink.baseStyles.ts
+++ b/src/core/Link/SkipLink/SkipLink.baseStyles.ts
@@ -13,7 +13,6 @@ export const SkipLinkStyles = (theme: SuomifiTheme) => css`
     background: ${theme.colors.highlightLight3};
     border: 1px solid ${theme.colors.depthLight1};
     color: ${theme.colors.blackBase};
-    text-decoration: none;
   }
 
   &.fi-link--skip:focus {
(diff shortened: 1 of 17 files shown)


src/core/Link/SkipLink/SkipLink.baseStyles.ts
```diff
@@ -13,7 +13,6 @@ export const SkipLinkStyles = (theme: SuomifiTheme) => css`
     background: ${theme.colors.highlightLight3};
     border: 1px solid ${theme.colors.depthLight1};
     color: ${theme.colors.blackBase};
-    text-decoration: none;
```
LJKaski, line 16: Is this a necessary removal, since the skip link is quite a unique implementation in general? Not sure if the underline really brings any extra value here.

jenkrisu: This actually doesn't remove the underline by default, since SkipLink gets the styles from the BaseLink. So this style was redundant here.

LJKaski: Ah, so it seems. Good removal in that case 😁

LJKaski (approved):
Looks clean and seems to be working well. Good work :)

PR #669, opened 2022-11-04
[Feature] Integrate Tooltip to form components

riitasointi:
## Description

This PR integrates the `<Tooltip>` component into other applicable components. These include form components which use the `<Label>` component under the hood. Tooltip can now be neatly passed into the components like this:

```
<TextInput
  labelText="TextInput with a tooltip"
  tooltipComponent={
    <Tooltip
      ariaToggleButtonLabelText="TextInput with a tooltip, additional information"
      ariaCloseButtonLabelText="TextInput with a tooltip, close additional information"
    >
      <Heading variant="h5" as="h2">
        Tooltip
      </Heading>
      <Text>Text content for the tooltip</Text>
    </Tooltip>
  }
/>
```

While testing, encountered an issue with the Label component's current Tooltip implementation. It had an "always on" type of trick to force the Tooltip to always re-render when the label changed. This led to problems when the Label+Tooltip combo was present in more complicated components like SingleSelect which does constant re-rendering due to state changes. Therefore, removed the re-render forcing and added it as an optional prop in the Label component. The prop is called `forceTooltipRerender` and turns on the previous behaviour when needed. 

## Motivation and Context

Make our Tooltip component more usable by making it possible to use the component directly with all our applicable form components

## How Has This Been Tested?
Styleguidist, CRA

## Screenshots
image
image

### Tooltip
* Add possibility to use Tooltip directly in the following components (usage via `tooltipComponent` prop)
    - TextInput
    - TextArea
    - SingleSelect
    - MultiSelect
    - Dropdown
    - CheckboxGroup
    - RadioButtonGroup

### Label
* Add optional `forceTooltipRerender` boolean prop. The prop toggles on behaviour which forces a Tooltip component attached to a Label to always re-render when the Label changes


Riippi (commented):
Safari cuts out a small slice of the icon
Screenshot 2022-11-08 at 15 34 56

riitasointi:
> Safari cuts out a small slice of the icon Screenshot 2022-11-08 at 15 34 56

I think this happens with some monitor + resolution + zoom level combinations and not on others. I'd say it's not worth it to try and fix it, such a minor issue

src/core/Tooltip/TooltipContent/TooltipContent.tsx
```diff
@@ -51,7 +52,9 @@ class BaseTooltipContent extends Component<TooltipContentProps & InnerRef> {
       <HtmlDivWithRef
         {...passProps}
         forwardedRef={forwardedRef}
-        className={classNames(tooltipContentClassNames.content, className)}
+        className={classNames(tooltipContentClassNames.content, className, {
+          [tooltipContentClassNames.arrowHidden]: arrowOffsetPx < 1,
```
jenkrisu, line 56: Should this be arrowOffsetPx <= 0 in case of sub pixels, or does it matter in this context?

jenkrisu (approved):
Works nicely with inputs!

Riippi (approved):
LGTM

PR #670, opened 2022-11-07
[Fix] Add passProps for the hidden Label

Riippi:
## Description
This PR addresses a bug where the hidden variant of Label was missing the passed props

## Motivation and Context
The bug was noticed when developing Pagination component. For example for and id tags were not properly passed to VisuallyHidden inside the Label.

## How Has This Been Tested?
Firefox and Chrome on mac. Styleguidist.

### Label
* Fix passed props on hidden variant


LJKaski (approved):
Seems to be working well 👍

jenkrisu:
Good catch 👍

jenkrisu (approved):

PR #671, opened 2022-11-15
[Fix/BreakingChange] Multiselect onitemselect return type

riitasointi:
## Description

This PR removes `null` as a possible parameter in MultiSelect's `onItemSelect()` prop. Null was there because the code had been copied over from SingleSelect where `onItemSelect()` gets called also when the input is cleared. In MultiSelect there is no such behavior.

## Motivation and Context

This was reported as a bug by one of our users.

## How Has This Been Tested?

Styleguidist

### MultiSelect
* **Breaking change:** Remove redundant `null` from possible parameters in `onItemSelect()` prop


Riippi (dismissed):
LGTM

jenkrisu (approved):

LJKaski (approved):

PR #679, opened 2022-12-07
[Fix]: MultiSelect & SingleSelect: Associate label with input

riitasointi:
## Description

This PR connects the label and input in `<SingleSelect>` & `<MultiSelect>` (using the underlying `<FilterInput>` component). The label gets the input's ID as its `htmlFor` prop. 

NOTE: I tried to do the same thing for our `<Dropdown>`, but ReachUI's Listbox component does not allow an ID to be passed to the actual underlying `<input>` element. So no can do. 

## Motivation and Context

We want SingleSelect & MultiSelect to gain focus when their label is clicked

## How Has This Been Tested?
Styleguidist

### SingleSelect & MultiSelect
* Connect the component's label to its input via the `for` attribute to allow focus on label click


Riippi (approved):
LGTM

jenkrisu (approved):

PR #682, opened 2022-12-27
[Fix] SideNav: Remove aria-current=location from examples

riitasointi:
## Description

This PR removes aria-current="location" attributes from `<SideNavigation>` examples. 

## Motivation and Context

This came up in accessibility clinic. Having aria-current="location" in the parents of the currently active page was seen as more confusing than helpful. Having aria-current="page" in the item which is actually active is enough.

## How Has This Been Tested?
Styleguidist, VoiceOver

### SideNavigation
* Remove aria-current="location" attributes from examples


LJKaski (approved):
LGTM

LJKaski (approved):

PR #685, opened 2022-12-30
[Feature] Tree shakeable icons

LJKaski:
## Description
This PR implements the new suomifi-iconsversion to ui-components. In practice it replaces all `icon` props with props that accept a ReactElement instead of icon name as string. This allows for individual icon components to be given as props where needed. This should eliminate the need to bundle all the icons with ui-components "just in case" in order to support the `icon` props and thus dramatically reduce bundle size.

This PR also adjusts the vertical alignment of button icons in order to make them match the design.

**This PR should be published as a stable release only after `suomifi-icons` 7.0.0 release is out**

## Motivation and Context
Previously all suomifi-icons needed to be bundled with ui-components because some components took icon names as props and rendered the icon components based on the given name. This lead to an increased bundle size and inoptimal load times.

## How Has This Been Tested?
Mac: Chrome + Safari (and VoiceOver) in Styleguidist and custom CRA
Windows: Chrome, Edge, Firefox (and NVDA) in Styleguidist, CRA, NextJS
iOS: Safari, Chrome (and VoiceOver) in Styleguidist and Next.js project
Android: Firefox, Chrome

### General changes
* **Breaking change:** All `icon` props now take a `ReactElement` as value instead of a string.
### Dependencies
* Update suomifi-icons to 7.0.0
### Icon
* **Breaking Change:** Icon component no longer available through `suomifi-ui-components`. Individual icon components with same api as former ui-components implementation can be imported straight from `suomifi-icons`


.styleguidist/icons.md
```diff
@@ -0,0 +1,95 @@
+Suomi.fi icons
```
riitasointi, line 1: It's great to have all the icons displayed here!

I was just thinking, we don't have any place (neither here or in the icons repo) where there would be clear instruction how to use the components. Like:

```
import { IconAlert } from 'suomifi-icons'

<IconAlert color="red"/>
```

riitasointi (approved):

PR #686, opened 2023-01-16
[Fix] Tooltip close button size and appearance

LJKaski:
## Description
This PR changes the color, size and position of the tooltip content close button so that it meets the accessibility guidelines and matches designs.

During testing I also noticed that the snapshot test didn't catch changes made to the opened tooltip content, so I changed the test to open the tooltip before taking the snapshot.

## Motivation and Context
The close button in the tooltip content was unnecessarily small. The space could be utilized better so that the button size also better matches the accessibility guidelines. At the same time the color and size of the icon was adjusted so that it matches the design.

## How Has This Been Tested?
Tested by checking against designs and testing in styleguidist on Chrome and Firefox

Tooltip close button hover state:
image

### Tooltip
* **Breaking Change**: Change close button size and appearance.

### DIFF
--- a/src/core/Tooltip/Tooltip.test.tsx
+++ b/src/core/Tooltip/Tooltip.test.tsx
@@ -154,6 +154,8 @@ describe('Basic tooltip', () => {
       </Tooltip>
     );
     const { container } = render(BasicTooltip);
+    const toggleButton = screen.getAllByRole('button')[0];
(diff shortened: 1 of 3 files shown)


jenkrisu (dismissed):
Good catch about the tests!

jenkrisu (approved):

PR #690, opened 2023-01-31
[Breaking change] Derived tokens from tokens library

riitasointi:
## Description

This PR streamlines the library's design token implementation by importing all tokens from the `suomifi-design-tokens` npm package instead of having some of them coded in the ui-components repo. 

This also means some breaking changes in naming. The following token categories have changed
* radius --> radiuses
* focus --> focuses

So now you would use them as thus `suomifiDefaultTheme.focuses.absoluteFocus`

`suomifi-design-tokens` version is now 5.0.0 which also means a slight change in blackBase color. 

## Motivation and Context

All tokens are now available in the `suomifi-design-tokens` package and it makes sense to use them from there. 

## How Has This Been Tested?

Styleguidist, CRA-app

### SuomifiTheme
* **Breaking change:** Rename a couple of token categories
   - focus ---> focuses
   - radius ---> radiuses
* **Breaking change**: update `blackBase` color from `hsl(0, 0%, 16%)` to `hsl(0, 0%, 13%)`
* Refactor underlying implementation to use `suomifi-design-tokens` package for all tokens


LJKaski (approved):
Seems to be working well. Well done :)

PR #694, opened 2023-03-03
[Fix] Dynamic content snapshots

LJKaski:
## Description
This PR adjusts tests for SingleSelect, MultiSelect and LanguageMenu so that the snapshot tests test the opened state of the components instead of the initial state.

## Motivation and Context
Some components with dynamic content such as single- and multiselect had snapshot tests that only tested the initial state of the component, which left a lot of content outside the snapshot. Thus something could have been broken or changed inside the component without it being caught by the test.

## How Has This Been Tested?
By running tests and manually checking the contents of the new snapshots.

## Release notes
- none needed


jenkrisu (approved):
👍 I'd drop the bug label since this doesn't change behavior of any component, just improvement of tests

riitasointi (approved):

PR #696, opened 2023-03-06
[Breaking change] Dropdown without Reach

riitasointi:
## Description

This PR introduces a complete rewrite of the `<Dropdown>` component, implementing the component without ReachUI dependencies. 

The API has stayed as similar as possible, but some changes were inevitable. See release notes for changes and additions.

`<DateInput>` uses Dropdown, so some API related as well as minor visual changes had to be made there.  

## Motivation and Context

ReachUI has ceased development and it does not support React 18 anymore. We need to replace Reach stuff with our own code. 

## How Has This Been Tested?

Styleguidist on Mac Chrome, Safari, Firefox. VoiceOver on Mac. CRA with react 18

## Screenshots
image

### Dropdown
- **Breaking change**: Completely rewrite the component without using ReachUI dependency
    - Remove props `dropdownButtonProps` and `dropdownPopoverProps` (these were provided by Reach)
    - Change ref type from `HtmlDivElement` to `HtmlButtonElement`
    - Add boolean prop `portal` which defaults to true. Controls whether the popover is rendered in a React portal
    - Add props `status`, `statustext` and `hintText`. These are familiar from the library's other form components
    - Add `onBlur()` prop
    - Add boolean prop `fullWidth` which makes the component fill its parent horizontally
    - Add prop `wrapperProps` which accepts `React.HTMLAttributes` and is placed to the outermost div of the component
    - All other props (for example `data-testid`) are now placed to the component's interactive `<button>` element instead of the outermost div
    - Adjust the component's visual appearance so that it is in line with `<SingleSelect>` and `<MultiSelect>`

### DIFF
--- a/src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
+++ b/src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
@@ -4,12 +4,35 @@ import { element, input } from '../../theme/reset';
 
 export const baseStyles = (theme: SuomifiTheme) => css`
   width: 290px;
-  /* stylelint-disable no-descending-specificity */
   .fi-dropdown_item-list {
     padding-top: 0;
   }
   &.fi-dropdown {
     display: inline-block;
+
+    &--open {
+      .fi-dropdown_button {
+        border-bottom: 0;
+        border-bottom-left-radius: 0;
+        border-bottom-right-radius: 0;
+        padding-bottom: 8px;
+      }
+    }
+
+    &--error {
+      .fi-dropdown_button {
+        border-color: ${theme.colors.alertBase};
+        border-width: 2px;
+      }
+    }
+
+    &--italicize {
+      .fi-dropdown_button {
+        font-style: italic;
+        color: ${theme.colors.depthDark2};
+        opacity: 1;
+      }
+    }
 
     .fi-dropdown_label--visible {
       margin-bottom: ${theme.spacing.xs};
@@ -23,19 +46,6 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       line-height: 18px;
       &.fi-dropdown_statusText--has-content {
         margin-top: ${theme.spacing.xxs};
-      }
-    }
-
-    &:not(.fi-dropdown--open) {
-      .fi-dropdown_button {
-        &:focus {
-          outline: 0;
-          position: relative;
-
-          &:after {
-            ${theme.focuses.absoluteFocus}
-          }
-        }
       }
     }
 
@@ -118,27 +128,16 @@ export const baseStyles = (theme: SuomifiTheme) => css`
       }
     }
 
-    &--open {
+    &:not(.fi-dropdown--open) {
       .fi-dropdown_button {
-        border-bottom: 0;
-        border-bottom-left-radius: 0;
-        border-bottom-right-radius: 0;
-        padding-bottom: 8px;
-      }
-    }
+        &:focus {
+          outline: 0;
+          position: relative;
 
-    &--error {
-      .fi-dropdown_button {
-        border-color: ${theme.colors.alertBase};
-        border-width: 2px;
-      }
-    }
-
-    &--italicize {
-      .fi-dropdown_button {
-        font-style: italic;
-        color: ${theme.colors.depthDark2};
-        opacity: 1;
+          &:after {
+            ${theme.focuses.absoluteFocus}
+          }
+        }
       }
     }
   }
--- a/src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
+++ b/src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
@@ -19,7 +19,7 @@ export const baseStyles = (theme: SuomifiTheme) => css`
     }
 
     .fi-status-text {
-      line-height: 18px;
+      line-height: 1.1rem;
       &.fi-dropdown_statusText--has-content {
         margin-top: ${theme.spacing.xxs};
       }
--- a/src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
+++ b/src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
@@ -69,7 +69,7 @@ export const baseStyles = (theme: SuomifiTheme) => css`
     .fi-dropdown_display-value {
       width: 100%;
       height: 100%;
-      display: block;
+      display: inline-block;
       line-height: 1.5;
       overflow: hidden;
     }
@@ -144,7 +144,7 @@ export const baseStyles = (theme: SuomifiTheme) => css`
     &:not(.fi-dropdown--open) {
       .fi-dropdown_button {
         &:focus {
-          outline: 0;
+          outline: 3px solid transparent;
           position: relative;
 
           &:after {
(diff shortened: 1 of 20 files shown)


src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
```diff
@@ -1,110 +1,145 @@
 import { css } from 'styled-components';
 import { SuomifiTheme } from '../../theme';
-import { element, inputButton } from '../../theme/reset';
+import { element, input } from '../../theme/reset';
 
 export const baseStyles = (theme: SuomifiTheme) => css`
+  width: 290px;
+  /* stylelint-disable no-descending-specificity */
```
jenkrisu, line 7: Does this need to be disabled or styles moved accordingly?

src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
```diff
@@ -1,110 +1,145 @@
...
-      outline: 0;
-      position: relative;
+    .fi-hint-text {
+      margin-bottom: ${theme.spacing.xs};
+    }
 
-      &:after {
-        ${theme.focuses.absoluteFocus}
+    .fi-status-text {
+      line-height: 18px;
```
jenkrisu, line 23: Should line-height be in rem in case user changes font size in browser settings

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -1,49 +1,79 @@
...
+  getRecursiveChildText,
+} from '../../../utils/common/common';
+import { Popover } from '../../../core/Popover/Popover';
+import { SelectItemList } from '../../Form/Select/BaseSelect/SelectItemList/SelectItemList';
+import { HintText } from '../../Form/HintText/HintText';
+import { StatusText } from '../../Form/StatusText/StatusText';
+import { StatusTextCommonProps } from '../../Form/types';
+
+// Dropdown is a button but it also has a listbox popup
+/* eslint-disable jsx-a11y/role-supports-aria-props */
```
jenkrisu, line 24: Maybe this disable should be placed more specifically to where the rule is disabled, only for that place, if possible

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -1,49 +1,79 @@
...
+  italicize: `${baseClassName}--italicize`,
+};
+
+export interface DropdownProviderState {
+  /** Callback for communicating DropdownItem click to parent  */
+  onItemClick: (itemValue: string) => void;
+  /** Currently selected DropdownItem */
+  selectedDropdownValue: string | undefined | null;
+  /** Currently focused DropdownItem */
+  focusedItemID: string | null | undefined;
```
jenkrisu, line 49: I think this should be focusedItemId, so props are in camelCase

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -258,52 +429,90 @@ class BaseDropdown extends Component<DropdownProps> {
...
-            ref={forwardedRef}
-            name={name}
-            value={selectedValue || ''}
-          >
-            <ListboxButton
-              ref={this.buttonRef}
-              {...dropdownButtonProps}
+          <HintText id={hintTextId}>{hintText}</HintText>
+          <HtmlDiv className={classnames(dropdownClassNames.inputWrapper)}>
+            <HtmlSpan
```
jenkrisu, line 434: Could this element be button, since it has role button? Could also then move part of onKeyDown logic to onClick

src/core/Dropdown/Dropdown/Dropdown.tsx
riitasointi: There are multiple props this this affects so I thought it best to just put at the top of the file

jenkrisu: Is it a bit of "code smell" if some aria props are not compatible element's role?

riitasointi: Ha, making the element a <HtmlButton> and removing role="button" made the linter happy with the props

src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
riitasointi: I couldn't get all the styles working and the linter to be happy at the same time. 

This linter rule has been problematic in some other components as well

src/core/Dropdown/Dropdown/Dropdown.tsx
riitasointi: Fixd

src/core/Dropdown/Dropdown/Dropdown.tsx
riitasointi: Changed the element to be a button. 

onMouseDown was used as a trick so that focus ring does not disappear for a split second. But now it seems onClick works as well 🤷

src/core/Dropdown/Dropdown/Dropdown.baseStyles.tsx
riitasointi: Fixd

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -180,11 +186,241 @@ class BaseDropdown extends Component<DropdownProps> {
...
+      showPopover: false,
+      focusedDescendantId: itemValue,
+    });
+    /**
+     * Timeout is used here to ensure screen reader reads the
+     * currently selected option when focusing back to the button
+     */
+    setTimeout(() => {
+      this.buttonRef.current?.focus();
+    }, 200);
```
jenkrisu, line 247:
```suggestion
  this.setState({ showPopover: false })
    }, 200);
```

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -180,11 +186,241 @@ class BaseDropdown extends Component<DropdownProps> {
...
+    if (!!this.props.onChange) {
+      this.props.onChange(itemValue);
+    }
+    this.setState({
+      selectedValue: itemValue,
+      selectedValueText: BaseDropdown.parseSelectedValueText(
+        itemValue,
+        this.props.children,
+      ),
+      showPopover: false,
```
jenkrisu, line 238:
```suggestion
```

src/core/Dropdown/Dropdown/Dropdown.tsx
jenkrisu: Setting showPopover false after focusing to button should prevent NVDA from starting to read out the document title during the timeout

src/core/Form/DateInput/DatePicker/DatePicker.tsx
jenkrisu, line 1: Can't add suggestion to correct line, but focus trap tabbing is not working here (from last button, user should be able to tab to the first select):

row 219 should be changed to ` const firstElement = yearSelectRef.current;`

src/core/Dropdown/Dropdown/Dropdown.test.tsx
jenkrisu, line 1: Could also add a snapshot for when Dropdown is rendered when it is open.

Looks & works nice!

LJKaski: Snapshot tests changed so that they test all content.

LJKaski (approved):
Seems to be working well! 💪

PR #698, opened 2023-03-21
[Breaking change] wrapperProps typing

jenkrisu:
## Description

PR fixes typing of wrapperProps in DateInput and TextInput components. 

## Motivation and Context

Fix allows users to add data-attributes in wrapperProps.

## How Has This Been Tested?
 
CRA with TypeScript

### Label, SearchInput, TextInput
- **Breaking change:** `wrapperProps` prop typing is changed to include also data-attributes

### DIFF
--- a/src/core/Form/DateInput/DateInput.tsx
+++ b/src/core/Form/DateInput/DateInput.tsx
@@ -14,10 +14,11 @@ import { AutoId } from '../../utils/AutoId/AutoId';
 import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
 import { Debounce } from '../../utils/Debounce/Debounce';
 import { getConditionalAriaProp } from '../../../utils/aria';
+import { getLogger } from '../../../utils/log';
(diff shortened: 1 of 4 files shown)


src/utils/common/common.ts
```diff
@@ -46,3 +46,15 @@ export const forkRefs =
...
   };
+
+/**
+ * The following interface allows data-* attributes.
+ * The basic React.HTMLAttributes interface throws errors when trying to do something like
+ * <Component wrapperProps={{ 'data-testid': 'something' }}
+ * Solution copied from: https://github.com/microsoft/TypeScript/issues/28960
+ */
+type DataAttributeKey = `data-${string}`;
+export interface HTMLAttributesIncludingDataAttributes
```
LJKaski, line 57: Will this now allow all html attributes to the elements of this type regardless of the type of the element in question?

jenkrisu: Updated these same way as in Ilkka's PR

riitasointi (approved):
Helpful change 👍

PR #699, opened 2023-03-24
[Fix] Safari margin issue with CheckboxGroup and RadiobuttonGroup

Riippi:
## Description

Fix for margin issues on CheckboxGroup and RadiobuttonGroup.

Reason for this bug is probably that `<legend>` contains `<div>` (that has `<span>` children) element which is against [html specifications](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend)  "Permitted content | Phrasing content and headings (h1–h6 elements)." and Safari messes up the bottom margin of that inner `span`.
Issue is somewhat isolated in this example https://codepen.io/riippi/pen/xxJbYYP  (try with Safari and compare to other browsers).

Removing that div+span structure is currently not feasible so in this PR the margin is removed from span. Instead it's added to div element in question when hintText is present. 

Solution here is a bit complicated and pure CSS fix would be "ideal", but this was the only way to get it working.

## Related Issue

https://jira.dvv.fi/browse/SFIDS-669

## Motivation and Context

Fixes margin issues in Safari so that the components look same in all browsers

## How Has This Been Tested?
Safari + Styleguidist, Firefox, Chrome

#### Before:
Screenshot 2023-03-24 at 9 49 41
Screenshot 2023-03-24 at 9 49 49
#### After:
Screenshot 2023-03-24 at 9 51 49
Screenshot 2023-03-24 at 9 51 55

### CheckboxGroup
* Fix missing margin of the hintText on Safari
### RadioButtonGroup
* Fix missing margin of the hintText on Safari

### DIFF
--- a/src/core/Form/Checkbox/CheckboxGroup.baseStyles.tsx
+++ b/src/core/Form/Checkbox/CheckboxGroup.baseStyles.tsx
@@ -5,11 +5,15 @@ import { element } from '../../theme/reset';
 export const baseStyles = (theme: SuomifiTheme) => css`
   ${element(theme)}
 
-  & .fi-checkbox-group_label--visible {
+  & .fi-checkbox-group_legend {
     margin-bottom: 10px;
+
+    .fi-hint-text {
+      margin-bottom: 0;
+    }
   }
 
-  & .fi-checkbox-group_legend .fi-hint-text {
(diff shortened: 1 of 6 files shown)


riitasointi (approved):
Seems to do the trick 👍

PR #714, opened 2023-04-27
[Fix] Dropdown screen reader problems

riitasointi:
## Description

PR fixes issues regarding Dropdown usage with screen readers

- Keyboard navigation is implemented by moving actual browser focus() to the list elements
- When selecting an item, a small 10ms timeout is used before focusing back to the dropdown button. This hack seemed to be the only way to make NVDA + Firefox work properly

## How Has This Been Tested?
Mac: Chrome and Safari with VoiceOver
iOS: Safari with VoiceOver
Windows: Chrome, Firefox, Edge with NVDA

### Dropdown
* Fix screen reader problems


jenkrisu (approved):

Riippi (approved):
LGTM

PR #717, opened 2023-05-05
[Fix] Dropdown itemlist adjustments

riitasointi:
## Description

This PR adjusts how Dropdown item list works with mouse and keyboard combo. Previously, mouse hover did not move the actual focus but rather just applied a similar looking style. Now it actually changes the focus(). So using a mouse in the item list is basically the same as navigating through it with a keyboard. This ensures there is only one "active" item in the list at all times (as opposed to the keyboard navigation having its own and another one with a mouse hover)

Also made a small adjustment to `<SelectItemList>` so that the scrolling function is not used when hovering with a mouse.

## Motivation and Context

Prevent multiple "active" looking elements from existing in the Dropdown item list.

## How Has This Been Tested?

MacOS Chrome and Safari VoiceOver
iOS Safari VoiceOver
Windows Chrome Firefox Edge NVDA

### Dropdown
* Fix and issue with multiple "active" looking items appearing in the list. Keyboard and mouse navigation now use the same logic

### DIFF
--- a/src/core/Dropdown/Dropdown/Dropdown.tsx
+++ b/src/core/Dropdown/Dropdown/Dropdown.tsx
@@ -58,6 +58,8 @@ export interface DropdownProviderState {
    * Callback for communicating DropdownItem Tab key press to the parent
    */
   onItemTabPress: () => void;
+  /** Callback for communicating DropdownItem mouse over to parent  */
+  onItemMouseOver: (itemIndex: string) => void;
 }
 
 const defaultProviderValue: DropdownProviderState = {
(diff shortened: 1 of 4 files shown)


Riippi (approved):

src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -58,6 +58,8 @@ export interface DropdownProviderState {
    * Callback for communicating DropdownItem Tab key press to the parent
    */
   onItemTabPress: () => void;
+  /** Callback for communicating DropdownItem mouse over to parent  */
+  onItemMouseOver: (itemIndex: string) => void;
```
jenkrisu, line 62: Should index be type number?

jenkrisu (approved):
Looks good!

src/core/Dropdown/Dropdown/Dropdown.tsx
riitasointi: The variable was actually badly named, it's not the item's index but rather the string value which is the parameter here. Changed it from `itemIndex ` to `itemValue`

jenkrisu (approved):

PR #722, opened 2023-05-11
[Feature] LanguageMenu without Reach

riitasointi:
## Description

This PR completely rewrites the `<LanguageMenu>` component without using ReachUI dependency. The code is now very similar to `<ActionMenu>` as they are practically the same component with visual differences. 

## Motivation and Context

We wanted to get rid of the ReachUI dependency and make LanguageMenu work with React 18

## How Has This Been Tested?
Windows: Chrome, Edge, Firefox (& NVDA), high contrast mode
iOS: Chrome, Safari (& VoiceOver)
Mac: (Browserstack) Chrome, Safari (& VoiceOver)
Android: Chrome + Talkback

### General

- **Breaking change**: Library no longer exports main.css file. Import line for main.css should be removed
  - Main.css file contained styles for Reach-dependency that is now removed   

### LanguageMenu
- **Breaking change**: Completely rewrite the component without using ReachUI dependency
    - Rename prop `name` to `buttonText` and change its type from `any` to `string`. This name better describes its purpose
    - Add mandatory prop `aria-label` to inform assistive technologies about usage of LanguageMenu
    - Remove props `languageMenuButtonClassName` and `languageMenuOpenButtonClassName`
    - Add mandatory `lang` prop to `<LanguageMenuItem>`
    - Remove child element `<LanguageMenuLink>`. All children should now be `<LanguageMenuItem>`s. `<LanguageMenuItem>` can be given either an href or an onClick() prop and it will be rendered as a link or a button based on that
    - Add props `onBlur()`, `onClick()`, `onOpen()` and `onClose()`
    - Add prop `wrapperProps` which accepts React.HTMLAttributes and is placed to the outermost div of the component
    - Introduce high contrast mode improvements

### Misc
- Remove ReachUI dependency from the project


package.json
```diff
@@ -155,9 +154,6 @@
   "dependencies": {
     "@babel/runtime": "7.18.9",
     "@popperjs/core": "2.11.5",
-    "@reach/listbox": "0.16.2",
```
jenkrisu, line 158: 🥳

src/core/LanguageMenu/LanguageMenu.md
```diff
@@ -0,0 +1,21 @@
+```js
+import {
+  LanguageMenu,
+  LanguageMenuItem
+} from 'suomifi-ui-components';
+
+<LanguageMenu
+  buttonText="In English (EN)"
```
jenkrisu, line 8: I would change this default to be Suomeksi (FI), since that selection only logs the value to console. When clicking it, the language menu seems a bit buggy since "nothing happens".

riitasointi: This point is not really applicable anymore since all the items only accept programmatic navigation through onSelect() now

src/core/LanguageMenu/LanguageMenu.md
```diff
@@ -0,0 +1,31 @@
...
+  aria-label="Change language. Current language is English"
```
riitasointi, line 9: `aria-selected` or `aria-current` do not seem to work with role="menu" and "menuitem". So I thought maybe it's enough that the current language is indicated when arriving to the menu toggle button? Users with vision also see this information so screen reader users should too

jenkrisu: This does move the responsibility to announce current language to the library user. Maybe more realistic example would be that the  value is same that it is in buttonText, so:
ria-label="Change language. Current language In English (EN)"

jenkrisu: Documentation should also direct clearly to use aria-label to also announce the current selected language. 🤔

src/core/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,217 @@
...
+  } = props;
+
+  const openButtonRef = forwardedRef || useRef<HTMLButtonElement>(null);
+  const [menuVisible, setMenuVisible] = useState<boolean>(false);
+  const [ariaExpanded, setAriaExpanded] = useState<boolean>(false);
+  const [selectFirstItem, setSelectFirstItem] =
+    useState<InitialActiveDescendant>('none');
+
+  const menuId = `${id}-menu`;
+  const buttonId = `${id}`;
```
jenkrisu, line 90: Should these ids be more unique?

jenkrisu: Realized that this id will be unique from AutoId, so no need to change this.

LJKaski (approved):

PR #730, opened 2023-05-26
[Fix]: Dropdown selected item parsing

riitasointi:
## Description

PR removes problematic code which tried to parse only the **children text content** of a selected `<DropdownItem>` into the Dropdown button element. 

This logic created problems with elements like `<FormattedMessage />` (from react-intl) and patterns like this

```
const Foo = () => 'This text will not show up in button when selected'
...
<DropdownItem>
   <Foo />
</DropdownItem>
```

PR also raises version to 10.0.3 for a hotfix release

## Motivation and Context

This bug was reported by a user and severely affects their application

## How Has This Been Tested?
Styleguidist, CRA with react-intl `<FormattedMessage />`

image (8)

☝️ Initial bug report image

### Dropdown
* Fix a bug where selected item text did not correctly appear in Dropdown button

### DIFF
--- a/src/core/Dropdown/Dropdown/Dropdown.tsx
+++ b/src/core/Dropdown/Dropdown/Dropdown.tsx
@@ -12,7 +12,6 @@ import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
 import {
   forkRefs,
   getOwnerDocument,
-  getRecursiveChildText,
   HTMLAttributesIncludingDataAttributes,
 } from '../../../utils/common/common';
 import { Popover } from '../../../core/Popover/Popover';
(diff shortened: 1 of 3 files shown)


src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -222,18 +221,18 @@ class BaseDropdown extends Component<DropdownProps> {
       | Array<ReactElement<DropdownItemProps>>
       | ReactElement<DropdownItemProps>
       | undefined,
-  ): string | undefined {
+  ): ReactNode | undefined {
```
jenkrisu, line 224: This function could be renamed getSelectedValueNode and selectedValueText => selectedValueNode

jenkrisu:
Tested also in Styleguidist with NVDA (Firefox, Edge, Chrome). ✔️

jenkrisu (approved):

PR #738, opened 2023-05-30
[Fix] Support nested array children in Dropdown

jenkrisu:
## Description

PR adds support to giving nested array children in Dropdown.

```
<Dropdown ...>
  {['Item 1', 'Item 2', 'Item 3'].map((item) => (
    <DropdownItem value={`dropdown-item-${item}`}>
      {item}
    </DropdownItem>
  ))}
  <DropdownItem value="Static">Static</DropdownItem>
</Dropdown>
```

PR removes old NVDA fix that was needed for v10.0.0-beta.1 but not as of v10.0.0.

## Motivation and Context

Giving a nested array as children was supported by previous Dropdown version and is a common pattern.

Old fix for NVDA with` <VisuallyHidden> `text causes confusion with mobile screen readers, that can navigate to the hidden text.

## How Has This Been Tested?

Windows: Firefox, Chrome, Edge (NVDA)
iOS: Chrome, Safari (VoiceOver)
Android: Chrome (TalkBack)
Mac: Chrome, Safari (VoiceOver)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "suomifi-ui-components",
-  "version": "10.0.3",
+  "version": "10.0.4",
   "description": "Suomi.fi UI component library",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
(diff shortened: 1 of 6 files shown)


riitasointi (approved):
LGTM

PR #749, opened 2023-06-21
[Bugfix] LinkList ariaDescribedBy prop handling

LJKaski:
## Description
`ariaDescribedBy` prop accidentally went unhandled in the LinkList component. This PR fixes it so that it's implemented in the component properly.

## Motivation and Context
Previously the prop simply passed to the DOM without any handling causing an error to show up in console.

## How Has This Been Tested?
Tested locally in styleguidist on Chrome. The prop is now handled correctly.

Before: 
image

After: 
image


jenkrisu (approved):
Looks good 👍

PR #755, opened 2023-07-18
[Breaking change] Add spacing props to components

riitasointi:
## Description

PR refactors implementation of spacing in component  `<Block>` and  `<Paragraph>`. PR adds spacing props to most other components.

These components were left out (* = not added to keep margin styles according to design in component's internal design):
- ActionMenuItem*
- ActionMenuDivider*
- BreadCrumbLink*
- DropdownItem*
- ExpanderTitleButton*
- ExpanderContent*
- ExpanderTitle*
- LanguageMenuItem*
- LinkListItem*
- Modal - positioned to center of window
- ModalContent*
- ModalTitle*
- ModalFooter*
- Paragraph - has marginBottomSpacing
- ServiceNavigationItem*
- SideNavigationItem*
- SkipLink - absolutely positioned and hidden by default
- VisuallyHidden - hidden from view

## Motivation and Context

The goal was to make the changes to components minimal and hide all the ugly logic in a helper file `src/core/theme/utils/spacing.ts`

This PR does not contain tests and documentation, those will be implemented in other PR:s to keep PR size reasonable.

### Block
- **Breaking change:** changed implementation of margins and spacing from CSS-based styles to style-attribute

### Paragraph
- **Breaking change:** changed implementation of margin  from CSS-based styles to style-attribute
- **Breaking change:** `marginBottomSpacing` prop is replaced by `mb`
- New props `margin`, `mx`, `my`, `mt`, `mr`, `mb`, `ml`

### General
- New props `margin`, `mx`, `my`, `mt`, `mr`, `mb`, `ml` to following components
    - ActionMenu
    - Alert
    - BreadCrumb
    - Button
    - Checkbox
    - CheckboxGroup
    - Chip
    - DateInput
    - Dropdown
    - Expander
    - ExpanderGroup
    - ExternalLink
    - Heading
    - HintText
    - InlineAlert
    - Label
    - LanguageMenu
    - Link
    - LinkList
    - LoadingSpinner
    - MultiSelect
    - Notification
    - Pagination
    - RadioButton
    - RadioButtonGroup
    - RouterLink
    - SearchInput
    - ServiceNavigation
    - SideNavigation
    - SingleSelect
    - StaticChip
    - StatusText
    - Text
    - Textarea
    - TextInput
    - Toast
    - ToggleButton
    - ToggleInput
    - Tooltip
    - WizardNavigation


aappoalander:
Maybe you could reconsider leaving out spacing props from Paragraph. marginBottomSpacing works if your strategy is to define bottom margins, but it might as well be defining top margins for everything. In addition, there might be, and probably are, use cases where different kinds of layouts might need any of the spacing props. There should not be any significant negative effects for allowing spacing props for Paragraph as well?

jenkrisu:
> Maybe you could reconsider leaving out spacing props from Paragraph. marginBottomSpacing works if your strategy is to define bottom margins, but it might as well be defining top margins for everything. In addition, there might be, and probably are, use cases where different kinds of layouts might need any of the spacing props. There should not be any significant negative effects for allowing spacing props for Paragraph as well?

Good point @aappoalander - it would also be beneficial to unify how margin is used, so that this one component (Paragraph) does not have different approach to margins. Will add this to PR.

styleguide.config.js
```diff
@@ -6,6 +6,9 @@ const propFilter = (prop) => {
   if (prop.description.length === 0) {
     return false;
   }
+  // if (['margin', 'mx', 'my', 'mt', 'mr', 'mb', 'ml'].includes(prop.name)) {
```
Riippi, line 9: Is this needed for something?

Riippi (changes requested):

styleguide.config.js
jenkrisu: It will hide the margin props from props list in the next PR

jenkrisu: Left here commented out so it was easier to see/test which components had the margin props support added

Riippi (approved):

tlehtimaki:
Out of curiosity, I would have asked that isn't it usually better to have margin applied via parent (or as high in the tree as possible) component instead of setting it straight in the component. However if needed to set individually, it makes sense to do it like this. :)

PR #761, opened 2023-08-10
[Fix]: DateInput ref

riitasointi:
## Description

There was a bug in `<DateInput>` ref. User given ref was never actually set to the input, because there was an internal ref taking its place.

The implementation now uses our custom `forkRefs` utility to set both refs to the input.

## How Has This Been Tested?

Styleguidist, running Jest tests.

### DateInput
* Fix ref

### DIFF
--- a/src/core/Form/DateInput/DateInput.tsx
+++ b/src/core/Form/DateInput/DateInput.tsx
@@ -16,7 +16,10 @@ import { SuomifiThemeProp, SuomifiThemeConsumer } from '../../theme';
 import { Debounce } from '../../utils/Debounce/Debounce';
 import { getConditionalAriaProp } from '../../../utils/aria';
 import { getLogger } from '../../../utils/log';
-import { HTMLAttributesIncludingDataAttributes } from '../../../utils/common/common';
+import {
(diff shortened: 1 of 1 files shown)


src/core/Form/DateInput/DateInput.tsx
```diff
@@ -245,6 +248,9 @@ const BaseDateInput = (props: DateInputProps) => {
     datePickerDefaultTexts.fi,
   );
 
+  // Remove the possibility to have undefined forwardedRef as a parameter for forkRefs
+  const definedRef = forwardedRef || null;
```
jenkrisu, line 252: Maybe forwardedRef could be set in props to default null to handle this also?

riitasointi: Probably would work, yes. I just copied this line from other components where `forkRefs` is used

jenkrisu: Ok it seems a bit verbose but it does make it more clear why type is being changes here. 👌

jenkrisu (approved):

PR #771, opened 2023-09-04
[Fix] Breadcrumb underline

riitasointi:
## Description

PR makes the `<BreadcrumbLink>` component follow the `underline` prop, same as all other link components. 

The breadcrumb link did not have any underlining previously. 

## Motivation and Context

`<BreadcrumbLink>` should work with the `underline` prop since it is part of the component's API (via BaseLinkProps)

## How Has This Been Tested?

Styleguidist

### Breadcrumb
- Fix `underline` prop behavior


jenkrisu (approved):
Good catch 👍

PR #773, opened 2023-09-06
[Fix] SingleSelect to update when items array is changed

Riippi:
## Description

Currently if the selected object in SingleSelect is updated elsewhere so that the label value is changed, the new value is updated to the popover list, but not in the input.
This PR fixes it so that the input value is also updated.

## Related Issue

https://jira.dvv.fi/browse/SFIDS-852

## Motivation and Context

This issue came from a project using SingleSelect in a form where some other input changes the item array.

## How Has This Been Tested?
Styleguidist. New unit test. 

There is a branch `fix/singleselect-modify-selected-test`  that has an styleguidist example demoing this fix. It's identical to this PR except the .md file.

### SingleSelect
- Fix input to show updated value when item array is changed


jenkrisu (approved):
Nice work! 👍

PR #783, opened 2023-10-06
[Fix] LanguageMenu popover width

riitasointi:
## Description

This PR makes the LanguageMenu component's widths behave as thus
- The menu button will have a dynamic width based on the text in the button (`buttonText`) prop
- The popover will always have the width of the widest item in the list. Previously, the popover used to be the same width as the menu button

## Motivation and Context

It was reported to us by users that the current behavior of the component is not optimal

## How Has This Been Tested?

Styleguidist

## Screenshots
Before: 
image

After:
image

### LanguageMenu
- Change the popover width to match the width of the widest element in the list


LJKaski (approved):
Seems to do the trick. Well done :)

PR #786, opened 2023-10-13
[Fix] brandBase color docs

riitasointi:
## Description

This PR hardcodes the brandBase color hex value as #003479 in colors documentation. Due to HSL conversion, it has been previously shown as #00357a.

This same thing has already been implemented to DS-site. 

## Motivation and Context

In the beginning of time, the brand blue color of DVV was determined to be hex #003479. This color as HSL is 214.21, 100%, 23.73% (according to MacOS Chrome)

Suomifi-tokens uses HSL values for colors and when the brand color was added there it was rounded to 214, 100%, 24%

Now when this HSL is turned back to hex it's not #003479 anymore but rather #00357a. This has caused some confusion and we figured it's just easiest to display the original hex value. 

## Release notes

None needed? Silent change

### DIFF
--- a/src/docs/Colors/Colors.tsx
+++ b/src/docs/Colors/Colors.tsx
@@ -122,6 +122,10 @@ const StyledFigure = styled.figure`
 
 const ColorFigure = (props: ColorProps & SuomifiThemeProp) => {
   const { color, keyName } = props;
+  // Brand color is hardcoded in these docs due to HSL conversion producing
+  // a hex which differs slightly from the value DVV defined as
+  // their brand color initially. It is reasonable to display the original value
+  // #003479 to avoid confusion.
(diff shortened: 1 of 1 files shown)


src/docs/Colors/Colors.tsx
```diff
@@ -118,7 +118,7 @@ const StyledFigure = styled.figure`
 
 const ColorFigure = (props: ColorProps & SuomifiThemeProp) => {
   const { color, keyName } = props;
-  const hslaAsHex = hslaToHex(color);
+  const hslaAsHex = keyName !== 'brandBase' ? hslaToHex(color) : '#003479';
```
LJKaski, line 121: Should this be commented for future generations? It might be raising some questions 😅

riitasointi: Added a comment for this

LJKaski (approved):

PR #788, opened 2023-10-19
[Fix] Style attribute support

LJKaski:
## Description
This PR unifies the use of style prop in the components, extends some component properties with HTML properties, and gets rid of `wrapperProps` props.

The purpose of the change is to unify usage of styling, improve support for `data-testid` type non-javascript attributes and overall make the DX more coherent regarding these.

## Motivation and Context
Developers have reported problems with data-attributes and styling. These were entwined so it made sense to tackle both at the same time.

## How Has This Been Tested?
Tested so far in styleguidist on chrome. Needs to be tested in different environments to check funcion, styling, and typing.

### Dropdown, 
- Remove `wrapperProps` and forward style prop to the outer wrapper.


riitasointi (approved):
This makes things more uniform across the board 👍 Well done

PR #791, opened 2023-10-24
[Fix] Expander and WizardNavigation example accessibility

danielck:
## Description and Release notes

- Change all "expanders" to "sections" in ariaOpenAllText and ariaCloseAllText in Expander example code. 
- Use the same text in `aria-label` as in `heading` in WizardNavigation. 

## Motivation and Context

Improves accessibility when example code is used as-is. However, in the case of the WizardNavigation component, ideally there should be no need for an aria-label at all, unless it is desired to have a different text from the one visually shown.

### DIFF
--- a/src/core/Expander/Expander/Expander.md
+++ b/src/core/Expander/Expander/Expander.md
@@ -130,9 +130,9 @@ import {
 
 <ExpanderGroup
   openAllText="Open all"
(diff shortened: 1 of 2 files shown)


LJKaski (approved):
These look like good clarifications to me 👍

PR #794, opened 2023-10-30
[Fix] TimeInput: Remove wrapperProps in favor of style

riitasointi:
## Description

This PR removes `wrapperProps` from TimeInput and places the generic `style` prop to the outer div element

## Motivation and Context

Other components in the library already follow this pattern

## How Has This Been Tested?

Unit test and Styleguidist

### TimeInput
- Remove `wrapperProps` and place the generic `style` attribute to the outermost div element of the component


LJKaski (approved):
Looks good! 👍

PR #802, opened 2023-11-16
[Feature] Add generics to Dropdown

ketsappi:
## Description

Add generics to Dropdown (+ DropdownItem).
But it should not allow just any value, but value that extends String. 

This should be valid:
```tsx
type CountryKey = 'finland' | 'sweden' | 'norway' | 'denmark' | 'iceland';
```

This should not be valid (e.g number):
```tsx
type CountryKey =  1 | 2 | 3 | 4 | 5;
```

## Motivation and Context

Would ease the component usage as the given type safety.

## How Has This Been Tested?
Locally in the project and then in new next.js app started from scratch.

### With the changes
SCR-20231116-imui

### Without the change (suomifi-ui-components: 12.0.2)
SCR-20231116-inwg

To get rid of that error, we would need to cast it which is not nice:
```tsx
onChange={(newValue) => dropdownOnChange(newValue as CountryKey)}
```

## Example code

```tsx
import { Dropdown, DropdownItem } from 'suomifi-ui-components';

// ...

type CountryKey = 'finland' | 'sweden' | 'norway' | 'denmark' | 'iceland';

interface CountryItem {
  name: string;
  key: CountryKey;
};

const countries: CountryItem[] = [
  {
    name: 'Finland',
    key: 'finland'
  },
  {
    name: 'Sweden',
    key: 'sweden'
  },
  {
    name: 'Norway',
    key: 'norway'
  },
  {
    name: 'Denmark',
    key: 'denmark'
  },
  {
    name: 'Iceland',
    key: 'iceland'
  }
];

// ...

const [selectedCountry, setSelectedCountry] = React.useState<CountryKey>();

const dropdownOnChange = (newValue: CountryKey) => {
  setSelectedCountry(newValue);
};

// ...

<Dropdown
  labelText="Country"
  hintText="Select your current country of residence"
  visualPlaceholder="Choose country"
  value={selectedCountry}
  onChange={(newValue) => dropdownOnChange(newValue)}
>
  {countries.map((country) => (
    <DropdownItem key={country.key} value={country.key}
    >
      {country.name}
    </DropdownItem>
  ))}
</Dropdown>
```

## Links
Interesting article which I read before this>
https://gertrude.app/blog/generic-react-components-typescript

## Release notes

- Add generic support for Dropdown (+DropdownItem)

### DIFF
--- a/src/core/Dropdown/Dropdown/Dropdown.tsx
+++ b/src/core/Dropdown/Dropdown/Dropdown.tsx
@@ -1,4 +1,4 @@
-import React, { Component, ReactNode, ReactElement, forwardRef } from 'react';
+import React, { Component, ReactNode, ReactElement } from 'react';
 import { default as styled } from 'styled-components';
 import classnames from 'classnames';
 import { getConditionalAriaProp } from '../../../utils/aria';
@@ -46,13 +46,13 @@ export const dropdownClassNames = {
   fullWidth: `${baseClassName}--full-width`,
 };
 
-export interface DropdownProviderState {
+export interface DropdownProviderState<T extends string = string> {
   /** Callback for communicating DropdownItem click to parent  */
-  onItemClick: (itemValue: string) => void;
+  onItemClick: (itemValue: T) => void;
   /** Currently selected DropdownItem */
-  selectedDropdownValue: string | undefined | null;
+  selectedDropdownValue: T | undefined | null;
   /** Currently focused DropdownItem */
-  focusedItemValue: string | null | undefined;
+  focusedItemValue: T | null | undefined;
   /** ID of the Dropdown component.
    * Used in DropdownItem to create a derived ID for each item
    */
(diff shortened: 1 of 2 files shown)


src/core/Dropdown/Dropdown/Dropdown.tsx
```diff
@@ -664,34 +668,43 @@ class BaseDropdown extends Component<DropdownProps> {
...
+export const Dropdown = React.forwardRef(DropdownInner) as <
+  T extends string = string,
+>(
+  props: DropdownProps<T>,
+  ref: React.RefObject<HTMLButtonElement>,
+) => ReturnType<typeof DropdownInner>;
 
-Dropdown.displayName = 'Dropdown';
-export { Dropdown, DropdownProvider, DropdownConsumer };
+DropdownInner.displayName = 'Dropdown';
```
ketsappi, line 709: Needed to do this to get it working also in the styleguidist; without this it could not show the props in page.

LJKaski: Might be a good idea to add this as a comment in the code as well just to clarify. Would also come in handy when we encounter similar cases elsewhere.

ketsappi: Comment is now added.

LJKaski (approved):
Does what it says. Thanks for the contribution and good work! 👍

PR #803, opened 2023-11-20
[Feature] Add export for characterCounterProps

ketsappi:
## Description

Add export for `characterCounterProps` that is used with `TextInput` & `TextArea` components.

## Motivation and Context

Is the user of the library needs to for e.g alter the `TextInputProps` like following, it would then require to implement the props from the `characterCounterProps`.

Reproducing the issue without the fix:
```tsx
import { TextInput, TextInputProps } from 'suomifi-ui-components';

interface EnhancedTextInputProps extends Omit<TextInputProps, 'value'> {
  something: string;
}

const EnhancedTextInput = (props:EnhancedTextInputProps) => {
  const {something, ...restOfTheProps } = props;

  return (
      <TextInput {...restOfTheProps} />
      )
};
```

The error with this example is something like:

```
Types of property 'ariaCharactersRemainingText' are incompatible.
  Type '((amount: number) => string) | undefined' is not assignable to type '(amount: number) => string'.
    Type 'undefined' is not assignable to type '(amount: number) => string'.
```

### After the changes of this PR

After exporting the `characterCounterProps` out of the library this case can be solved like the following:

```tsx
const EnhancedTextInput = (props:EnhancedTextInputProps & characterCounterProps) => {
  // ..
}
```

## How Has This Been Tested?

Tested locally in the project and in fresh new Next.js project.

## Release notes

- Add export for `characterCounterProps`

### DIFF
--- a/src/core/Form/TextInput/TextInput.tsx
+++ b/src/core/Form/TextInput/TextInput.tsx
@@ -25,7 +25,11 @@ import { VisuallyHidden } from '../../VisuallyHidden/VisuallyHidden';
 import { Label, LabelMode } from '../Label/Label';
 import { StatusText } from '../StatusText/StatusText';
 import { HintText } from '../HintText/HintText';
-import { InputStatus, StatusTextCommonProps } from '../types';
+import {
+  characterCounterProps,
+  InputStatus,
+  StatusTextCommonProps,
+} from '../types';
 import { baseStyles } from './TextInput.baseStyles';
 
 const baseClassName = 'fi-text-input';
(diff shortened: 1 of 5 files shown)


src/core/Form/types.ts
```diff
@@ -13,3 +13,24 @@ export interface StatusTextCommonProps {
    */
   statusTextAriaLiveMode?: AriaLiveMode;
 }
+
```
ketsappi, line 16: Just a note that would it be useful to also export other types out from the library?

src/core/Form/TextInput/TextInput.tsx
```diff
@@ -124,7 +107,7 @@ interface BaseTextInputProps
 
 export type TextInputProps = characterCounterProps & BaseTextInputProps;
 
-const BaseTextInput = (props: characterCounterProps & TextInputProps) => {
```
ketsappi, line 127: Removed the double use of `characterCounterProps` here.

LJKaski (approved):
Looks like a good enhancement.

src/core/Form/types.ts
LJKaski: Certainly worth considering 👍

PR #807, opened 2023-12-01
[Fix] Button support for icon only implementation

LJKaski:
## Description
This PR improves support for Buttons with only an icon and no text content out of the box. Some changes:
- Icon wrapper elements are no longer rendered if corresponding icon is not specified
- Buttons are square with equal proportions when no text content is given
- Buttons require either children or an `aria-label` attribute
- Button component's `icon` and `iconRight` prop types changed from `ReactNode` to `ReactElement`
- `ActionMenuProps` changed from an interface to a type to accommodate changes

## Motivation and Context
This was requested by users and the redundant wrappers were noticed by the team.

## How Has This Been Tested?
Tested by running locally on styleguidist and by trying out the new typing inside the project. Needs to be tested in another project to check the typing just in case.

### Button
**Breaking change:** Adjusted paddings to accommodate use with only icon and no content.
**Breaking change:** ButtonProps now require either `children` or `aria-label`
### ActionMenu
**Breaking change:** Remove redundant custom styling via "icon only" class.

### DIFF
--- a/src/core/ActionMenu/ActionMenu.baseStyles.tsx
+++ b/src/core/ActionMenu/ActionMenu.baseStyles.tsx
@@ -7,15 +7,7 @@ export const baseStyles = (theme: SuomifiTheme) => css`
   }
 
   & .fi-action-menu_button {
-    &--icon-only {
-      & .fi-button_icon--right {
-        & .fi-icon {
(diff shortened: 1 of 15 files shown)


riitasointi (approved):
Tested in a separate project and seems to work well!

PR #814, opened 2023-12-14
[Feature] Notification close button to optional

LJKaski:
## Description
This PR makes the presence of the close button in Notification optional. This is controlled via the `showCloseButton` prop. the `closeText` prop is now required depending on the value of the `showCloseButton` prop. The new union type is also exported to help mitigate any issues that may rise due to the typing.

## Motivation and Context
This feature was requested by a service, discussed among the DS team and deemed worth implementing.

## How Has This Been Tested?
By testing the typing inside the project and checking that the outcome is desired in styleguidist on Chrome.

### Notification
- Make close button optional via `showCloseButton` prop.


riitasointi (approved):
LGTM!

PR #816, opened 2023-12-18
[Fix] DateInput aria-selected

LJKaski:
## Description
This PR adds `aria-selected=true` for the selected date in DateInput's date picker. It also addresses an issue where a redundant `aria-disabled=false` attribute would be present in DOM for most day selectors in date picker.

## Motivation and Context
This issue came up in discussions and is an accessibility oversight that needed to be fixed.

## How Has This Been Tested?
Tested in styleguidist on chrome with NVDA and TalkBack and by checking the DOM.

### DateInput
- Add `aria-selected="true"` and `role="option"` for the selected date in date picker
- Remove redundant aria-attributes from DOM


danielck:
Unfortunately role="option" can only be used inside a `listbox` or `group`. After reviewing a few other implementations and options, semantically it seems the best bet is to do the following:

- Set the `table`to role="grid". This will allow us to use aria-selected later. A grid must have an accessible name, so it needs an aria-label. This can be simply the month + year.
- Set role="gridcell" on individual `td`elements, however this might not strictly be necessary according to the spec. 
- Set aria-selected="true" on the wrapping td, **not the button**.
- Because there seems to be crappy support for everything mentioned above, I suggest adding "selected" to the aria label of the selected date.

riitasointi (approved):

PR #817, opened 2023-12-21
[Fix] Button height

LJKaski:
## Description
This PR fixes the height in buttons with borders. The padding was slightly too big, so the height of the button was 42px instead of the intended 40px by default when borders were present.

## Motivation and Context
This was a clear bug and needed to be fixed

## How Has This Been Tested?
By checking the styles in styleguidist on chrome.

### Button, Actionmenu
- **Breaking change:** Decrease button top and bottom padding by 1px to fix wrong total height. Increase horizontal padding by total of 2px in borderless buttons.


riitasointi (dismissed):
Good fix!

riitasointi (approved):

PR #822, opened 2024-01-16
[Fix] LinkList aria label

LJKaski:
## Description
This PR changes the `LinkList` component to use `aria-labelledby` instead of `aria-describedby` to connect the list heading to the list itself.

## Motivation and Context
This is the correct aria prop to use in this case and may improve the behavior on some devices and screen readers.

## How Has This Been Tested?
Tested on NVDA on Chrome and TalkBack on Android.

### LinkList
- **Breaking change:** Change `ariaDescribedBy` prop to `ariaLabelledBy`.


riitasointi (approved):
Looks good, this change makes sense 👍

PR #824, opened 2024-01-21
[Feature] SingleSelect & MultiSelect: Add onChangeWithoutDebounce()

riitasointi:
## Description

This PR adds a new prop called `onChangeWithoutDebounce()` to `<SingleSelect>` & `<MultiSelect>`. The prop is basically the same as `onChange()` but does not use the debounce time even if it is specified. 

Its inteded use case is mainly enabling the `loading` state of the component. It makes sense to show the loading spinner immediately when the user starts filtering and only make a backend call after a certain debounce time. 

In addition, this PR also fixes an oversight in the SingleSelect component `getDerivedStateFromProps()` logic. Previously, when the value of the `items` prop or the `selectedItem` prop changed, the filter input's value was always set to an empty string `''`. Now it only does that if `selectedItem` is set to null. 

## Motivation and Context

It was reported by users that having the `loading` state of the component only be applied after the debounce time is not ideal. It causes the existing items to be filtered briefly by the new input before the loading process begins. 

## How Has This Been Tested?

Styleguidist on Chrome & library's programmatic tests

### SingleSelect
- Fix filter input value parsing when `items` or `selectedItem` prop changes
- Add `onChangeWithoutDebounce()` prop. Inteded to be used to set the `loading`prop

### MultiSelect
- Add `onChangeWithoutDebounce()` prop. Inteded to be used to set the `loading`prop

### DIFF
--- a/src/core/Form/Select/SingleSelect/SingleSelect.tsx
+++ b/src/core/Form/Select/SingleSelect/SingleSelect.tsx
@@ -243,14 +243,20 @@ class BaseSingleSelect<T> extends Component<
               (item) =>
                 item.uniqueItemId === prevState.selectedItem?.uniqueItemId,
             );
-      const resolvedInputValue = selectedItemChanged
-        ? selectedItem !== null
-          ? selectedItem?.labelText || prevState.filterInputValue
-          : ''
-        : propItems.find(
-            (item) =>
-              item.uniqueItemId === prevState.selectedItem?.uniqueItemId,
-          )?.labelText || prevState.filterInputValue;
+      let resolvedInputValue = prevState.filterInputValue;
+      if (selectedItemChanged) {
+        resolvedInputValue = selectedItem
+          ? selectedItem.labelText || prevState.filterInputValue
+          : '';
+      } else {
+        const matchingItem = propItems.find(
+          (item) => item.uniqueItemId === prevState.selectedItem?.uniqueItemId,
+        );
+        if (matchingItem) {
+          resolvedInputValue =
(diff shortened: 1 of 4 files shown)


src/core/Form/Select/SingleSelect/SingleSelect.tsx
```diff
@@ -239,11 +244,13 @@ class BaseSingleSelect<T> extends Component<
                 item.uniqueItemId === prevState.selectedItem?.uniqueItemId,
             );
       const resolvedInputValue = selectedItemChanged
-        ? selectedItem?.labelText || ''
+        ? selectedItem !== null
+          ? selectedItem?.labelText || prevState.filterInputValue
+          : ''
```
riitasointi, line 249: Just a note: This starts to be a little difficult to read with the nested if-logic

LJKaski: Might be a good idea to refactor into a slightly more readable form.

One potential solution (still not super clear thought):

```js
      let resolvedInputValue = prevState.filterInputValue;
      if (selectedItemChanged) {
        resolvedInputValue = selectedItem
          ? selectedItem.labelText || prevState.filterInputValue
          : '';
      } else {
        const matchingItem = propItems.find(
          (item) => item.uniqueItemId === prevState.selectedItem?.uniqueItemId,
        );
        if (matchingItem) {
          resolvedInputValue =
            matchingItem.labelText || prevState.filterInputValue;
        }
      }
```

LJKaski (approved):

PR #825, opened 2024-02-02
[Feature] ErrorSummary

riitasointi:
## Description

PR introduces the `<ErrorSummary>` component. It is intended to be used to display form validation errors in a centralized manner, especially in static form validation cases. See `ErrorSummary.md` for more detailed instructions. 

## How Has This Been Tested?

In Styleguidist on macOS Chrome & Safari + VoiceOver. More testing and accessibility validation needed.

## Screenshots

image

### ErrorSummary
- Introduce the `<ErrorSummary>` component


src/core/Form/ErrorSummary/ErrorSummary.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+interface BasicErrorSummaryItemProps {
+  /** Visible text of the error item */
+  text: string;
+}
+
+type InputReferenceProps =
+  | {
+      /** HTML id of an input element. The error corresponds to this input. Provide either this or `inputRef`. */
+      inputId: string;
```
LJKaski, line 35: Should the naming a descriptions of these be changed in to something more generic to avoid confusion. Would "field" as a word be a more generic refence to where the ref should point? Now the naming and descriptions seem pretty strict even though the typing isn't

src/core/Form/ErrorSummary/ErrorSummary.tsx
```diff
@@ -0,0 +1,178 @@
...
+      const element = document.getElementById(errorItem.inputId);
+      if (element) {
+        element.focus();
+      }
+    }
+  };
+
+  return (
+    <HtmlDivWithRef
+      as="section"
```
LJKaski, line 102: What does the `as="section"` accomplish here? It seems to make it a landmark <section> element in the DOM, but there's no `aria-label` or `aria-labelledby` attached to it.

LJKaski: Well, technically the aria-props from `passProps` go here, which gives it some use.

src/core/Form/ErrorSummary/ErrorSummary.tsx
riitasointi: I think we want to enforce the primary use case of setting focus to inputs. That's why I wouldn't change the prop naming. 

And the comments are actually for internal use only, due to Styleguidist limitations that particular type is not displayed in the documentation page.

LJKaski: All good here then :) Let's keep then as they are 👍

src/core/Form/ErrorSummary/ErrorSummary.tsx
LJKaski: Let's leave it as is for the moment and maybe see if linking the heading to the section could be beneficial in the future.

LJKaski (approved):
Tested in different projects and on mobile. Seems to be working really well. Great work! 👍

PR #827, opened 2024-02-16
[Fix] Unintended typing change in callback function props

LJKaski:
## Description
This PR fixes some overlap in typing related to callbacks such as `onChange`, `onClick` and `onBlur`. After the recent change of extending component props with relevant HTML props, the typings were accidentally changed in components where there were differing typings for the callbacks in the component APIs and the HTML props they extended.

The issue was fixed by omitting the overlapping props from the extended HTML props.

## Motivation and Context
There was an unintended change in typings which caused issues for developers.

## How Has This Been Tested?
Tested in a React+Vite test project with latest Typescript to see if the issues were indeed fixed.

## Release notes
- Fix unintentionally changed typings in `onChange`, `onClick` and `onBlur` props in several components

### DIFF
--- a/src/core/Form/DateInput/DateInput.tsx
+++ b/src/core/Form/DateInput/DateInput.tsx
@@ -158,7 +158,7 @@ export interface DateInputProps
   extends DatePickerProps,
     MarginProps,
     StatusTextCommonProps,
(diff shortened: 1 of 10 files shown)


LJKaski:
This is redundant now that the changes have been made to master and the hotfix from master PR is open.

PR #832, opened 2024-03-07
[Feature] Loading spinner in button

riitasointi:
## Description

PR adds the `loading` prop to `<Button>`. With the loading state, `ariaLoadingText` is also required. 

PR also adds a permanent `<VisuallyHidden>` aria-live region in the DOM next to the button. In normal use, the element is just an empty div node. When the loading state is applied, the node gets populated with the `ariaLoadingText` string and it gets announced by screen readers. 

## How Has This Been Tested?

In styleguidist on macOS Chrome + Voiceover. More testing needed with browsers and screen readers. 

https://github.com/vrk-kpa/suomifi-ui-components/assets/17459942/64474f02-5085-41a2-a31c-9146a4335e05

### Button
- Add `loading` and `ariaLoadingText` props. Applies a loading spinner to the button and disables it.

### DIFF
--- a/src/core/ActionMenu/ActionMenu.tsx
+++ b/src/core/ActionMenu/ActionMenu.tsx
@@ -18,6 +18,7 @@ import {
   ButtonProps,
   ButtonVariant,
   ForcedAccessibleNameProps,
+  LoadingProps,
 } from '../Button/Button';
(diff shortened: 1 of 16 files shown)


LJKaski (approved):
Looks and works nicely! Good work! 👏

PR #840, opened 2024-03-20
[Bugfix] Empty StatusText icon render

LJKaski:
## Description
This PR sets the StatusText to only render the icon corresponding the status when the component has content. This is to avoid showing orphaned icons when the StatusText is present without content.

## Motivation and Context
This was a clear but that was noticed while making other changes.

## How Has This Been Tested?
Tested in styleguidist locally on Chrome

### StatusText
- Set icon to render only when component has children


riitasointi (approved):
Simple fix which addresses the issue 👍

PR #841, opened 2024-03-21
[Fix] Dropdown selected value when children change (case language change)

riitasointi:
## Description

PR fixes an issue with `<Dropdown>` where its apparent selected value did not get updated when children change. 

Case: Switching application's language. The selected DropdownItem's `value` remains the same, but its displayed value changes. We need to handle this case and force the recalculation of `selectedValueNode` in the Dropdown's state. 

## How Has This Been Tested?

Styleguidist on Chrome

### Dropdown
- Fix an issue where visible selected value did not get updated when changing application language (DropdownItem visible text changes but value remains the same)


LJKaski (approved):
Code looks good and seems to do the trick! 👍

PR #854, opened 2024-05-03
[Feature] Margin provider and CSS-based margin props

LJKaski:
## Description
This PR contains two changes / additions to the spacing utilities of the library:

1. It changes the `margin` props implementation so that instead of inline styles they are applied using `!important` CSS styles
2. It introduces a `SpacingProvider` HOC, which allows developers to give context-specific spacing rules to easily create layouts and spacing rulesets for repeating use cases.

## Motivation and Context
The margin prop implementation needed to be changed due to strict content security policies having issues with inline styles. The spacing provider pattern is a general quality of life improvement for certain use cases.

## How Has This Been Tested?
Mostly in styleguidist on chrome to make sure the changes and additions work as intended and don't cause any unwanted side effects. Also tested briefly on a React + Vite project to detect any obvious typescript issues.

### General
- **Breaking change:** Margin props are now implemented using `!important` CSS styles instead of style attributes
### SpacingProvider
- Introduce `SpacingProvider` HOC for providing context or even site-specific spacing rules.

### DIFF
--- a/.styleguidist/spacingprops.md
+++ b/.styleguidist/spacingprops.md
@@ -12,9 +12,9 @@ import { Button } from 'suomifi-ui-components';
 </div>;
 ```
 
-### Overriding with style attribute
+### Margin props and style attributes
 
-Margin properties add inline styling to the rendered HTML. If the component interface also accepts a style attribute, it takes precedence over margin property.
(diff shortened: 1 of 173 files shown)


riitasointi (approved):
Great work with this! 👍

PR #862, opened 2024-06-19
[Feature] FileInput

riitasointi:
## Description

This PR introduces the `<FileInput>` component. It has the design according to latest Figma plans and functionality to add one or multiple files, optional file previews, drag & drop functions and controlled state with metadata options. See Styleguidist for all relevant use case examples. 

NOTE: In the unit tests of this component, many of the core functionalities of this component are not tested at all. This is because the component uses `DataTransfer` and `FileList` under the hood to manage the flow of files from the user to the input and the component's internal state, and these are not supported by Jest.  ❌ 

(PR also moves Dropdown under the Form-directory in the project and removes some redundant lines from spacing provider docs)

## How Has This Been Tested?

In Styleguidist on MacOS Chrome + Safari and VoiceOver. iOS Safari. Also tested in  Styleguidist and in a React + Vite project on Windows + NVDA + Firefox and Chrome as well as Android + Talkback in same setups.

## Screenshots:
image
image
image
image

### FileInput
- Introduce the `<FileInput>` component


src/core/Form/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,603 @@
...
+}
+
+export interface ControlledFileItem {
+  file: File;
+  status?: 'default' | 'error' | 'loading';
+  errorText?: string;
+  ariaLoadingText?: string;
+  buttonText?: string;
+  buttonIcon?: ReactElement;
+  buttonOnClick?: (file: File) => void;
```
LJKaski, line 155: These could benefit from documentation since these are exported and used outside the library as well.

src/core/Form/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,603 @@
...
+}
+
+type InternalFileInputProps = FileInputProps & {
+  globalMargins?: GlobalMargins;
+};
```
LJKaski, line 160: This could just be `GlobalMarginProps`.

src/core/Form/FileInput/FileInput.tsx
```diff
@@ -0,0 +1,603 @@
...
+    dragAreaRef.current?.classList.remove('highlight');
+    const dt = e.dataTransfer;
+    const filesFromInput = dt?.files;
+    if (!controlledValue && filesFromInput && inputRef.current?.files) {
+      addNewFiles(filesFromInput, true);
+    }
+    if (propOnChange) {
+      propOnChange(filesFromInput || new FileList());
+    }
+  };
```
LJKaski, line 272: This could probably be refactored to avoid duplicate code. Like a general drag handler that sets or unsets the highlight class which could then be called for each scenario.

src/core/Form/FileInput/FileInput.tsx
LJKaski: Could be an almost direct copypaste from the styleguidist controlled state example.

src/core/Form/FileInput/FileInput.tsx
riitasointi: I'm not sure that is correct? Looking at, for example, TextInput and DateInput they have it like this

src/core/Form/FileInput/FileInput.tsx
riitasointi: Comments added 👍

src/core/Form/FileInput/FileInput.tsx
riitasointi: Changed this as discussed elsewhere

src/core/Form/FileInput/FileInput.tsx
riitasointi: Refactored this a bit as you suggested

src/core/Form/FileInput/FileInput.tsx
LJKaski: I think these need to be the jsdoc format in order to enable the tooltip inside vscode. So the comment format should be `/**  The actual file object */` instead of `// The actual file object`.

riitasointi: Ah yes, fixed this!

LJKaski (approved):
Excellent work on this! 💪

PR #866, opened 2024-08-02
[Feature] React types update

LJKaski:
## Description
This PR updates the version of react types in order to avoid typing issues in projects using the library.

Some components needed minor adjustments due to the update, which are included in this PR.

## Motivation and Context
Multiple projects reported having to do some typing hacks in order to use the library with newer react types. Updating the types to the library was the easiest solution option, and didn't seem to cause any issues.

## How Has This Been Tested?
Tested by building and using in a react+vite project where the older version caused a typing issue. This seemed to fix it.

### Dependencies
- Update `@types/react` to 18.3.3

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "suomifi-ui-components",
-  "version": "14.0.0",
+  "version": "15.0.0-beta.0",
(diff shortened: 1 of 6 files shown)


riitasointi (approved):
Seems to do the trick 👍

PR #872, opened 2024-08-23
[Fix] Redundant margin in CheckboxGroup and RadiobuttonGroup legends

LJKaski:
## Description
RadioButtonGroup and CheckboxGroup had redundant margins in their legends when label was hidden. This PR should fix the issue in a way that doesn't reintroduce an old Safari issue fixed in #699

## Motivation and Context
This was a bug that was noticed by a developer using the library.

## How Has This Been Tested?
So far tested by running locally on Styleguidiston Chrome + Windows

**Needs to be tested thoroughly in Safari**

### RadioButtonGroup,  CheckboxGroup
- **Breaking change:** fix group legend margin logic


riitasointi (approved):
Seems to do the trick, and works on Safari as well 👍

PR #874, opened 2024-08-30
[Fix] Adjust FileInput accessibility

riitasointi:
## Description

PR fixes various a11y related issues in the `<FileInput>` component as discussed in ticket https://jira.dvv.fi/browse/SFIDS-974 

- The component's structure now has a HTML fieldset + legend at the top level to limit the number of `<label>` elements
- In single file mode, the actual input is not focusable after a file has been selected. Focus moves to the "Remove" button instead
- File names are not focusable anymore when `filePreview` is not used
- Fixed redundant role=listitem when using single file mode

## How Has This Been Tested?

Styleguidist on macOS Chrome + Safari

### FileInput
- Fix various accessibility related issues

### DIFF
--- a/src/core/Form/FileInput/FileInput.baseStyles.tsx
+++ b/src/core/Form/FileInput/FileInput.baseStyles.tsx
@@ -121,13 +121,6 @@ export const baseStyles = (
     }
 
     .fi-file-input_input-outer-wrapper {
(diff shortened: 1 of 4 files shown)


LJKaski (approved):
Seems to address all the issues. Great work! 👍

PR #875, opened 2024-09-02
[Foc]FileInput visibility hidden in single file mode when file has been added

riitasointi:
## Related Issue

Closes #


src/core/Form/FileInput/FileInput.baseStyles.tsx
```diff
@@ -143,6 +143,10 @@ export const baseStyles = (
           display: flex;
           align-items: center;
 
+          &.fi-file-input_input_wrapper--hidden {
```
danielck, line 146: The classname in CSS and in FileInput.tsx do not match. `input_wrapper--hidden` (here) vs `input-wrapper--hidden` (in FileInput.tsx). The former is used consistently in CSS but the latter form (with a hyphen between input and wrapper) would match the class used for the base state of the wrapper.

danielck (changes requested):
Typo needs to be fixed: change `fi-file-input_input_wrapper--hidden` to `fi-file-input_input-wrapper--hidden` in CSS. Due to this the wrapper containing the input and label is still available to screen readers.

src/core/Form/FileInput/FileInput.baseStyles.tsx
riitasointi: Good point, fixed now

LJKaski (approved):
Seems to do the trick! 👍

PR #878, opened 2024-09-13
[Feature] RadioButtonGroup: Add groupStatus & groupStatusText

riitasointi:
## Description

PR adds `groupStatus` and `groupStatusText` props to `<RadioButtonGroup>`. They work similarly compared to the ones found in `<CheckboxGroup>` and the code is mostly copy-paste from there. 

As a "side effect", an individual `<RadioButton>` component can also now take in a `status` prop, but its usage is not illustrated on the documentation page. Radios are typically used in a group. 

## Motivation and Context

This feature is a no-brainer but has not been implemented before for one reason or another.

## How Has This Been Tested?

macOS Chrome/Safari + VoiceOver

## Screenshots

image
image

### RadioButtonGroup
- Add `groupStatus` and `groupStatusText` props


LJKaski (approved):
Looks good and clean and works well! Excellent and thorough work 👏

PR #886, opened 2024-11-29
[Feature] ConfigProvider and AutoID prefix

LJKaski:
## Description
This PR introduces a new HOC to the library, `ConfigProvider`. At this point it is used to give the user the option to give the internal AutoID utility a custom prefix to use in the generated IDs. In the future it can be expanded to be used for other global settings for the library.

## Motivation and Context
The AutoID prefix feature was requested by a project, and the ConfigProvider seemed like the best way to implement that.

## How Has This Been Tested?
So far tested by running locally in styleguidist and wrapping components in the provider

### ConfigProvider
* Implement `ConfigProvider` HOC. Can currently be used to give a custom prefix for autogenerated IDs of its children


riitasointi (approved):
Simple 👍

PR #888, opened 2024-12-09
[Feature] Update dev dependencies

LJKaski:
## Description
This PR updates a lot dev dependencies and adds the necessary changes to accommodate these updates. The biggest updates are:
- `React` from version 17 to 18
- `Styled-components` from version 5 to 6
- `Rollup` from version 2 to 4
- `@Testing-library/react` from 12 to 16 

Some smaller dependencies still need to be updated later, but this addresses the major ones.

## Motivation and Context
It had been a while since the last update for many of these dependencies, and in order to be able to support newer versions of these libraries in the projects using the library these need to be kept up to date.

## How Has This Been Tested?
Tested by running locally in a React + Vite project, a project using cjs and by running styleguidist on Chrome. Also tested in several Suomi.fi projects, some of which also provide a cjs build and/or use SSR. 

### General
* **Breaking change:** Update a lot of dev dependencies
  * For commonJS builds to work as previously, use named import for `styled` from styled-components. E.g. `import { styled } from 'styled-components';`


riitasointi (approved):
Very good work 👍

PR #890, opened 2024-12-16
[Fix] DatePicker escape key event propagation

LJKaski:
## Description
This PR fixes an issue where an escape key press in `DatePicker` would propagate outside the component, causing unwanted side effects. One example of a side effect is when a `DateInput` with `DatePicker` is used inside a modal. Before this fix, pressing escape to close the date picker would also close the containing modal.

## Motivation and Context
This was a bug reported by a service using the library.

## How Has This Been Tested?
Tested by placing a DateInput inside of a modal and checking the behavior before and after the fix.

### DateInput
* Stop event propagation of escape keypress


riitasointi (changes requested):
Placing the stopPropagation call there breaks keyboard navigation inside the calendar

riitasointi (approved):

PR #898, opened 2025-02-07
[Feature] Pagination accessibility and custom button variant

LJKaski:
## Description
This PR adds the possibility to replace the page change buttons with a custom component in Pagination. It also updates all the examples to show the preferred accessibility behavior upon page change, which is to focus the heading of the content area or a relevant focusable list element.

## Motivation and Context
This was a request from a service using the library. Using buttons for browsing the pages prevents search engines from indexing the paginated content. Thus they wanted the possibility to use links instead.

## How Has This Been Tested?
Tested by running locally on styleguidist

image

### Pagination
- Allow replacing browse buttons with custom components

### DIFF
--- a/src/core/Pagination/Pagination.md
+++ b/src/core/Pagination/Pagination.md
@@ -250,13 +250,17 @@ import {
   Block,
   Heading,
   Link,
+  Text,
   IconArrowRight,
   IconArrowLeft,
   suomifiDesignTokens
 } from 'suomifi-ui-components';
+import { useRef } from 'react';
 
 const [current, setCurrent] = React.useState(2);
 const lastPage = 8;
+const prevPageLinkRef = useRef();
+const nextPageLinkRef = useRef();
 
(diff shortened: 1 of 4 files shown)


src/core/Pagination/Pagination.md
```diff
@@ -228,6 +248,108 @@ const fakePageNode = (
...
+        </Link>
+      )
+    }
+    aria-label="Pagination"
+    pageInputProps={{
+      invalidValueErrorText: (value) => `"${value}" is not allowed`,
+      inputPlaceholderText: 'Go to',
+      buttonText: 'Jump to page',
+      labelText: 'Page number'
+    }}
```
riitasointi, line 349: These are redundant in this case, no pageinput in use

LJKaski: This should now be cleaned 👍

riitasointi (approved):

PR #902, opened 2025-02-21
[Bugfix] Drag and drop file handling in single file variant

LJKaski:
## Description
This PR fixes the drag and drop logic in `FileInput` to correctly handle cases where files get dropped into a single file FileInput. Previously dropping a file into a single file FileInput with already a file in it resulted in "bricking" the component for the user. The case was the same for dropping multiple file into a single file FileInput.

This PR should fix these issues without any adverse effects.

## Motivation and Context
This was a bug that was noticed while making other changes to the component. Bugs need to be fixed.

## How Has This Been Tested?
Tested locally on styleguidist on chrome. 

### FileInput
* Fix drag and drop file handling in single file variant


riitasointi (approved):
Seems to do the trick 👍

PR #909, opened 2025-03-24
[Fix] Table horizontal scroll a11y

riitasointi:
## Description

PR makes the `<Table>` component's wrapping div focusable when it has horizontal scroll. This is due to Safari not being able to handle scrolling with a keyboard otherwise.

## Motivation and Context

This issue was reported by an accessibility expert

## How Has This Been Tested?

macOS Chrome + Safari in styleguidist

### Table
- Make the component's wrapping div focusable when it has horizontal scroll to allow keyboard scrolling on Safari

### DIFF
--- a/src/core/Table/Table.tsx
+++ b/src/core/Table/Table.tsx
@@ -282,7 +282,7 @@ const BaseTable = <TColumns extends readonly TableColumn[]>(
   }
 
   const tableCaptionId = `${id}-caption`;
-  const getTableCaptionId = () => (caption ? `${id}-caption` : ariaLabelledBy);
+  const getTableCaptionId = () => (caption ? tableCaptionId : ariaLabelledBy);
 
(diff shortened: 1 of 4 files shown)


src/core/Table/Table.tsx
```diff
@@ -264,8 +281,17 @@ const BaseTable = <TColumns extends readonly TableColumn[]>(
     }
   }
 
+  const tableCaptionId = `${id}-caption`;
+  const getTableCaptionId = () => (caption ? `${id}-caption` : ariaLabelledBy);
```
LJKaski: Could/should this refer to the `tableCaptionId` on the previous line to avoid duplicate code?

riitasointi: Check

LJKaski (approved):

PR #913, opened 2025-05-15
Feature/search auto suggest

LJKaski:
## Description
This PR adds an autosuggest feature to SearchInput. It also changes the mark logic in popover lists to work so that instead of the user entry being bolded, the rest of the matching label is. SearchInputValue type is changed to just string too.

## Motivation and Context
This feature has been requested for a long time, and it allows changing third party solutions to our component in Suomi.fi services in cases where the suggestions are needed.

## How Has This Been Tested?
Tested by running locally on styleguidist as well as in a React + vite project.

image

### SearchInput
* **Breaking change:** remove SearchInputValue type and replace it with `string`
* Add search suggestion popover feature
### SingleSelect, MultiSelect
* **Breaking change:** flip bolding logic for matching text in popover menu items

### DIFF
--- a/src/core/Form/SearchInput/SearchInput.md
+++ b/src/core/Form/SearchInput/SearchInput.md
@@ -82,6 +82,13 @@ import { SearchInput } from 'suomifi-ui-components';
 import { useState } from 'react';
 
 const [controlledValue, setControlledValue] = useState('');
+const suggestedResults = [
+  { id: 1, label: 'Jalkapallo' },
+  { id: 2, label: 'Sulkapallo' },
+  { id: 3, label: 'Jääpallo' },
+  { id: 4, label: 'Koripallo' },
+  { id: 5, label: 'Pallo' }
+];
 
 <SearchInput
   labelText="Search the site"
@@ -90,7 +97,13 @@ const [controlledValue, setControlledValue] = useState('');
   visualPlaceholder="Write search terms..."
   onSearch={(value) => console.log(`Searching for ${value}...`)}
   value={controlledValue}
-  onChange={(newValue) => setControlledValue(newValue)}
+  onChange={(newValue) => {
+    console.log(newValue);
+    console.log(controlledValue.length >= 3);
+    setControlledValue(newValue);
+  }}
+  autosuggest={controlledValue.length >= 3}
+  suggestions={suggestedResults}
 />;
 ```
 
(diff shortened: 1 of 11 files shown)


src/core/Form/SearchInput/SearchInput.md
```diff
@@ -171,6 +174,64 @@ import { SearchInput } from 'suomifi-ui-components';
 />;
 ```
 
+### Search suggestions
+
+You can provide search suggestions for the user using the `autosuggest` and `suggestions` props. The suggestions are shown in a popover list under the input field. Provide a descriptive `suggestionHintText` to let screen reader users know there will be a suggestion list under the component, as the usual accessible pattern is not available for search input component. Also provide `ariaOptionsAvailableText` to inform screen reader users about the updating amount of suggestions.
+
+When the user selects a suggestion from the list, the `onSuggestionSelected` callback gets called with the `uniqueId` of the element.
+
+It's recommended to use debounce on fetching the suggestions to avoid unnecessary fetches. Do not use debounce prop of the component for this purpose.
```
riitasointi, line 183: I think this needs a bit more clarification for the user. Why should they not use the native debounce? What should they do instead? (Make them pay attention to the example below)

src/core/Form/SearchInput/SearchInput.md
```diff
@@ -10,6 +10,7 @@ Examples:
 - [Debounce](./#/Components/SearchInput?id=debounce)
 - [Full width](./#/Components/SearchInput?id=full-width)
 - [Hidden label](./#/Components/SearchInput?id=hidden-label)
+- [Search suggestions](./#/Components/SearchInput?id=search-suggestions)
```
riitasointi, line 13: At the top of the file, it still reads "The component is intended to be used as part of a traditional page search in the header of a website." This is probably not the case anymore?

danielck:
Looking pretty good now, all the previously reported accessibility issues are now fixed. In older versions of macOS & Safari browsing the suggestions doesn't work though. A workaround could be to update the input value to match the highlighted suggestion, and/or displaying the suggestions in the DOM immediately after the input.

danielck:
Tested in NVDA + Chrome, NVDA + Firefox, JAWS 2024 + Chrome, VO + Safari (macOS 14, 15 and 26 beta) and VO + Safari (iOS 18). All work fine with the exception of VO + Safari on iOS, where it does not seem to be possible to move screen reader focus to the search suggestions without closing the dropdown. I'm not sure what to suggest here: the [APG example](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/) does not have this issue, nor does our own SingleSelect. And the search suggestions work fine on iOS _without_ a screen reader. How does this work differently from e.g. SingleSelect in this respect?

riitasointi (approved):

PR #954, opened 2025-12-29
[Feature] SingleSelect & MultiSelect: Change popover focus logic (a11y improvement)

riitasointi:
This draft PR converts SingleSelect and MultiSelect to work similarly to the Dropdown component in terms of the popover. When browsing the popover list the browser's focus now actually goes to the `<li>` elements in the list. This fixes screen reader problems on macOS VoiceOver. Needs further testing though.

Making this change would also change the behavior of the components a little bit: So far, when browsing the popover list via keyboard, the browser's focus has remained in the input field and it has thus been possible to filter the list while browsing. With this change, that would not be possible anymore.

image


jenkrisu:
SingleSelect & MultiSelect will be updated with this PR that enables filtering: https://github.com/vrk-kpa/suomifi-ui-components/pull/967, so closing this draft.

PR #955, opened 2026-01-22
[Fix] SearchInput border radius and internal paddings

LJKaski:
## Description
This PR fixes some minor style issues in SearchInput after the previous update to the component. Namely the internal paddings were wrong in some cases and the right side border radiuses were wrong when used without built-in search button.

## Motivation and Context
Some styles were either missing or erroneous, and needed to be corrected.

## How Has This Been Tested?
Tested by running locally on styleguidist on Chrome and checking that the styles look and work correctly.

### SearchInput
* **Breaking change:** Minor internal padding and border radius fixes

### DIFF
--- a/src/core/Form/SearchInput/SearchInput.test.tsx
+++ b/src/core/Form/SearchInput/SearchInput.test.tsx
@@ -39,6 +39,31 @@ describe('snapshot', () => {
     const { container } = searchInputRendered;
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  it('should have matching structure with suggestions open', async () => {
+    const suggestions = [
+      { uniqueId: '1', label: 'apple' },
+      { uniqueId: '2', label: 'banana' },
+      { uniqueId: '3', label: 'cherry' },
+    ];
+    const { baseElement, getByRole, getAllByRole } = render(
+      TestSearchInput({
+        autosuggest: true,
+        suggestions,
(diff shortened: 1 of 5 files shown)


src/core/Form/SearchInput/SearchInput.test.tsx
```diff
@@ -44,6 +44,31 @@ describe('snapshot', () => {
     const { container } = searchInputRendered;
     expect(container.firstChild).toMatchSnapshot();
   });
+
+  it('should have matching structure with suggestions open', async () => {
```
jenkrisu, line 43: Should this have also explicit expects to test the structure, that apple is first, banana second, cherry third?

jenkrisu (dismissed):

src/core/Form/SearchInput/SearchInput.test.tsx
LJKaski: Could be a reasonable addition 👍 Though any changes to that should get caught in the snapshot as well since it contains the popover as well.

jenkrisu: That's true, snapshot changes though are in my experience a bit easier to overlook instead of when explicitly stating what the order would be!

jenkrisu (dismissed):

jenkrisu (approved):

PR #970, opened 2026-04-30
[Fix] Dropdown and Breadcrumb text scaling

LJKaski:
## Description
This PR Fixes a couple of accessibility issues where an increased font size caused the layout to break or some content not be legible. These issues are fixed with minor tweaks to styles in Dropdown and Breadcrumb components.

## Motivation and Context
These are the last two remaining accessibility issues known in this library, and needed to be addressed.

## How Has This Been Tested?
So far tested in styleguidist using Chrome and Firefox

### Breadcrumb
- **Breaking change:** remove fixed height to support font sizes incresed by user
### Dropdown
- **Breaking change:** Adjust styles of main dropdown button content to support larger font sizes

### DIFF
--- a/.styleguidist/accessibility.md
+++ b/.styleguidist/accessibility.md
@@ -18,11 +18,9 @@ Using the Suomi.fi component library does not automatically make your project ac
 
 ## Known issues
 
-Updated 24th April 2026.
+Updated May 8th, 2026.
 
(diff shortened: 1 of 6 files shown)


jenkrisu (dismissed):

jenkrisu (approved):
Tested with MacOS Chrome, Edge, Firefox. With Browserstack on Windows 11 Firefox, Edge.

PR #974, opened 2026-06-11
[Feature] DatePicker conditional rendering to DOM

jenkrisu:
## Description

Updates DatePicker to render to DOM conditionally. Previously, DatePicker was always in DOM, when `datePickerEnabled=true`.

## Motivation and Context

When DatePicker was always in DOM, the position calculation sometimes appended extra scrollable whitespace to DOM. There was also a bug where closing the DatePicker sometimes kept the month and year dropdowns visible in DOM.

## How Has This Been Tested?

- MacOS: Chrome and Safari with VoiceOver
- Windows: Edge and Chrome with NVDA (Browserstack)
- Android: Chrome with TalkBack
- iOS: Chrome and Safari with VoiceOver

### DateInput
- **Breaking change:** DatePicker is rendered to DOM conditionally when opened, instead of being always hidden in DOM. Fixes issue where DatePicker's dropdowns were sometimes visible when DatePicker was closed. Fixes issue where position calculation sometimes appended extra whitespace to DOM.

### DIFF
--- a/src/core/Form/DateInput/DateInput.md
+++ b/src/core/Form/DateInput/DateInput.md
@@ -118,7 +118,7 @@ Use the `minDate` and `maxDate` props to control the available dates in the date
 import { DateInput } from 'suomifi-ui-components';
 
 const minDate = new Date(2023, 1, 2);
-const maxDate = new Date(2024, 11, 31);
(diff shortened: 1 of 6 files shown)


LJKaski (approved):
Seems to be working well still! Good work 💪

PR #976, opened 2026-06-12
Feature/reorderable list

LJKaski:
## Description
This PR introduces the `ReorderableList` component

## Motivation and Context
This component/pattern was requested by multiple projects using the library, so it was implemented.

image

### ReorderableList
- **Breaking change:** Icons library updated to new major release 8.0.0.
- Introduce `ReorderableList` and `ReorderableList` components.


src/core/ReorderableList/IntearctionBlocker/InteractionBlocker.tsx
```diff
@@ -0,0 +1,62 @@
...
+      const element = el;
+      if (disableInteraction) {
+        element.dataset.prevTabindex = element.getAttribute('tabindex') ?? '';
+        element.setAttribute('tabindex', '-1');
+        element.setAttribute('aria-disabled', 'true');
+      } else {
+        const prev = element.dataset.prevTabindex;
+        if (prev === '') {
+          element.removeAttribute('tabindex');
+        } else if (prev != null) {
```
jenkrisu, line 41: Should this be !== null
```suggestion
        } else if (prev !== null) {
```

src/core/ReorderableList/ReorderableList/ReorderableList.baseStyles.tsx
```diff
@@ -0,0 +1,47 @@
...
+      gap: ${theme.spacing.s};
+      margin-bottom: ${theme.spacing.s};
+    }
+
+    & .fi-reorderable-list_instruction {
+      margin-bottom: ${theme.spacing.m};
+    }
+
+    & .fi-reorderable-list_list {
+      & > ul {
```
jenkrisu: Should this be styled with a class?

src/core/ReorderableList/ReorderableList/ReorderableList.tsx
```diff
@@ -0,0 +1,686 @@
...
+  /** Text for the save/done button shown in edit mode */
+  saveButtonText: string;
+  /** Text for the cancel/revert button shown in edit mode. If omitted, no revert button is shown. */
+  revertButtonText?: string;
+  /** Heading text for the instruction shown in edit mode */
+  editModeInstructionHeading: string;
+  /** Instruction text shown in edit mode */
+  editModeInstructionText: string;
+  /** Screen reader announcement callbacks (required for i18n) */
+  announcements: ReorderableListAnnouncements;
```
jenkrisu, line 113: Announcements description could have the ReorderableListAnnouncements interface copied, like DateInput has DatePickerTextProps.
```
/** 
* Screen reader announcement callbacks (required for i18n)
* <pre>
* ReorderableListAnnouncements {
*     editModeActivated: () => string;
 ...
```

src/core/ReorderableList/IntearctionBlocker/InteractionBlocker.tsx
LJKaski: Yes and no. The type for prev here is either string or undefined. If we use the strict check it would have to be `prev !== undefined`. It would maybe be clearer in terms of syntax, but this current one accounts for that case and also handles any freak accidents that would cause the value to be null (I don't know if there are any, but I feel like it's safer this way).

src/core/ReorderableList/ReorderableList/ReorderableList.baseStyles.tsx
LJKaski: Yes. Thanks for spotting it. Also switched to our internal `ul` component on the JSX side.

src/core/ReorderableList/ReorderableList/ReorderableList.tsx
LJKaski: Good point. Implemented your suggestion 👍

jenkrisu (dismissed):
Looks good! 👏

src/core/ReorderableList/ReorderableList/ReorderableList.md
```diff
@@ -0,0 +1,506 @@
...
+All text labels and screen reader announcements must be provided by the consumer for accessibility and i18n support.
+
+The optional `revertButtonText` prop adds a secondary cancel button in edit mode. When clicked, it restores the order to what it was when edit mode was entered and calls `onReorder` with the original order.
+
+Examples:
+
+- [Basic use](./#/Components/ReorderableList?id=basic-use)
+- [Items with form content](./#/Components/ReorderableList?id=items-with-form-content)
+- [Small screen](./#/Components/ReorderableList?id=small-screen)
+- [Move to top/bottom controls](./#/Components/ReorderableList?id=move-to-top/bottom-controls)
```
jenkrisu: Should fix this link

jenkrisu (approved):

PR #977, opened 2026-06-18
[Fix] ErrorSummary margin props and global margins

LJKaski:
## Description
This PR applies global margins and adds support for margin props to `ErrorSummary`. It also fixes a copy-paste error where ErrorSummary class name object was named `inlineAlertClassNames`

## Motivation and Context
Missing feature needed to be added and a simple code mistake fixed.

## How Has This Been Tested?
Locally on styleguidist on Chrome.

### ErrorSummary
- Add support for margin props and listener for global margins


jenkrisu (approved):
Good catch!

PR #979, opened 2026-06-23
[Fix] Input height variance

LJKaski:
## Description
This PR addresses an issue where input height would vary based on browser and display. This was due to the recent font change and some styles just happening to play nicely with the old one but not with the new one.

Decreasing the vertical padding in some components and using min-height fixes the issue and improves user experience when the user sets a higher font size value from the browser settings.

## Motivation and Context
This was a clear bug that was reported by projects using the library

## How Has This Been Tested?
Tested in styleguidist on brave, chrome and firefox with default and increased font sizes

### SingleSelect, MultiSelect, Pagination, TextInput, Dropdown, DateInput, SearchInput
- **Breaking change:** Change internal paddings and make component more flexible in terms of user-increased font sizes 
  - Fixes height of input from elements fluctuating between 40-42px (depending on browser and screen) with Source Sans 3

### DIFF
--- a/src/core/Form/DateInput/__snapshots__/DateInput.test.tsx.snap
+++ b/src/core/Form/DateInput/__snapshots__/DateInput.test.tsx.snap
@@ -297,7 +297,7 @@ exports[`snapshots match date input error status with statustext 1`] = `
   font-weight: 400;
   min-width: 245px;
   max-width: 100%;
-  padding: 8px 16px;
+  padding: 6px 16px;
   border: 1px solid hsl(202, 7%, 80%);
   border-radius: 2px;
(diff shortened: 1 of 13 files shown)


jenkrisu (dismissed):
Looks good! 👏

jenkrisu:
Release notes could also include that this fixes form element height changing between 40-42px depending on browser and screen

* Breaking change: Change internal paddings and make component more flexible in terms of user-increased font sizes
    * Fixes height of input from elements fluctuating between 40-42px (depending on browser and screen) with Source Sans 3

jenkrisu (approved):

PR #981, opened 2026-07-01
[Feature] RadioButton: display text as block

jenkrisu:
## Description

Aligns text in RadioButton to left, in case of long text that wraps to multiple lines.

## Motivation and Context

Text is aligned to keep the text style unified with hintText. Additionally this unifies styling with CheckBox, where text is likewise aligned to left if it wraps to multiple lines.

## How Has This Been Tested?

Tested with Styleguide, MacOS Chrome and Safari.

Old style:
image

New style:
Screenshot 2026-07-01 at 14 19 53

### RadioButton
- **Breaking change**: Text content is displayed as `block` to align text to left, in case long text wraps to multiple lines


LJKaski (approved):
Nice and simple solution 👍

PR #984, opened 2026-07-28
[Fix] Adjust dropdown and select min-width

LJKaski:
## Description
Dropdown had a somewhat high min-width value of 245px previously. This PR lowers it to 160px and also sets the same minimum width to the select components. Some adjustments were made to datepicker styles in order to accomodate these changes.

## Motivation and Context
Some projects needed dropdown to be somewhat narrower in certain context. This PR makes that possible and there was no huge reason to have the minimum width as high as it was.

## How Has This Been Tested?
Tested locally on styleguidist on chrome

### Dropdown
**Breaking change:** Set minimum width of container and open button to 160px
### SingleSelect, MultiSelect
**Breaking change:** Set minimum width of container and input to 160px


jenkrisu (approved):
Looks good!

PR #989, opened 2026-08-13
[Fix] Chip remove icon position on large fonts

LJKaski:
## Description
Sometimes the chip content might span multiple lines so that the component itself grows vertically as well. This PR fixes the icon position, so that it is vertically centered in those cases as well.

## How Has This Been Tested?
Tested locally in styleguidist with regular and forced larger font sizes on firefox and chrome.

Before:
image
After:
Screenshot 2026-08-13 at 13 16 33

### Chip
- Center remove icon vertically when height is larger than default


jenkrisu (approved):
Looks good!

LJKaski:
It was decided to keep this as it is.

PR #992, opened 2026-08-28
[Fix] Simplify ReorderableList announcements and example

LJKaski:
## Description
This PR gets rid of some redundant announcement messages and cleans up the examples as well. It also makes the `revertButtonText` prop mandatory.

## How Has This Been Tested?
Discussed with accessibility specialist, tested locally on styleguidist + chrome + voiceover.


src/core/ReorderableList/ReorderableList/ReorderableList.tsx
jenkrisu: These should also be removed from docs (movedUp, movedDown)

jenkrisu (approved):

PR #994, opened 2026-09-09
[Feature] Table: whole row selection

jenkrisu:
## Description

Adds option to select a row by clicking anywhere on the row. With assistive technology, the row selection remains the same. This feature is only an addition for people selecting rows by clicking.

## Motivation and Context

An option to select an item with a bigger click area was requested for better accessibility on some devices.

## How Has This Been Tested?

Tested with MacOS 26.6.2 VoiceOver Safari that selection works as before when navigating with keyboard.

State when user hovers over the row:
image

Comparison of selected row and hovered row:
image

### Table
- **Breaking:** Renamed internal css class `highlighted` to table `fi-table_tr--highlighted`
- Adds a new prop `wholeRowSelection`


src/core/Table/Table.tsx
```diff
@@ -46,6 +46,8 @@ import { Block } from '../Block/Block';
 const baseClassName = 'fi-table';
 
 const tableClassNames = {
+  selectedTr: 'highlighted',
```
LJKaski, line 49: should this be something along the lines of tr: `${baseClassName}_tr--highlighted` to match the general naming convention of states?

jenkrisu: This could be refactored, this was the existing classname so I didn't do any additional changes in scope of this change

LJKaski (approved):
Good work 😊

